PR #150, opened 2013-09-03
Experimentation with a better lifting model.

scshunt:
--DO NOT MERGE--

The purpose of this is just to let me run automated tests on these modifications to the lifting system, to see whether this preserves the functionality for IDs and embeddings, and in particular passes the tests. I'm unfamiliar, in particular, if introducing Maybes to those areas of code will break things (although the non-Maybe version could be restored easily).


gregwebs:
The code in question has changed a lot since this patch was created. I think some of the idea about improving SqlTypeExp was incorporated.

PR #250, opened 2014-07-08
Make withSqlPool actualy destroy the pool after finishing

feuerbach:


gregwebs:
created #437 for this

PR #253, opened 2014-07-21
persistent-mysql: provide all definitions through an exposed "Internal"  module.

fmap:
I'd like to use a few of the action renderers elsewhere.


snoyberg:
Looks like we dropped the ball on this, sorry @fmap. I'm going to have to close this for now. If this is still something you're interested in, please send a new PR against master. I'll make sure it doesn't bitrot to death this time :(

PR #340, opened 2014-12-23
refactor to expose some utilities for Esqueleto

albertov:
Moved some functionality regarding composite/custom primary keys to an exposed module (perhaps it should be under an "Internal" folder?) so Esqueleto can reuse it. See https://github.com/prowdsponsor/esqueleto/issues/87

### DIFF
--- a/persistent-test/persistent-test.cabal
+++ b/persistent-test/persistent-test.cabal
@@ -88,6 +88,7 @@ library
                      Database.Persist.Sql.Migration
                      Database.Persist.Sql.Internal
                      Database.Persist.Sql.Types
+                     Database.Persist.Sql.Util
                      Database.Persist.Sql.Raw
                      Database.Persist.Sql.Run
                      Database.Persist.Sql.Class
@@ -149,6 +150,7 @@ library
                    , resource-pool
                    , exceptions
                    , tagged
+                   , old-locale
 
    if flag(nooverlap)
      cpp-options: -DNO_OVERLAP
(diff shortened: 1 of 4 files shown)


andrewthad:
@gregwebs Could you merge this in? You might have to change the version number. I think that @albertov may have stalled on this, but I am going to see if I can at least get esqueleto to not fail with all tables that use CPKs. I'm not concerned at all about being able to use the `Key` definitions in queries, so I think I can solve this reasonably quickly if this branch gets in.

albertov:
@andrewthad take a look at this [PR](https://github.com/prowdsponsor/esqueleto/pull/88) if you haven't already. I managed to to be able to use composite/non-id keys as filter conditions in queries/joins and return entities which contain composite/non-id keys. 
I stalled on being able to return single keys (or as part of a tuple) since it requires some refactoring in Esqueleto as it was not designed to treat a column as a collection of them. I asked @meteficha for some guidance but haven't heard from him yet as I suppose he must be quite busy IRL.

andrewthad:
@albertov So do you think, even without that feature, that your branches are suitable to be merged in. Or, better put, my real question is: are any previously available features of esqueleto broken right now in your branch? If not, then I'd like to see it get merged in because there is value (namely, not crashing for entities with composite keys) in your branch that I would like to have available.

albertov:
@andrewthad, no, nothing was broken if i remember correctly. At least
according to the test

gregwebs:
thanks for the reminder. This is on hackage now.

PR #348, opened 2015-01-06
mongodb: switch from pool-conduit to resource-pool

gregwebs:


gregwebs:
@snoyberg: the `Pool` type exported from persistent-mongoDB is coming from a different package now. Does this require a major version bump? Technically it is a different type, but semantically it is the same type.

snoyberg:
In this case, it's the same `Pool` type anyway, since pool-conduit simply re-exports from resource-pool. So this is *not* breaking. I see no problem with this at all, were you waiting for me to merge it?

gregwebs:
oh, right, I should have figured that out, a sign I am a little too sleepy :)

PR #349, opened 2015-01-07
add support for jsonb type

Philonous:
postgresql 9.4 adds support for the jsonb type (a more efficient representation of json data). These changes seem to suffice to use it with sqltype=jsonb on Text fields.


gregwebs:
Cool! In the long-term postgres should be using jsonb and never encoding to a JSON string. That is, to use jsonb you would just define the field type as being one of the Entities that are defined.

snoyberg:
Thanks!

PR #354, opened 2015-01-16
Bump persistent-postgresql version to match hackage (2.1.2.2).

creichert:
I noticed the version number was one revision behind: http://hackage.haskell.org/package/persistent-postgresql-2.1.2.2

It might be worth noting that a user was having compile errors with this version earlier in IRC: http://dpaste.com/3RZE40T. I've encouraged them to open an issue.

### DIFF
--- a/persistent-postgresql/persistent-postgresql.cabal
+++ b/persistent-postgresql/persistent-postgresql.cabal
@@ -1,5 +1,5 @@
 name:            persistent-postgresql
-version:         2.1.2.1
+version:         2.1.2.2
 license:         MIT
 license-file:    LICENSE
 author:          Felipe Lessa, Michael Snoyman <michael@snoyman.com>


creichert:
Fixed some redundant imports I noticed in the build as well.

gregwebs:
doh, I forgot to commit the cabal version bump.

I will leave merging the redundant import warnings for @snoyberg. It isn't necessarily safe because they could be needed with a different (older) package set.

creichert:
Ah, that's a good point, thanks. Depending on the feedback, I could look into covering some of the other cases in the travis build.

gregwebs:
The way to cover it in travis would be to drop in a cabal.config file that has the lowest versions possible to build persistent with. This would also require specifying ghc 7.6 (although now that 7.10 is coming out it may soon be possible to drop 7.6 support).

snoyberg:
For now, if you can just update the pull request to undo the warnings fixes, let's get that merged in. @gregwebs is right about needing to be careful about breaking old versions.

creichert:
I removed the warnings fix. I'll try to look into getting some travis coverage as @gregwebs mentioned. Thanks! I'll merge the version bump.

PR #357, opened 2015-02-11
Fix ToJSON/FromJSON instance declarations

spl:
Based on the types, the previous instance declarations were not possible.


snoyberg:
Looks good to me. @gregwebs you agree?

gregwebs:
Thanks! Maybe we should be using doctest

PR #359, opened 2015-02-12
fix postgres array serialization

eryx67:
convert PersistList to Database.PostgreSQL.Simple.Types.PGArray instead of JSON

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -251,7 +251,7 @@ instance PGTF.ToField P where
     toField (P (PersistTimeOfDay t))   = PGTF.toField t
     toField (P (PersistUTCTime t))     = PGTF.toField t
     toField (P PersistNull)            = PGTF.toField PG.Null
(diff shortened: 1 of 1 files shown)


gregwebs:
This could be a good change if Postgres is flexible enough in its array storage. We have tests for a lot of different use cases, and some of them are failing with this change.

gregwebs:
#411 should cover this use case even if using a PGArray is more ideal for simple array use cases.

Vlix:
I've added PersistentArray to get the `array[]` syntax in [#863](https://github.com/yesodweb/persistent/pull/863), but that was to make the JSON operators work.
I don't know if it's desireable for `PersistList` to just use the `array[]` syntax, but I think I could integrate that change into my PR and drop the new `PersistArray` data constructor.
I saw PersistList using JSON marshalling so just thought it'd be better not to touch it, to be honest.

If anyone has an opinion about this, do share.

parsonsmatt:
I'm in favor of closing this one for #863 . Changing how `PersistList` is serialized seems like a bigger potential for breaking than I'd like to incur.

PR #362, opened 2015-02-16
added PersistField and PersistFieldSql instances for IntMap

fgaz:


gregwebs:
looks good to me except there is no test case.

fgaz:
fast-logger-2.2.3 failed during the building phase. The exception was:
    ExitFailure 1

That's odd. Can you trigger another build and see if it errors again?

MaxGabriel:
@fgaz I think that build failure happens consistently at this moment. I think its related to https://github.com/yesodweb/wai/issues/335 and maybe https://github.com/yesodweb/wai/issues/340

fgaz:
You're right. I was just about to open an issue on [fast-logger](http://github.com/kazu-yamamoto/logger), but I suppose it's not necessary, since this issue is already being worked on.

snoyberg:
I don't think we should block this merge on the fast-logger problem. If no one's opposed, I'd say merge it in. In case I'm missing some objection, I'll wait till tonight before merging (though Greg and Max should feel free to merge if they wish).

gregwebs:
Thanks for the test case!

PR #366, opened 2015-02-20
Multirow insert for SqlBackend + Postgres

AKurilin:
DO NOT MERGE: her @gregwebs , this is just a feeler, not the complete thing, want to get an idea if this is the direction you were thinking I could go with this new feature.

Let me know if there are details you are specific about (indentation, naming, more refactoring, function positioning etc) and I'll adjust them to fit the rest of the codebase as best as I can.

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -59,7 +59,7 @@ import qualified Blaze.ByteString.Builder.Char8 as BBB
 
 import Data.Text (Text)
 import Data.Aeson
-import Control.Monad (forM, mzero)
+import Control.Monad (forM)
 import Data.Acquire (Acquire, mkAcquire, with)
 import System.Environment (getEnvironment)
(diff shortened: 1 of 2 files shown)


gregwebs:
I don't actually do much on the SQL side of things, but this looks good. It would probably be better to start the change off in a way that makes the code available to all SQL backends if the backends are going to end up sharing their insert SQL.

creichert:
Looks good to me. What is left to be done?

> (indentation, naming, more refactoring, function positioning etc)

I didn't see anything on a first pass. The code style/etc looks fine. It would be nice to have this work with all SQL backends, though.

AKurilin:
Actually, as I was mentioning in another issue, it looks like insertMany_ for Postgres already does multi-row insertions by default, so this would be unnecessary. Wonder if someone more familiar with that codebase could confirm.

gregwebs:
`insertMany_` is implemented properly for SQL. The issue is `insertMany`

AKurilin:
Ok, thanks for clarifying.

Issue #330, opened 2014-12-05
persistent-mysql (MariaDB) migration issues


angerman:
Using Yesod with MariaDB through persistent-mysql, I get the following issues:

1. ```| crc              | bigint(20)   | YES  |     | NULL    |                |``` is being repeatedly migrated with ```Migrating: ALTER TABLE `CrashReport` CHANGE `crc` `crc` bigint(20) NULL``` where the model says: `    crc              Int      Maybe                        maxlen=20  sqltype=bigint(20)`.  I'm quite certain it has to be the `(Int, Maybe, sqltype=bigint(20))` combination, because it does *not* happen for a `(Int, sqltype=biging(11))` column.
2. How do I tell persistent to create an index, such that it would not issue the drop index migration: ```Migrating: ALTER TABLE `Application` DROP INDEX `name_UNIQUE````?
3. Trying to add a uniqueness constraint fails: I have `    name             Text                                  maxlen=45  sqltype=varchar(45)` as well as `UniqueName name` in the model, which results in the following migration: ```Migrating: ALTER TABLE `Application` ADD CONSTRAINT `unique_name` UNIQUE(`name`(200))```, which subsequently fails with `ConnectionError {errFunction = "query", errNumber = 1089, errMessage = "Incorrect prefix key; the used key part isn't a string, the used length is longer than the key part, or the storage engine doesn't support unique prefix keys"}`

If I can help somehow let me know!

angerman:
Can someone shed some light onto where the `(200)` comes from? I assume it's supposed to be the fields length. Now if one defines the length with `maxlen`, that's probably going to fail, right? How would one obtain the `maxlen` property and use it instead?

https://github.com/yesodweb/persistent/blob/master/persistent-mysql/Database/Persist/MySQL.hs#L684-L686

gregwebs:
in general one would add `maxlen=300` in the schema and then find that in the [fieldAttrs](https://github.com/yesodweb/persistent/blob/master/persistent-postgresql/Database/Persist/Postgresql.hs#L526)

angerman:
Thank you @gregwebs! Pulling the maxlen through into the AddUniqueContraint looks quite tedious. As I saw, the Postgres backend not having the `(200)` suffix, I just removed it from the mysql backend and had it working.

Now I'm a bit confused if the `(200)` is strictly necessary, or if `maxlen` in it's place would be strictly necessary, of if MySQL would not figure out the length on it's own based on the columns anyhow?

[...]

PR #367, opened 2015-02-22
Properly lookup maxlen, for the unique constraints key length.

angerman:
This resolves the issues mentioned in #330.

according to http://dev.mysql.com/doc/refman/5.1/en/alter-table.html, the syntax is:

```
ADD [CONSTRAINT [symbol]]
        UNIQUE [INDEX|KEY] [index_name]
        [index_type] (index_col_name,...) [index_option] ...
```
where
```
index_col_name:
    col_name [(length)] [ASC | DESC]
```

The `(200)` above, is the `length` part, and therefore will likely only work on fields that are at least `200` long. I suggest we remove the fixed length, until we can properly use the `maxlen` field of the appropriate column, if needed at all.
But as it is, it breaks using Uniques in Yesod. Say:

```
Table
  name    Text maxlen=45 sqltype=varchar(45)
  TableName name
```
would fail.

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -30,6 +30,8 @@ import Data.Text (Text, pack)
 import System.Environment (getEnvironment)
 import Data.Acquire (Acquire, mkAcquire, with)
 
+import Safe (readMay)
+
 import Data.Conduit
 import qualified Blaze.ByteString.Builder.Char8 as BBB
 import qualified Blaze.ByteString.Builder.ByteString as BBS
@@ -299,7 +301,7 @@ migrate' connectInfo allDefs getter val = do
         let uniques = flip concatMap udspair $ \(uname, ucols) ->
                       [ AlterTable name $
                         AddUniqueConstraint uname $
-                        map (findTypeOfColumn allDefs name) ucols ]
+                        map (findTypeAndMaxLen name) ucols ]
         let foreigns = do
               Column { cName=cname, cReference=Just (refTblName, a) } <- newcols
               return $ AlterColumn name (refTblName, addReference allDefs (refName name cname) refTblName cname)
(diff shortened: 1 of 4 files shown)


snoyberg:
I don't believe I understand why we'd want to do this if it breaks things. Can you clarify what problems this solves?

angerman:
Didn't catch that test failure. Will see how I can adapt it. 

What it would solve is the issue where you cannot create a unique constraint with a length that is larger than the maxlen of that field. The last model snipped in the comment fails with the current persist-mysql.

angerman:
Alright. This is now a little more involved. But adds a test case as well as the `maxlen=` lookup. I'm not completely satisfied with the outcome though, which I attribute to my lack of haskell experience. Any style and other suggestions are much appreciated.

Does this look good though?

creichert:
I can't speak on the bug or any potential compatibility issues, unfortunately. The test case doesn't make it clear to me that anything involving uniqueness is being tested.

Other than that the code does look good to me. I would probably suggest not bringing in `safe` for the single function (looks like you can avoid it?).

gregwebs:
@creichert's suggestion to not take one function from the `safe` package is a good one.

gregwebs:
This change looks good otherwise now that the tests are passing. Thanks for digging in and figuring it out.

angerman:
@creichert the model entries starting with a capital letter, followed by field names produce uniqueness constraints on the table. Hence those `ML...` fields produce the uniqueness fields, which in combination with fields where `fieldAttrs` are set to `maxlen=3` fall over, when the constraint is tried to be created with a key length of 200 instead of three.

After searching in `Text.Read` for decimal, instead of `Data.Text.Read`, I found `readMaybe`, which does the trick as well.

MaxGabriel:
Is it ok to use Text.Read.readMaybe, which according to its [docs](http://haddocks.fpcomplete.com/fp/7.8/20140916-162/base/Text-Read.html#v:readMaybe) was introduced in base 4.6.0 (which came with [GHC 7.6](http://downloads.haskell.org/~ghc/7.6.1/docs/html/users_guide/release-7-6-1.html#id476492)). I wasn't sure if persistent tried to support back to GHC 7.4.

gregwebs:
Thanks for catching that. We generally try to limit it to supporting 3 versions. So once 7.10 is released I will definitely be in favor of dropping 7.4. So I would be fine with this change since 7.10 will be out very soon.

snoyberg:
If we're dropping 7.4 support, can someone adjust the .cabal file to have a proper lower bound on base?

PR #369, opened 2015-02-26
Up dependency on the base package to >= 4.6 (removes GHC 7.4 support)

MaxGabriel:
Based on the discussion in #367 about dropping support for GHC 7.4


snoyberg:
Good move :)

MaxGabriel:
@snoyberg Should I apply this patch to the other Yesod repos as well?

snoyberg:
I don't think it's necessary. They're still free to work with older
versions of persistent for 7.4 support.

PR #372, opened 2015-03-02
break the EntityDef cycle with a SelfReference marker

gregwebs:


snoyberg:
I don't think I understand this code path well enough to be able to properly code review it. Is there something in particular you need feedback on?

gregwebs:
What should the version bump be?

gregwebs:
technically I think it should be a bump to 2.2 because it can ruin a case analysis. However, I think the data type is obscure enough that it is unlikely anyone outside of persistent is doing a case analysis.

snoyberg:
I'm OK with a minor version bump in this case.

gregwebs:
@snoyberg I finally finished this off. I am planning on bumping the persistent and persistent-template versions to 2.2 now. Let me know if that sounds good.

gregwebs:
The rest of the libs don't have to bump versions, it is just that the situation using persistent-template 2.1 and persistent 2.2 together needs to be avoided.

snoyberg:
I'd probably skip the major version bump, and if deemed necessary, we can use cabal file revisions to add upper bounds to the 2.1.X versions of persistent-template on Hackage.

gregwebs:
on hackage as 2.1.3 (for both packages). I also did a revision of persistent-template

PR #374, opened 2015-03-09
Add `Show` instances for MySQLConf, SqliteConf and PostgresqlConf

MaxGabriel:
Derives `Show` instances for configuration values. The main advantage is easier debugging + it let's you derive `Show` for records that contain a Conf value, like `AppSettings` in the scaffolding.

(MongoDB already derives `Show` for its configuration record).

### DIFF
--- a/persistent-mysql/ChangeLog.md
+++ b/persistent-mysql/ChangeLog.md
@@ -1,3 +1,7 @@
+## Master
+
+Added a `Show` instance for `MySQLConf`.
+
 ## 2.1.2.1
 
 Documentation typo fix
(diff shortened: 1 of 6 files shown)


persistent-mysql/ChangeLog.md
```diff
@@ -1,3 +1,7 @@
+## Master
```
snoyberg: This is fine, but typically we just go ahead and bump the version number based on the kind of change. In this case, it would be a minor version bump (e.g. 2.1.2.1 ==> 2.1.3).

snoyberg:
Good call!

persistent-mysql/ChangeLog.md
MaxGabriel: Cool, I'll bump the version in the future

PR #385, opened 2015-04-15
Treat unknown extension types as PersistDbSpecific values

albertov:
I'm in the process of implementing a library to create PostgreSQL full text search queries using Esqueleto (https://github.com/albertov/esqueleto-textsearch) and I need to create `PersistField` instance for postgres' `tsvector`, `regconfig` and `tsquery` types. I was tempted to add another set of converters for these types in `builtInExtensionGetters` but since it's not the first time I've done this (see #313) I think a more future-proof solution would be to treat all unknown oids as `PersistDbSpecific` values and let the `PersistField` instances parse them as they see fit.

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -227,9 +227,8 @@ withStmt' conn query vals =
                 rowRef   <- newIORef (LibPQ.Row 0)
                 rowCount <- LibPQ.ntuples ret
                 return (ret, rowRef, rowCount, oids)
-      getters <- forM ids $ \(col, oid) -> do
-          getter <- getGetter conn oid
-          return $ getter $ PG.Field rt col oid
+      let getters
+            = map (\(col, oid) -> getGetter conn oid $ PG.Field rt col oid) ids
       return (rt, rr, rc, getters)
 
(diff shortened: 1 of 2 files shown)


snoyberg:
@meteficha Since (I believe) you wrote this code originally, do you have any thoughts on this change?

meteficha:
IIUC, this pull request is reverting most of e0413f509c08196ea0137ef33e14e442b6e131a0, which will bring us to the state we were at afea41e4e44093aa962bd68bef2d25bb913e238f.  I've never used this stuff, but it seems @albertov knows what he's doing since he's the author of both commits above :).

I just don't understand why we should keep things like `(k PS.uuid, convertPV (PersistDbSpecific . unUnknown))` on `builtinGetters` since this is becoming the default behavior.

albertov:
@meteficha, Indeed, we're almost going back full circle :) I guess my rationale for e0413f5 was to be more explicit but now that I need access to more types I think it's better to map to a `PersistDbSpecific`  any unknown oids (which semantically makes sense since they're actually database-specific types) instead of going through the update builtInExtensionGetters + PR + version bump + release cycle every time someone needs to do something special with non-conventional types.

Regarding the the redundancy in `builtinGetters` I'm +1 on removing it. I can update the PR if y'all ok with it too.

snoyberg:
Merged, thanks!

Issue #375, opened 2015-03-09
Compile error when trying to use PersistMap with MongoDB


m-blazejczyk:
Hi,

To reproduce this problem, create a vanilla scaffolded site, and add the following to the models file:

```
Autor
    imiona Text Maybe
    nazwisko Text
    deriving Typeable
Kopalnia
    lookupId Int64
    autorzy (Map Text AutorId)
    UniqueKopalnia lookupId
    deriving Typeable
```

No other changes are required.

This is the error you will get when running `yesod devel`:

```
Model.hs:14:7:
    Not in scope: type constructor or class `AutorId'
    In the splice: $(persistFileWith upperCaseSettings "config/models")
```

Kind regards,
Michal

gregwebs:
What version of persistent are you using?

m-blazejczyk:
* persistent
    Installed versions: 2.1.1.7

* persistent-mongoDB
    Installed versions: 2.1.2.1

* persistent-template
    Installed versions: 2.1.0.1

* yesod-persistent
    Installed versions: 1.4.0.2

gregwebs:
Current workaround would be to not use the synonym and instead use `Key`.

```
Map Text (Key Autor)
```

Although this form is quite buggy (with mongodb). Please wait for the release of #387

m-blazejczyk:
Thanks!  For now, I have simply created an intermediate collection to solve this problem.

gregwebs:
#387 is merged

PR #387, opened 2015-04-26
Fix handling of (Key record) in embedding

gregwebs:
confirming the workaround for #375
led to writing a test case exposing bad behavior of

   field (Map Text (Key record))

This should not be marked as an embedding
It is just a normal map


snoyberg:
I have no objection to this, but I'm not familiar enough with the embedded code to be a useful code review.

Issue #342, opened 2014-12-31
Marshalling errors with explicit primary keys


Philonous:
An entity with an explicit primary key and a uniqueness constraint will produce marshalling errors when it is querried via the uniqueness type

example:
```haskell
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE MultiParamTypeClasses #-}


module Test where

import Control.Monad.Logger
import Control.Monad.Trans
import Database.Persist
import Database.Persist.Postgresql as Postgresql
import Database.Persist.TH

share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
Foo
    foo Int
    Primary foo
    bar Int
    UniqueBar bar
    deriving Show
|]

test = runStderrLoggingT . Postgresql.withPostgresqlConn "dbname=test" $
       \backend -> do
    runSqlConn (runMigration migrateAll) backend
    flip runSqlConn backend $ do
        k <- insert $ Foo 3 5
        test <- get k -- works
        liftIO $ print test
        getBy (UniqueBar 5) -- exception
```

ghci output:
```
Migrating: CREATe TABLE "foo"( PRIMARY KEY ("foo"),"foo" INT8 NOT NULL,"bar" INT8 NOT NULL)
[Debug#SQL] "CREATe TABLE \"foo\"( PRIMARY KEY (\"foo\"),\"foo\" INT8 NOT NULL,\"bar\" INT8 NOT NULL)" [] @(persistent-2.1.1:Database.Persist.Sql.Raw ./Database/Persist/Sql/Raw.hs:55:18)

Migrating: ALTER TABLE "foo" ADD CONSTRAINT "unique_bar" UNIQUE("bar")
[Debug#SQL] "ALTER TABLE \"foo\" ADD CONSTRAINT \"unique_bar\" UNIQUE(\"bar\")" [] @(persistent-2.1.1:Database.Persist.Sql.Raw ./Database/Persist/Sql/Raw.hs:55:18)

[Debug#SQL] "INSERT INTO \"foo\"(\"foo\",\"bar\") VALUES(?,?)" [PersistInt64 3,PersistInt64 5] @(persistent-2.1.1:Database.Persist.Sql.Raw ./Database/Persist/Sql/Raw.hs:55:18)

[...]

PR #388, opened 2015-04-27
Get unique primary

gregwebs:
1st commit is a somewhat unrelated cleanup for the generic use case (with a primary key). The 2nd commit fixes issue #342 for SQL.


gregwebs:
@snoyberg 3rd commit here prints out all SQL statements during testing. You might have some different ideas about how it should work, feel free to let me know and we can change it on master.

snoyberg:
It looks like a good change to me

PR #392, opened 2015-05-11
connect to local docker postgres container

gregwebs:


gregwebs:
This is a simple change just for my local integration. I am merging now but will check on the Travis builds.

PR #393, opened 2015-05-12
use modifyFailure for better error messages

### LINKED ISSUE #60: Error handling in configuration
Postgresql will print out an error:

```
user error (key "port" not present)
```

We should be using a default port. But also we should be using file-location or some technique so that the user knows the error is occurring in Persistent.

gregwebs:
closes #60


meteficha:
The respective cabal files need to have their aeson dependency bumped to `>= 0.6.2` if I understand correctly from the `modifyFailure` docs. Otherwise looks good to me :+1:.

snoyberg:
Thanks for taking care of this, and good catch @meteficha on the necessary lower bound, I added it in 8356fc6. Any objection to uploading these four packages to Hackage? (Not sure if there are other in-progress changes to be aware of.)

gregwebs:
I just pushed a similar change for redis & zookeeper, but I will leave it up to the authors of those packages to upload those.

gregwebs:
persistent packages are uploaded now (with stackage-upload).

snoyberg:
> with stackage-upload

:)

PR #397, opened 2015-05-16
add showMigation function

gregwebs:


snoyberg:
I don't see anything close to objectionable about this. Which is making me wonder: am I missing something controversial that you needed my review on? If not, this can definitely be merged in.

gregwebs:
I prefer to have code review whenever possible. Thanks!

Issue #399, opened 2015-05-25
Foreign Key and ID types do not match on 32bit architectures


ivanperez-keera:
Hi

On 32bit machines, the following model:
```
User
    ident    Text
    password Text Maybe
    name     Text Maybe
    UniqueUser ident
    deriving Typeable
Email
    email Text
    user UserId Maybe
    verkey Text Maybe
    UniqueEmail email
```
produces the following tables:

```
mysql> DESCRIBE email;
+--------+------------+------+-----+---------+----------------+
| Field  | Type       | Null | Key | Default | Extra          |
+--------+------------+------+-----+---------+----------------+
| id     | bigint(20) | NO   | PRI | NULL    | auto_increment |
| email  | text       | NO   | UNI | NULL    |                |
| user   | int(11)    | YES  | MUL | NULL    |                |
| verkey | text       | YES  |     | NULL    |                |
+--------+------------+------+-----+---------+----------------+
mysql> DESCRIBE user;
+----------+------------+------+-----+---------+----------------+
| Field    | Type       | Null | Key | Default | Extra          |
+----------+------------+------+-----+---------+----------------+
| id       | bigint(20) | NO   | PRI | NULL    | auto_increment |
| ident    | text       | NO   | UNI | NULL    |                |
| password | text       | YES  |     | NULL    |                |
| name     | text       | YES  |     | NULL    |                |
+----------+------------+------+-----+---------+----------------+
```
The types of ```user``` in the table ```email``` and ```id``` in the table ```user``` should match. Otherwise, foreign key constraints cannot be added. Yesod devel fails because of this.

See also: https://groups.google.com/forum/#!topic/yesodweb/DhciiGm3vBc

@gregwebs  pointed out the following:
* Are you on a 32 or 64 bit machine? I am thinking this issue must be attributed to the usage of the Int type resulting in different database types for different platforms:
 https://github.com/yesodweb/persistent/blob/a4f3b1fe7e8896b9bfaf5f4019e5af378296421c/persistent/Database/Persist/Sql/Class.hs#L212
* This points out that our testing is only done on 64 bit.

meteficha:
I think #400's solution is a terrible idea.  

  * We should use `Int64` for keys on every platform.

  * There should be no mention of `Int` on the codebase except for giving it useful instances.

That commit (ac82709) now makes it impossible to use 32-bit and 64-bit binaries on the same database, unless I'm missing something.

gregwebs:
[...]

PR #400, opened 2015-05-25
use the platform SqlInt type for the id

gregwebs:
addresses #399


gregwebs:
branch renamed in #401

PR #401, opened 2015-05-25
Platform sql int

gregwebs:


gregwebs:
closing in favor of #402

PR #402, opened 2015-05-26
don't use Int internally and throw an error on 32 bit machines when Int is used

### LINKED ISSUE #399: Foreign Key and ID types do not match on 32bit architectures

gregwebs:
Int32 or Int64 should be explicitly used instead

closes #399

### DIFF
--- a/persistent-redis/tests/basic-test.hs
+++ b/persistent-redis/tests/basic-test.hs
@@ -10,6 +10,7 @@ import Database.Persist.TH
 import Language.Haskell.TH.Syntax
 import Control.Monad.IO.Class (liftIO)
 import Data.Text (Text, pack, unpack)
+import Data.Int (Int64)
 
 let redisSettings = mkPersistSettings (ConT ''RedisBackend)
  in share [mkPersist redisSettings] [persistLowerCase| 
--- a/persistent-sqlite/test/Spec.hs
+++ b/persistent-sqlite/test/Spec.hs
@@ -9,6 +9,7 @@
 
 import Control.Monad.IO.Class  (liftIO)
 import Data.Time
+import Data.Int (Int64)
 import Database.Persist.Sqlite
 import Database.Persist.TH
 import Test.Hspec
--- a/persistent-template/test/main.hs
+++ b/persistent-template/test/main.hs
@@ -15,6 +15,7 @@ import Database.Persist.TH
 import Database.Persist.Types (PersistValue(..))
 import Data.Text (Text, pack)
 import Data.Aeson
+import Data.Int (Int64)
 
 share [mkPersist sqlSettings { mpsGeneric = False }, mkDeleteCascade sqlSettings { mpsGeneric = False }] [persistUpperCase|
 Person json
--- a/persistent-zookeeper/tests/basic-test.hs
+++ b/persistent-zookeeper/tests/basic-test.hs
@@ -10,6 +10,7 @@ import Database.Persist.Zookeeper
 import Database.Persist.Zookeeper.Internal
 import Database.Persist.TH
 import Language.Haskell.TH.Syntax ()
+import Data.Int (Int64)
 import Data.Maybe
 import Data.Pool ()
 import Test.Hspec
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -356,7 +356,7 @@ mkAutoIdField ps entName idName idSqlType = FieldDef
       , fieldType = FTTypeCon Nothing $ keyConName $ unHaskellName entName
       , fieldSqlType = idSqlType
       -- the primary field is actually a reference to the entity
-      , fieldReference = ForeignRef entName (FTTypeCon Nothing "Int")
+      , fieldReference = ForeignRef entName (FTTypeCon Nothing "Int64")
       , fieldAttrs = []
       , fieldStrict = True
       }
--- a/persistent/Database/Persist/Sql/Class.hs
+++ b/persistent/Database/Persist/Sql/Class.hs
@@ -211,7 +211,8 @@ instance PersistFieldSql Html where
     sqlType _ = SqlString
 instance PersistFieldSql Int where
     sqlType _
-        | bitSize (0 :: Int) <= 32 = SqlInt32
+        | bitSize (0 :: Int) <= 32 = error
+            "PersistFieldSql: Int not supported for 32 bit machines. Use Int32 or Int64 instead"
         | otherwise = SqlInt64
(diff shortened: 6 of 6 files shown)


snoyberg:
I'm OK with moving ahead with this. Will this result in a major version bump?

gregwebs:
Technically no bump is needed so it is up to us to decide.

snoyberg:
How is no bump needed if this results in compile-time errors for code that previously compiled and ran correctly? (Yes, I realize that some things weren't running correctly, but there still seems to be a large subset of things that *did* work on 32-bit systems.)

gregwebs:
I say no bump is needed because there is no API change. The breakages are related to generated code.

For a 32 or 64 bit user, they need to make sure `Int64` is imported.
For a 32 bit user, they need to replace `Int` with `Int32` or `Int64`.

Major version bumps are good for communicating breaking changes. So I would favor changing this patch to just remove the `Int` instance right now (so that we have 1 breaking change instead of 2 separate breaking changes so we only need 1 major version bump) and releasing with a major version bump.

snoyberg:
You and Felipe are following this issue better than I am, so I'd rather let you make the decision. I'm fine with the options that have been presented so far.

But I'd like to take a step back for a moment and make sure we're not blowing the problem out of proportion. It seems like the original issue was that which integral type (INTEGER4 or INTEGER8) was used for key references was architecture-dependent, to which the simple solution would appear to be to force them to always be INTEGER8/Int64.

As far as removing the Int instance: it's true that the Int datatype allows for overflow which is arch specific, but that's hardly something persistent has introduced here. We can have a discussion about whether the Int datatype was a good idea or not as the primary thing people interact with, but it's pretty moot.

Looking over the discussion again, this looks close to what Felipe said with https://github.com/yesodweb/persistent/issues/399#issuecomment-105286459

meteficha:
So it seems the three of us agree on the solution for keys, which is to use `Int64` everywhere.

The question is what to do about `Int`.  Looking at the tradeofffs, it seems there are three options:

  1. Do not do anything.  `Int`'s representation depends on the host.  I consider this behavior a bug.  However, it's not uncommon to keep documented bugs in order to preserve the API.

  1. Really bite the bullet once and for all.  Remove the `Int` instance.  Send the message to our users that they should be explicit about their intentions.

  1. Make `Int` behave as `Int64` when it has enough bits, refuse code otherwise.  Although @gregwebs said he'd prefer to bite the bullet, I can see the advantage of allowing most of our users to not do anything and have the convenience of not typing `Int64`.

Right now I'm torn between the tradeoffs so I can't choose between those options above.  If I was designing the API from scratch with the knowledge we have today, I'd avoid giving an instance to `Int`.  The persistent model is, after all, a representation of the data, so it matters how the data is represented.  Also, we shouldn't be silently dropping bits from numbers.

I'd also like to raise the question about version bumps regarding the change of the key type.  If we had any user who always developed on 32-bit machines, the change of key type to `Int64` is going to break their code, right?  I'm not sure if such user exists, though.

gregwebs:
I think the fact that this issue exists and was just reported now (the issue may not have existed before 2.0, but that release was over 6 moths ago) shows there are virtually no new 32 bit users of persistent. I could see 32 bit being a possibility for a new user using embedded SQLite though.

gregwebs:
With respect to Michael's comment, fixing just the basic issue can introduce the need to import Int64, which is a reason to perform a major version bump.

I am back to the idea of merging this as is and doing a major version bump with the change. That means keeping the `Int` instance without any immediate plan for removal, but making it mean 64 bits (and error on 32). I believe this will work well because I am under the impression that almost everyone is using 64 bits now.

snoyberg:
> With respect to Michael's comment, fixing just the basic issue can introduce the need to import Int64, which is a reason to perform a major version bump.

That seems like something we can easily work around by putting in some logic about where to find `Int64`. I think we already do that for other datatypes.

The 64-bit assumption is *not* true on Windows. Furthermore, this patch "sort of" breaks things at compile time, assuming that the code is all being run through TH at compile time. I feel pretty nervous about such an approach to be honest.

gregwebs:
> That seems like something we can easily work around by putting in some logic about where to find Int64

How do we do that?

> The 64-bit assumption is not true on Windows

Weird. I am on a 5 year old Windows machine right now and it is 64 bit.

snoyberg:
On Windows, 64-bit GHC is buggy, so many people don't use it.

I'm not sure we're on the same page with everything here, but I believed the compile-time errors you were talking about were generated code referring to a missing `Int64`. To fix that, you can just generate an `Exp` that refers to the `Int64` by complete package-module-name combo instead of by whatever's in scope (via `''Int64` instead of something like `ConT (mkName "Int64")`).

gregwebs:
ok, I added the special case to import Int64 in #406 and opened #405 for removing the instance.

PR #406, opened 2015-05-28
use Int64 for foreign keys

### LINKED ISSUE #399: Foreign Key and ID types do not match on 32bit architectures

gregwebs:
closes #399

### DIFF
--- a/persistent-template/Database/Persist/TH.hs
+++ b/persistent-template/Database/Persist/TH.hs
@@ -56,6 +56,7 @@ import qualified System.IO as SIO
 import Data.Text (pack, Text, append, unpack, concat, uncons, cons, stripPrefix, stripSuffix)
 import Data.Text.Encoding (decodeUtf8)
 import qualified Data.Text.IO as TIO
+import Data.Int (Int64)
 import Data.List (foldl')
 import Data.Maybe (isJust, listToMaybe, mapMaybe, fromMaybe)
 import Data.Monoid (mappend, mconcat)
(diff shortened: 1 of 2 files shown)


gregwebs:
The build passed. I can self-merge and release this.

@snoyberg it seems as though there are 2 duplicate Travis builds though: should we remove one of them?

gregwebs:
@snoyberg I released a new version of persistent-template. Unfortunately, persistent-template depends on persistent, so I cannot force coherent versions. once we release a new version of persistent, if anyone is stuck on an older version of persistent-template then they still could end up needing to import `Int64`. Any thoughts on how we should handle this?

snoyberg:
I'm not concerned about that case. Someone would have to specifically pin down there persistent version while letting the persistent-template version to float, which is an uncommon use case.

PR #407, opened 2015-06-02
Use Postgres' RETURNING capability to optimize `insertMany`.

### LINKED ISSUE #365: Single-statement bulk insert for SQL backends
I noticed earlier today that insertions of several rows with the Postgres backend are done invididually. I browsed throught he source and noticed that it's indeed called out explicitly in there:

    -- | Create multiple records in the database.
    --
    -- If you don't need the inserted @Key@s, use 'insertMany_'
    --
    -- SQL backends currently use the slow default implementation of
    -- @mapM insert@
    insertMany :: (MonadIO m, backend ~ PersistEntityBackend val, PersistEntity val)
               => [val] -> ReaderT backend m [Key val]
    insertMany = mapM insert

It sounds like we might want to flip things around and have the SqlBackend function `connInsertSql` be one that supports multiple rows at once, and have a single-row insert use that function instead, passing a [[PersistValue]] to it internally.

We would have to rewrite `insertSql'` to build multiple rows in its logic for Postgres, MySQL and SQLite (which does support multirow insertion starting with v. 3.7.1). Basically this function would have to be upgraded:

    -- | SQL code to be executed when inserting an entity.
    insertSql' :: EntityDef -> [PersistValue] -> InsertSqlResult
    insertSql' ent vals =
      let sql = pack $ concat
                    [ "INSERT INTO "
                    , escapeDBName $ entityDB ent
                    , "("
                    , intercalate "," $ map (escapeDBName . fieldDB) $ entityFields ent
                    , ") VALUES("
                    , intercalate "," (map (const "?") $ entityFields ent)
                    , ")"
                    ]
      in case entityPrimary ent of
           Just _ -> ISRManyKeys sql vals
           Nothing -> ISRInsertGet sql "SELECT LAST_INSERT_ID()"

Thoughts on this? You guys want me to take a stab? Not sure this will break some form of backwards compat, it shouldn't assuming we only change the interface of SqlBackend.

MaxGabriel:
Closes #365.

This PR uses Postgres' `RETURNING` capability to optimize bulk-inserts. Instead of inserting one-at-a-time, the rows can be inserted all at once and have their IDs returned. Based on my comment [here](https://github.com/yesodweb/persistent/issues/365#issuecomment-107081911), I don't think this is possible for SQLite/MySQL in a reliable way.

I believe test coverage is already pretty good on this, because several tests use `insertMany`, get back IDs, and then use those IDs to lookup the inserted rows.

Things I'm unsure about:
* The different `InsertSqlResult` types. Is `ISRManyKeys` supposed to be used when a composite primary key is defined?
* Version bumps. Is this considered a breaking change because it adds a new record field to public API?

### DIFF
--- a/persistent/Database/Persist/Sql/Class.hs
+++ b/persistent/Database/Persist/Sql/Class.hs
@@ -59,11 +59,11 @@ instance PersistField a => RawSql (Single a) where
     rawSqlProcessRow _     = Left $ pack "RawSql (Single a): wrong number of columns."
 
 instance (PersistEntity a, PersistEntityBackend a ~ SqlBackend) => RawSql (Key a) where
-  rawSqlCols escape key = (length $ keyToValues key, [])
+  rawSqlCols _ key         = (length $ keyToValues key, [])
   rawSqlColCountReason key = "The primary key is composed of "
                              ++ (show $ length $ keyToValues key)
                              ++ " columns"
-  rawSqlProcessRow = keyFromValues
+  rawSqlProcessRow         = keyFromValues
 
 instance (PersistEntity a, PersistEntityBackend a ~ SqlBackend) => RawSql (Entity a) where
     rawSqlCols escape = ((+1) . length . entityFields &&& process) . entityDef . Just . entityVal
--- a/persistent/Database/Persist/Sql/Class.hs
+++ b/persistent/Database/Persist/Sql/Class.hs
@@ -58,6 +58,13 @@ instance PersistField a => RawSql (Single a) where
     rawSqlProcessRow [pv]  = Single <$> fromPersistValue pv
     rawSqlProcessRow _     = Left $ pack "RawSql (Single a): wrong number of columns."
 
+instance (PersistEntity a, PersistEntityBackend a ~ SqlBackend) => RawSql (Key a) where
+  rawSqlCols escape key = (length $ keyToValues key, [])
+  rawSqlColCountReason key = "The primary key is composed of "
+                             ++ (show $ length $ keyToValues key)
+                             ++ " columns"
+  rawSqlProcessRow = keyFromValues
+
 instance (PersistEntity a, PersistEntityBackend a ~ SqlBackend) => RawSql (Entity a) where
     rawSqlCols escape = ((+1) . length . entityFields &&& process) . entityDef . Just . entityVal
         where
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -103,6 +103,7 @@ open' ci logFunc = do
         { connPrepare    = prepare' conn
         , connStmtMap    = smap
         , connInsertSql  = insertSql'
+        , connInsertManySql = Nothing
         , connClose      = MySQL.close conn
         , connMigrateSql = migrate' ci
         , connBegin      = const $ MySQL.execute_ conn "start transaction" >> return ()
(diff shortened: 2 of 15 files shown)


gregwebs:
This should re-use functionality from Sql/Util.hs. In particular, this will fail at least for composite keys and should use `dbIdColumns` for the returning clause. The test case for this should be added, perhaps in CompositeTest or PrimaryTest.

MaxGabriel:
Ok I added a failing test case for the composite primary key case, and then fixed that by adding a `RawSql` instance for `Key`s and using `dbIdColumnsEsc` from `Database.Persist.Sql.Util`. Probably I should add more tests for the `RawSql` instance, but is this generally correct otherwise?

gregwebs:
I actually haven't touched RawSql before, but it looks good now. Perhaps @snoyberg can look at the RawSql instance.

gregwebs:
One more thing, the documentation for `insertMany_` and `insert` should be updated now to reflect which SQL databases have efficient implementations.

persistent/Database/Persist/Sql/Class.hs
```diff
@@ -58,6 +58,13 @@ instance PersistField a => RawSql (Single a) where
     rawSqlProcessRow [pv]  = Single <$> fromPersistValue pv
     rawSqlProcessRow _     = Left $ pack "RawSql (Single a): wrong number of columns."
 
+instance (PersistEntity a, PersistEntityBackend a ~ SqlBackend) => RawSql (Key a) where
+  rawSqlCols _ key         = (length $ keyToValues key, [])
+  rawSqlColCountReason key = "The primary key is composed of "
+                             ++ (show $ length $ keyToValues key)
+                             ++ " columns"
```
MaxGabriel: Is there a better way to write `rawSqlColCountReason`? Maybe a way to list out the components of the key?

persistent/Database/Persist/Sql/Class.hs
```diff
@@ -58,6 +58,13 @@ instance PersistField a => RawSql (Single a) where
     rawSqlProcessRow [pv]  = Single <$> fromPersistValue pv
     rawSqlProcessRow _     = Left $ pack "RawSql (Single a): wrong number of columns."
 
+instance (PersistEntity a, PersistEntityBackend a ~ SqlBackend) => RawSql (Key a) where
+  rawSqlCols _ key         = (length $ keyToValues key, [])
```
MaxGabriel: Is it inefficient to use `keyToValues` just to determine the number of columns? Didn't come up with a way to get the column count otherwise.

gregwebs: Laziness may save us from any inefficiencies

gregwebs:
@MaxGabriel The changes look good now, thanks for the great patch with the updated ChangeLog and documentation!

@snoyberg this bumps persistent to 2.2 because the `SqlBackend` gets a `connInsertManySql` field added. However, this is only really an API breakage for someone implementing a new SQL backend, so we could probably avoid the major version bump if we wanted to. What do you think?

snoyberg:
I'm OK skipping the major version bump, we typically do so for internal-only changes like this.

gregwebs:
hmm, I am having second thoughts on the minor version bump. If someone pegs their backend version, but not persistent, then re-installing dependencies can cause a breakage. If we do the major bump, then at least we can say persistent should have been constrained to a major version.

@snoyberg are you ok with making this a major bump?

snoyberg:
Yes, I'm OK with it. Just to throw out one other possibility: we can add upper bounds via Hackage revisions to prevent the currently-released backends from using the newer persistent. I slightly prefer that approach, but only slightly.

gregwebs:
I tagged and released them all as version 2.2. That is enough work for me without changing a bunch of revisions on hackage.

PR #411, opened 2015-06-14
[WIP] add SqlType constructors to support JSONB

gregwebs:
This PR is opened at a very early state and nowhere near done.

The intention of this PR is to have PostgreSQL use the JSONB data type for embedded data rather than serialize things to JSON. Theoretically it may be possible to use PostgreSQL arrays and custom types to achieve this result, but I am assuming that JSONB is more performant particularly as the size of the data that is stored increases.

- [ ] postgresql-simple support?

- [ ] Migration support for JSONB
  - [x] SqlType support for embedding
  - [ ] Use the JSONB type in the migration

- [ ] Serialization support for JSONB
  - [ ] Convert to JSONB
  - [ ] Convert from JSONB

- [ ] PostgreSQL version
  - [ ] Have a check at startup for postgres version 9.4
  - [ ] Determine how to handle < 9.4 (maybe don't support it)

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -306,10 +306,10 @@ migrate' connectInfo allDefs getter val = do
         let foreigns = do
               Column { cName=cname, cReference=Just (refTblName, a) } <- newcols
               return $ AlterColumn name (refTblName, addReference allDefs (refName name cname) refTblName cname)
-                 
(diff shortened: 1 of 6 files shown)


MaxGabriel:
@gregwebs Can we keep an issue open about this? It seems like a good feature to have, especially for people serializing algebraic data types to JSON

gregwebs:
Sure, there are definitely issues where this has been mentioned already. I don't actually think the specific approach in this PR is a good one though.

isairz:
Any news about jsonb ?

Issue #413, opened 2015-07-03
Typeable instance for Entity


andrewthad:
Can we add a typeable instance to `Entity`?

snoyberg:
I see no reason not to. Want to send a PR?

andrewthad:
Sure.

snoyberg:
PR merged, thanks!

PR #414, opened 2015-07-03
Add a Typeable instance for PersistEntity

andrewthad:
Here it is
https://github.com/yesodweb/persistent/issues/413


snoyberg:
Thanks!

PR #418, opened 2015-07-13
allow read-only access to sqlite dbs with optional param in connection string

rubenmoor:
Sqlite connection can be set to read-only access now.
The default for sqlite connections remains intact (WAL on and READWRITE)

By adding "READONLY " behind an optional "WAL=off ", it is now possible to forbid write actions to the database. Also, when no db file exists, an error is thrown (in the READWRITE default case: the file is created).

The c function `sqlite3_open` is replaced by `sqlite3_open_v2` which allows to set the appropriate flag for read-only access and behaves like `sqlite3_open` otherwise.

### DIFF
--- a/persistent-sqlite/Database/Persist/Sqlite.hs
+++ b/persistent-sqlite/Database/Persist/Sqlite.hs
@@ -64,7 +64,12 @@ open' connStr logFunc = do
             | Just cs <- T.stripPrefix "WAL=off " connStr -> (cs, False)
             | otherwise                                   -> (connStr, True)
 
-    conn <- Sqlite.open connStr'
+    let (connStr'', readOnly) = case () of
+          ()
+            | Just cs <- T.stripPrefix "READONLY " connStr'  -> (cs, True)
+            | Just cs <- T.stripPrefix "READWRITE " connStr' -> (cs, False)
+            | otherwise                                      -> (connStr', False)
+    conn <- Sqlite.open connStr'' readOnly
     wrapConnectionWal enableWal conn logFunc
 
 -- | Wrap up a raw 'Sqlite.Connection' as a Persistent SQL 'Connection'.
--- a/persistent-sqlite/Database/Persist/Sqlite.hs
+++ b/persistent-sqlite/Database/Persist/Sqlite.hs
@@ -58,19 +58,21 @@ withSqliteConn = withSqlConn . open'
 
 open' :: Text -> LogFunc -> IO SqlBackend
 open' connStr logFunc = do
-    let (connStr', enableWal) = case () of
+    let (connStr', readOnly) = removeFirst "READONLY " connStr
+    let (connStr'', enableWal) = case () of
           ()
-            | Just cs <- T.stripPrefix "WAL=on "  connStr -> (cs, True)
-            | Just cs <- T.stripPrefix "WAL=off " connStr -> (cs, False)
-            | otherwise                                   -> (connStr, True)
+            | Just cs <- T.stripPrefix "WAL=on "  connStr' -> (cs, True)
+            | Just cs <- T.stripPrefix "WAL=off " connStr' -> (cs, False)
+            | otherwise                                    -> (connStr', True)
 
-    let (connStr'', readOnly) = case () of
-          ()
-            | Just cs <- T.stripPrefix "READONLY " connStr'  -> (cs, True)
-            | Just cs <- T.stripPrefix "READWRITE " connStr' -> (cs, False)
-            | otherwise                                      -> (connStr', False)
     conn <- Sqlite.open connStr'' readOnly
     wrapConnectionWal enableWal conn logFunc
+  where
+    removeFirst :: Text -> Text -> (Text, Bool)
+    removeFirst sub str = case T.splitOn sub str of
+        [_]    -> (str, False)
+        _:strs -> (T.intercalate sub strs, True)
+        []     -> (str, False) -- impossible case, splitOn never returns []
 
 -- | Wrap up a raw 'Sqlite.Connection' as a Persistent SQL 'Connection'.
 --
--- a/persistent-sqlite/Database/Persist/Sqlite.hs
+++ b/persistent-sqlite/Database/Persist/Sqlite.hs
@@ -58,21 +58,14 @@ withSqliteConn = withSqlConn . open'
 
 open' :: Text -> LogFunc -> IO SqlBackend
 open' connStr logFunc = do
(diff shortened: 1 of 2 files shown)


persistent-sqlite/Database/Sqlite.hs
```diff
@@ -183,24 +183,28 @@ sqlError maybeConnection functionName error = do
...
+                                error <- return $ decodeError error'
+                                case error of
+                                  ErrorOK -> do
+                                             database' <- peek database
+                                             active <- newIORef True
+                                             return $ Left $ Connection active $ Connection' database'
+                                  _ -> return $ Right error))
+open :: Text -> Bool -> IO Connection
+open path readOnlyFlag = do
+  databaseOrError <- openError path readOnlyFlag
```
MaxGabriel: nb: Adding a parameter would be a breaking change for people already using `open` to e.g. open a SQLite connection and enable foreign key checking (see https://github.com/yesodweb/yesod/wiki/Activate-foreign-key-checking-in-Sqlite)

persistent-sqlite/Database/Sqlite.hs
```diff
@@ -183,24 +183,28 @@ sqlError maybeConnection functionName error = do
...
-openError :: Text -> IO (Either Connection Error)
-openError path' = do
+foreign import ccall "sqlite3_open_v2"
+  openC :: CString -> Ptr (Ptr ()) -> Int -> CString -> IO Int
+openError :: Text -> Bool -> IO (Either Connection Error)
+openError path' readOnlyFlag = do
+  -- https://www.sqlite.org/c3ref/open.html
+  -- 1 = SQLITE_OPEN_READONLY
+  -- 6 = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE (behavior of sqlite3_open)
+  let flag = if readOnlyFlag then 1 else 6
```
MaxGabriel: Doesn't seem ideal to hardcode bitmask combinations. What about listing out the potential values and bitwise-ORing them manually? This approach would also make it easier to support the rest of the SQLite bitmask options, too. Something like this maybe?

```haskell
import Data.Bits

sqliteFlagReadWrite = 2
sqliteFlagCreate = 4

2 .|. 4 -- 6
```

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -64,7 +64,12 @@ open' connStr logFunc = do
             | Just cs <- T.stripPrefix "WAL=off " connStr -> (cs, False)
             | otherwise                                   -> (connStr, True)
 
-    conn <- Sqlite.open connStr'
+    let (connStr'', readOnly) = case () of
+          ()
+            | Just cs <- T.stripPrefix "READONLY " connStr'  -> (cs, True)
+            | Just cs <- T.stripPrefix "READWRITE " connStr' -> (cs, False)
+            | otherwise                                      -> (connStr', False)
+    conn <- Sqlite.open connStr'' readOnly
```
MaxGabriel: If I'm reading this code right, does this mean you have to have `WAL=on READONLY databaseName` in exactly that  order (i.e. you can't do `READONLY WAL=on databaseName`)? If so that seems like it'll mess people up

persistent-sqlite/Database/Sqlite.hs
rubenmoor: in this case, I might want to implement an entirely new function instead (e.g. `openV2` or `openWithFlags`). Would that be better?

persistent-sqlite/Database/Sqlite.hs
rubenmoor: Agreed, I was about to do that (there did not seem to be that many interesting flags/cases apart from 1 and 6)

persistent-sqlite/Database/Persist/Sqlite.hs
rubenmoor: Yap, that's as hacky as it looks. I'm hesitating to implement something more general before having discussed where this ship is sailing, first. A more general approach would be string parsing, i guess.

persistent-sqlite/Database/Sqlite.hs
MaxGabriel: Yeah I think that's better; having a readonly database seems like more of an edge case, so having just the database name as a parameter keeps things simple for the common case.

I personally like `openReadonly` as a function name if it's just the boolean parameter, and something like `openWithFlags` would be good if there were multiple flags or options supported by the function.

persistent-sqlite/Database/Persist/Sqlite.hs
MaxGabriel: One thought is that, long-term, this function should probably move away from passing and parsing a connection string. If there was a `SqliteOpts` data structure with default values (or something like that), this code would be simpler and more scalable for supporting future options. The main downside is that you'd have to duplicate the existing APIs that accept a connection string. Thoughts @gregwebs ?

MaxGabriel: Alternatively, would it be sufficient to just check for a substring to lookup the config option, and remove it if present? That wouldn't be sensitive to argument order.

gregwebs:
@MaxGabriel is this good to merge now?

MaxGabriel:
@gregwebs I don't think so, two reasons:
1. Adding an extra parameter to `open` is a breaking change that I don't think is unnecessary (also adding a `Bool` value for a readonly database isn't very scalable—there are many connection options that could be passed).
2. The connection parsing is pretty user unfriendly: `WAL=on READONLY databaseName` will work with this PR, but `READONLY WAL=on databaseName` won't. I don't see a good reason why that should be the case.

MaxGabriel:
@gregwebs With @rubenmoor's recent changes I think this is good to merge. Looks like an excellent PR—this opens up a bunch of configuration options for SQLite users (everything under "URI Filenames" [here](https://www.sqlite.org/c3ref/open.html)).

@rubenmoor Do you want to mention the URI filename feature on the [Database Configuration](https://github.com/yesodweb/persistent/wiki/Database-Configuration) Persistent wiki page?

MaxGabriel:
#526 can be closed after this is merged

snoyberg:
LGTM. I think it would be good to add some docs explaining that URIs are also supported. I don't want to block this PR for that though, are you interested in sending a separate PR?

Thanks!

MaxGabriel:
@rubenmoor added docs here, for reference https://github.com/yesodweb/persistent/wiki/Database-Configuration

PR #420, opened 2015-07-16
Have SQLite backend use multi-row inserts for insertMany_

### LINKED ISSUE #419: SQLite should support bulk-insert for insertMany_
Currently persistent-sqlite uses a one-by-one insert strategy for `insertMany_`. [A comment](https://github.com/yesodweb/persistent/blob/11d85dfe6eb36f72554a6b6239b93e339e57e044/persistent/Database/Persist/Sql/Orphan/PersistStore.hs#L218-L221) says this is because older SQLite versions don't support multi-row inserts in SQLite, but because persistent-sqlite bundles the SQLite C library this shouldn't be necessary.

This may create issues if people compile against an older system SQLite. @snoyberg pointed out:

> perhaps we should consider some kind of conditional compilation for older versions of SQLite, is such a thing is even possible.

[Related Yesod google group thread](https://groups.google.com/forum/#!topic/yesodweb/u3Rh1Of2Ta8)

MaxGabriel:
Closes #419 

This doesn't implement any checking of SQLite version before using the feature. I'm not sure if compiling Persistent with system SQLite is a common use case, though. Ubuntu LTS 12's default SQLite wouldn't support the multi-row insert, but Ubuntu LTS 14 would. Edit: Do you guys think it's worth implementing that? I imagine you'd pass the version of SQLite you're using as a config option?

Had some issues running the tests locally for this, but the existing test suite on Travis covers `insertMany_` in several spots so I'm not too concerned.

### DIFF
--- a/persistent/Database/Persist/Class/PersistStore.hs
+++ b/persistent/Database/Persist/Class/PersistStore.hs
@@ -95,10 +95,8 @@ class
 
     -- | Same as 'insertMany', but doesn't return any 'Key's.
     --
-    -- The MongoDB, PostgreSQL, and MySQL backends insert all records in
+    -- The MongoDB, PostgreSQL, SQLite and MySQL backends insert all records in
     -- one database query.
-    --
-    -- The SQLite backend inserts rows one-by-one.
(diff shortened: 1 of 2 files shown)


snoyberg:
Thanks!

Issue #422, opened 2015-07-19
Functions for getting free PersistField instances using JSON


andrewthad:
I've got some code that I copy and paste into a bunch of my projects. It looks like this:

    toPersistValueJSON :: ToJSON a => a -> PersistValue
    toPersistValueJSON = PersistText . Text.decodeUtf8 . LByteString.toStrict . encode

    fromPersistValueJSON :: FromJSON a => PersistValue -> Either Text a
    fromPersistValueJSON (PersistByteString bs) = case decode (LByteString.fromStrict bs) of
      Nothing -> Left "Could not parse the JSON"
      Just x -> Right x
    fromPersistValueJSON (PersistText t) = case decodeStrict (Text.encodeUtf8 t) of
      Nothing -> Left "Could not parse the JSON"
      Just x -> Right x
    fromPersistValueJSON a = Left $ "Expected PersistByteString, received: " <> Text.pack (show a)

I've actually got a slightly better version on another computer that works on different things that serialize to JSON values other than objects and arrays. The idea is that you can do this:

    instance PersistField MyData where
      fromPersistValue = fromPersistValueJSON
      toPersistValue = toPersistValueJSON

And assuming that you either wrote or derived those JSON instances, you get the `PersistField` instance for free. If this would be a welcome addition to persistent, I'd be happy to put up a PR for it.

gregwebs:
We already have `derivePersistFieldJSON`. But it would be much better to export 2 functions as you are proposing and to use those to derive the field.

andrewthad:
Cool. I'll put up a PR for this tomorrow.

andrewthad:
PR is up.

PR #423, opened 2015-07-20
Add functions to get PersistEntity instances from JSON instances

andrewthad:
https://github.com/yesodweb/persistent/issues/422

### DIFF
--- a/persistent/Database/Persist/Class.hs
+++ b/persistent/Database/Persist/Class.hs
@@ -41,6 +41,7 @@ module Database.Persist.Class
     -- * JSON utilities
     , keyValueEntityToJSON, keyValueEntityFromJSON
     , entityIdToJSON, entityIdFromJSON
+    , toPersistValueJSON, fromPersistValueJSON
     ) where
 
 import Database.Persist.Class.DeleteCascade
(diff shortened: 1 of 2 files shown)


creichert:
Just a heads up that the patch doesn't recognize you're account as the commiter (although it is under you're name). Not sure if that's intentional or not.

creichert:
Looks great. Is there a plan to support these in derivePersistFieldJSON like @gregwebs  suggested? Might also help to add the actual `instance` declaration snippet to to the haddocks.

andrewthad:
Ah! Thanks. I didn't realize the actual meaning of @gregwebs suggestion until you rephrased it. I don't actually know how to define template haskell functions at all, but I could take a stab at updating `persistent-template` after this gets merged in. I'll update the docs to put in an example, and I think I'm going to tweek the implementation one more time.

Also, thanks for the heads up about github not recognizing me as the commiter. I'll try to figure out how I did that and then not do it again.

creichert:
@andrewthad I went ahead and added the doc examples. Do you want to change the commiter at all before merging?

andrewthad:
Thanks. I looked into how to change the commiter, and I'll pass on changing it now. I think I would have to do force push, and I'm timid when it comes to doing that kind of thing. Feel free to go ahead and merge it in.

creichert:
Awesome patch, I'm looking forward to this one. Can this branch be safely deleted?

Issue #431, opened 2015-07-23
Migrate from Error to Except Module


psibi:
Since [Control.Monad.Trans.Error](https://hackage.haskell.org/package/transformers-0.4.3.0/docs/Control-Monad-Trans-Error.html) module has been deprecated, it is better to switch to [Control.Monad.Trans.Except](https://hackage.haskell.org/package/transformers-0.4.3.0/docs/Control-Monad-Trans-Error.html). What is the general opinion on this ? To port to `Except` or wait a bit more ?

gregwebs:
Sounds fine to me. @snoyberg probably has a better understanding of the implications.

psibi:
Also, I don't think much code has to be changed. Only `Persist.Types.Base` and `Database.Persist.MySQL` needs some modification, I guess.

gregwebs:
yeah, I did not think it affected us a lot. We generally prefer to use exceptions when in IO and this is a database library.

snoyberg:
For Persist.Types.Base: it looks like we're just providing an instance of the `Error` typeclass. I don't see any reason to stop providing that. In `MySQL`, we could switch to `Except` without any breakage (AFAICT).

PR #432, opened 2015-07-23
Port to ExceptT

psibi:
Related to #431

### DIFF
--- a/persistent/Database/Persist/Types/Base.hs
+++ b/persistent/Database/Persist/Types/Base.hs
@@ -6,7 +6,6 @@ module Database.Persist.Types.Base where
 import qualified Data.Aeson as A
 import Control.Exception (Exception)
 import Web.PathPieces (PathPiece (..))
-import Control.Monad.Trans.Error (Error (..))
 import Data.Typeable (Typeable)
 import Data.Text (Text, pack)
 import qualified Data.Text as T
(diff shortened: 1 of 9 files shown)


persistent/Database/Persist/Types/Base.hs
```diff
@@ -242,8 +241,6 @@ data PersistException
     deriving (Show, Typeable)
 
 instance Exception PersistException
-instance Error PersistException where
```
snoyberg: This is a breaking change. I'm not opposed to it in the future, but it shouldn't be undertaken immediately.

psibi: I'm opening this PR with changed bound version for transformers, so that this can be integrated whenever the new breaking version is released.

creichert:
I added as a milestone to 2.3 for the breakage.

erikd:
:+1: Looking forward to this landing, but I can wait.

psibi:
Yeah, I will try to integrate this along with this https://github.com/yesodweb/persistent/pull/618 which will likely result in a major version change for persistent.

snoyberg:
In my warnings cleanup, I already moved over to `runExceptT` internally. I think we should leave the `Error` instance for now, as it doesn't cause any harm to the package, and keeps backwards compatibility. Closing, please reopen if you disagree.

PR #433, opened 2015-07-25
Minor cleanup.

psibi:


persistent/Database/Persist/Class/PersistField.hs
```diff
@@ -232,9 +230,9 @@ instance PersistField UTCTime where
             _ ->
                 case parse8601 $ T.unpack t of
                     Nothing -> Left $ T.pack $ "Expected UTCTime, received " ++ show x
-                    Just x -> Right x
+                    Just y -> Right y
```
psibi: This is done because of the shadowing warning.

gregwebs:
Looks good. I don't know when parseTimeM was added though

psibi:
@gregwebs From `time 1.5`, I have included the version bump for that package in the patch.

creichert:
I'm not sure why the build failed, looking into it.

gregwebs:
The build failed because .travis.yml needs to use `-y` with `apt-get`. I pushed that change to master. @psibi can you rebase this branch?

We cannot move up to time-1.5 unless that works with ghc 7.6. I guess the travis build will tell us that.

psibi:
@gregwebs  Thanks, rebased it on top of master. I hope it works, waiting for travis build to show up the result.

gregwebs:
problems upgrading time. for every GHC version. Probabaly one of the persistent backends you are not building wants 1.4. The best approach would be to use CPP macros to allow time >= 1.4

psibi:
@gregwebs Out of the two build jobs in Travis, only one seems to be failing because of `time` package. I have bumped cabal version in that. Can [you trigger a travis rebuild] to see how it goes ?

gregwebs:
LTS Haskell is still on time-1.4 So I think we need to use a CPP macro to allow for both versions of time (or continue to use the deprecated API).

psibi:
@gregwebs Ah, okay. I will probably write a CPP macro, although I feel that that doesn't make the code good.

gregwebs:
I think that putting CPP macros in 1 place of import makes the code good. The problem is when the macros are used in multiple places. I would like to have an *internal* Persistent.Import module that all the other modules can import where common imports (including those that need CPP macros) reside.

psibi:
Closing this issue, I will work on the time related patch separately.  Have opened #434 which is the same patch except time related stuffs.

PR #435, opened 2015-07-27
Remove version constraints.

psibi:
Both the build jobs[1][2] are failing because of the package time.
I'm removing the version constraint in time since there is already a
constraint placed on time by the persistent package. And looking on the
code of persistent-zookeeper, I don't see why a minimal version of 1.4
is required.

https://travis-ci.org/yesodweb/persistent/jobs/72733457
https://travis-ci.org/yesodweb/persistent/jobs/72733456

### DIFF
--- a/persistent-zookeeper/persistent-zookeeper.cabal
+++ b/persistent-zookeeper/persistent-zookeeper.cabal
@@ -23,7 +23,7 @@ library
                    , template-haskell
                    , text                  >= 0.8
                    , aeson                 >= 0.6.2
-                   , time                  >= 1.4      && < 1.5
+                   , time                  
                    , attoparsec
                    , mtl
                    , transformers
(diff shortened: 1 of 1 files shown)


creichert:
Eek, this build script is quite unstable at the moment, sorry about that. I suppose postgres is waiting for input:
```
sudo -u postgres createuser -p 5432 travis &>/dev/null
Shall the new role be a superuser? (y/n)
```

but I don't understand why it happens sporadically.

psibi:
@creichert Yeah not sure why it's happening. For two out of the four build jobs, it doesn't happen. Very confusing.

psibi:
Adding more data points for build jobs failing because of the constraint in time package for zookeeper:
https://travis-ci.org/yesodweb/persistent/jobs/72916301
https://travis-ci.org/yesodweb/persistent/jobs/72916300

creichert:
Thanks for looking into this one, I've re-sparked the build and don't see any issue with doing this if I can get the build passing.

psibi:
@creichert This [build job](https://travis-ci.org/yesodweb/persistent/jobs/72796586) is failing but I don't think my patch is causing it to fail.

gregwebs:
yeah, the failure is on the hzk package, should not be related to this.

Issue #274, opened 2014-08-28
print Queries even if the database isn't actually present


psibi:
I have been using persistent basically as a data modelling tool to typecheck my table and in the future use it as an migration tool. For postgreSql, for printing out the Sql queries I have to write the following code:

````
connStr = "host=localhost dbname=test user=test password=test port=5432"

main :: IO ()
main = withPostgresqlPool connStr 10 $ \pool -> do
  flip runSqlPersistMPool pool $ do
    printMigration migrateAll
````

The problem with the above code is that if the actual database doesn't exist, it doesn't print any of the queries. I looked into the codebase for a while and wasn't able to write any functions to accomplish this task. Ideally, I would like to have any helper function which generates the sql queries for a particular database (somewhat like `parseTest` from attoparsec where you can test it interactively). Is this possible ?

snoyberg:
Closing out old issues, please reopen if still relevant

psibi:
@snoyberg I think only the project owners in github can re-open the issue. That being said, I would like to have this feature in persistent ?

snoyberg:
Would you be interested in attempting a PR for it then? I'd be happy to
provide any guidance necessary

psibi:
@snoyberg Thanks, I will attempt to patch this.

psibi:
@snoyberg  and others, can you review the pull request #436  ?
 The sample input and output using the `mockMigrate` function is demonstrated in this gist: https://gist.github.com/psibi/5ac89accf2e9c76f6d71

PR #436, opened 2015-07-27
Postgresql mock migrate

psibi:
For #274. 

I have created a `mockMigration` function which achieves the desired functionality for Postgresql.


gregwebs:
Thanks for digging in and figuring it out!

PR #437, opened 2015-08-02
destroy pool at the end of withPool

gregwebs:
originally #250


creichert:
@gregwebs nice, this looks great.

PR #438, opened 2015-08-02
Cleanup imports in zookeper releated modules

psibi:


gregwebs:
You have to be careful with cleaning up imports in libraries because different versions of dependencies have different exports. In this case different versions of base as can be seen from the Travis build. If you want to you can depend on base-compat

psibi:
Yes, I can see why the build failed.  Didn't knew about `base-compat`, but it looks good, I will try to look into that in my spare time. You don't mind the extra dependency of `base-compat` ?

gregwebs:
The dependency should be fine for persistent-zookeeper. We might want to avoid it for persistent itself, but I am not sure why.

parsonsmatt:
I'm going to close this since it is stale. Feel free to reopen if needed.

PR #440, opened 2015-08-02
Add helper functions for getting a PersistField instance for Enums

andrewthad:
Similar to the JSON helper function I created a while back. The haddocks I've added should explain the motivations. I'll copy them here: 

+-- | Convenience function for getting a free 'PersistField' instance
+-- from a type with an 'Enum' instance. The function 'derivePersistField'
+-- from the persistent-template package should generally be preferred.
+-- However, if you want to ensure that an @ORDER BY@ clause that uses 
+-- your field will order rows by the data constructor order, this is 
+-- a better choice.

### DIFF
--- a/persistent/Database/Persist/Class/PersistEntity.hs
+++ b/persistent/Database/Persist/Class/PersistEntity.hs
@@ -314,7 +314,7 @@ fromPersistValueJSON z = case z of
 --
 -- @
 -- data SeverityLevel = Low | Medium | Critical | High
---   deriving (Enum, Boundud)
+--   deriving (Enum, Bounded)
 -- instance PersistField SeverityLevel where
 --   fromPersistValue = fromPersistValueEnum
 --   toPersistValue = toPersistValueEnum
@@ -330,7 +330,7 @@ toPersistValueEnum = toPersistValue . fromEnum
 --
 -- @
 -- data SeverityLevel = Low | Medium | Critical | High
---   deriving (Enum, Boundud)
+--   deriving (Enum, Bounded)
 -- instance PersistField SeverityLevel where
 --   fromPersistValue = fromPersistValueEnum
 --   toPersistValue = toPersistValueEnum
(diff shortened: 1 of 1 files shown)


persistent/Database/Persist/Class/PersistEntity.hs
```diff
@@ -301,3 +303,43 @@ fromPersistValueJSON z = case z of
...
+-- from the persistent-template package should generally be preferred.
+-- However, if you want to ensure that an @ORDER BY@ clause that uses 
+-- your field will order rows by the data constructor order, this is 
+-- a better choice.
+-- 
+-- Example usage in combination with `fromPersistValueEnum`:
+--
+-- @
+-- data SeverityLevel = Low | Medium | Critical | High
+--   deriving (Enum, Boundud)
```
creichert: s/Boundud/Bounded, typo?

persistent/Database/Persist/Class/PersistEntity.hs
```diff
@@ -301,3 +303,43 @@ fromPersistValueJSON z = case z of
...
+
+-- | Convenience function for getting a free 'PersistField' instance
+-- from a type with an 'Enum' instance. This function also requires
+-- a `Bounded` instance to improve the reporting of errors.
+--
+-- Example usage in combination with `toPersistValueEnum`:
+--
+-- @
+-- data SeverityLevel = Low | Medium | Critical | High
+--   deriving (Enum, Boundud)
```
creichert: same here

creichert:
Interesting, looks good but I'm curious if augmenting `derivePersistField` was considered instead. Also, is this a function that should be exported along with `derivePersistField` from persistent-template?

creichert:
Ok, I jumped the gun on the questions. It's a non-TH solution.

andrewthad:
@creichert I've corrected the spelling of `Bounded` in both places. Unfortunately, I've never learned how to write template haskell, but it seems like having `derivePersistFieldEnum` in `persistent-template` could be helpful. Then again, it would only bring you down from three lines of code to one, which isn't that big of a win (especially when you consider how seldom this behavior is needed).

PR #451, opened 2015-08-15
Distinguish between binary and non-binary strings in MySQL

### LINKED ISSUE #448: A custom field used with persistent-mysql is serialized as PersistText but deserializes as PersistByteString
I'm investigating [an issue](https://github.com/Tarrasch/yesod-text-markdown/issues/12) that's affecting yesod-text-markdown, which adds `PersistField` and `PersistFieldSql` instances for `Markdown` from snoyberg's [markdown package](https://github.com/snoyberg/markdown). 

That package serializes `Markdown` (a newtype around lazy Text) to `PersistText`, but when looking up the value it gets a `PersistByteString` on the MySQL backend only. A couple of [stackoverflow](http://stackoverflow.com/questions/31898691/storing-markdown-in-a-yesod-persist-model#) [questions](http://stackoverflow.com/questions/31902061/what-does-not-a-persisttext-value-mean) are reporting this error.

I added a failing test case demonstrating this issue [on this branch](https://github.com/yesodweb/persistent/compare/mysqlSerializationError?expand=1) ([Travis CI results here](https://travis-ci.org/yesodweb/persistent/builds/75670539).

I'm investigating the cause of this, ~~potentially it was [this commit](https://github.com/yesodweb/persistent/commit/7410aa6b3d91cf97d887f6a2035b1488f1371cc7)?~~

MaxGabriel:
Closes #448 

This PR adds a failing test case demonstrating persistent-mysql writing a `PersistText` but reading a `PersistByteString`. The second commit of the PR uses character set information to distinguish between ~~TEXT and BLOB~~ binary and non-binary strings in MySQL, fixing the failing test.

This PR could break existing code. If code is relying on reading a `PersistByteString` from a MySQL `TEXT` column, that code will break (as it did for `getColumn`).

### DIFF
--- a/persistent-mysql/ChangeLog.md
+++ b/persistent-mysql/ChangeLog.md
@@ -1,3 +1,9 @@
+## 2.3
+
+* Distinguish between binary and non-binary strings in MySQL [451](https://github.com/yesodweb/persistent/pull/451)
+	* Previously all string columns (VARCHAR, TEXT, etc.) were being returned from Persistent as `PersistByteString`s (i.e. as binary data). Persistent now checks character set information to determine if the value should be returned as `PersistText` or `PersistByteString`. 
+	* This is a **breaking change** if your code is relying on a `PersistByteString` being returned for string-like MySQL values; persistent-mysql itself had several runtime errors that needed to be fixed because of this patch. High-level code dealing purely with `PersistEntities` should be unaffected.
+
 ## 2.2
 
 * Update to persistent 2.2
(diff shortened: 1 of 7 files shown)


MaxGabriel:
I realized this also applies to the other string types (not just `TEXT` and `BLOB`) and updated the PR accordingly.

creichert:
i'm not familiar with MySQL specifics but this looks good overall. Are you going to bump the major version?

MaxGabriel:
@creichert Yeah this probably should have a major version bump. But LTS 3 just started; should we wait awhile to merge this? I can make a PR to `yesod-text-markdown` to accept either `PersistText` or `PersistByteString` to fix the immediate issues people are reporting on StackOverflow.

gregwebs:
I think it should be merged whenever it is ready. We can create maintenance branches as needed.

MaxGabriel:
Forgot about this for a few days; I updated the changelog and cabal file. Will merge it in shortly.

MaxGabriel:
Thanks for fixing build times @snoyberg; these tests were failing on > 50 minute builds before.

PR #453, opened 2015-08-16
Remove ghc 7.10.2 from allowed_failures

creichert:


creichert:
Making sure this works with travis for good measure.

PR #458, opened 2015-08-23
Improved documentation for ||.

ecognium:
Added some examples on how to use `||.` and explained how to create a filter for `(A AND B AND (C OR D))`  using `++`. The comments has `&&.` in case people look for it, they will see this comment.

### DIFF
--- a/persistent/Database/Persist.hs
+++ b/persistent/Database/Persist.hs
@@ -67,7 +67,12 @@ f /<-. a = Filter f (Right a) NotIn
 
 infixl 3 ||.
 (||.) :: forall v. [Filter v] -> [Filter v] -> [Filter v]
--- | the OR of two lists of filters
+-- | the OR of two lists of filters. For example: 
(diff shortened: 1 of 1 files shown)


creichert:
Aweome, Thanks! One nit, with the haddocks I believe it only works with `-- | ` on the first line of the top-level declaration. https://www.haskell.org/haddock/doc/html/markup.html#idm140354810917952.

ecognium:
Hi @creichert: I have never done this before so thanks for pointing out the markup syntax. It looks like I need to do `{- | -}`.  What should I do to update it?

creichert:
No problem, I would simply remove the `|` from all but the first line:

```
-- | the OR of two lists of filters. For example: 
-- selectList([PersonAge >. 25, PersonAge <. 30] ||. [PersonIncome >. 15000, PersonIncome <. 25000]) []
-- will filter records where a person's age is between (25 and 30) OR a person's income is between (15,000 and 25000). 
... (Rest of the docs here)
```

Then, commit and push and it will just show up in this pull request.

ecognium:
Got it. Thanks. I just recommitted it.

creichert:
Perfect, thanks!

PR #461, opened 2015-08-28
Travis: more matrix entries, smaller individual builds (fixes #460)

### LINKED ISSUE #460: Failing Travis builds once SQLite is enabled
* [ ] There's a weird bug (seemingly in the compiler itself) with GHC 7.6.3 complaining about ``ghc: dist/build/Database/Persist/Sqlite.o: unknown symbol `sqlite3_changes'``
* [ ] The test suite is right on the border of taking 50 minutes

Proposal:

* Include the database backends as part of the matrix, so each test run only tests one backend
* Have a "lint" test with no backend that confirms that all of the other packages (not persistent-test) compile and pass their tests
* Disable the sqlite/GHC 7.6.3 combination, or possibly remove 7.6.3 altogether
* Try to enable caching. The only thing that may cause us trouble there is the zookeeper config changes requiring `sudo`
* Possibly: build using stack instead of cabal to increase cache hits

I'll get started on some of this on a separate branch. Feedback from others who have touched the Travis config more recently is welcome.

snoyberg:


creichert:
That's a lot of builds, looks great!

PR #463, opened 2015-09-02
Handle composite keys in RawSql (Entity a)

### LINKED ISSUE #457: Persistent 2.2 : rawSql "??" attempts to read with "id", although that field doesn't exist
This seems to be similar to issue #416 . The Sqlite backend does not add an id column to the database, but the generated template haskell code still defines an id column that is used by rawSql.

Is a column named id compulsory?
```
EmailRaw
    uidl T.Text
    rawMessage BS.ByteString
    Primary uidl
    deriving Show
EmailLinks
    uidl T.Text
    httpLink T.Text
    Primary uidl httpLink
    deriving Show
]
```
The following select:
```
dbEmailNoLinks :: MonadIO m => ReaderT SqlBackend m [Entity EmailRaw]
dbEmailNoLinks = do
  rawEmails :: [Entity EmailRaw] <- rawSql
      "select ?? from email_raw  \
         \where not exists (select * from email_links where uidl = email_raw.uidl)"  []
  return rawEmails
```
results in the error:

λ> dbWriteLinks
*** Exception: SQLite3 returned ErrorError while attempting to perform prepare "select \"email_raw\".\"id\", \"email_raw\".\"uidl\", \"email_raw\".\"date\", \"email_raw\".\"subject\", \"email_raw\".\"from\", \"email_raw\".\"to\", \"email_raw\".\"raw_message\" from email_raw where not exists (select * from email_links where uidl = email_raw.uidl)": no such column: email_raw.id

mkscrg:
This is an attempt to resolve #457 by teaching `RawSql` how to handle composite keys (i.e. `Primary`). The crux of it is [`keyFields`](https://github.com/yesodweb/persistent/pull/463/files#diff-360db80d5d02f984c61fb1c708497d3aR86), which (intends to!) find the "true" `FieldDef`s for an entity's key.

### DIFF
--- a/persistent-test/CompositeTest.hs
+++ b/persistent-test/CompositeTest.hs
@@ -269,11 +269,15 @@ specs = describe "composite" $
 
       newp2 <- get kp2
       newp2 @== Just p2
-    it "rawSql instance" $ db $ do
+    it "RawSql Key instance" $ db $ do
       key <- insert p1
       keyFromRaw <- rawSql "SELECT name, name2, age FROM test_parent LIMIT 1" []
       [key] @== keyFromRaw
-      
+    it "RawSql Entity instance" $ db $ do
+      key <- insert p1
+      newp1 <- rawSql "SELECT ?? FROM test_parent LIMIT 1" []
+      [Entity key p1] @== newp1
+
 #endif
 
 matchK :: (PersistField a, PersistEntity record) => Key record -> Either Text a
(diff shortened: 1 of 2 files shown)


gregwebs:
Thanks for the test case and code! I am starting a refactoring of it over on #464

mkscrg:
Excellent! Thanks @gregwebs

PR #464, opened 2015-09-02
[WIP] Rawsql entity composite

gregwebs:

PR #466, opened 2015-09-03
Fix adding a foreign key reference to a custom primary key for MySQL

### LINKED ISSUE #421: MySQL: Trying to create a Foreign Key Constraint to mythical field
I have the following structure:

    Tweet
        tweetId Int
        retweetCount Int
        favoriteCount Int
        statusText Text sqltype=varchar(170)
        Primary tweetId
        UniqueTweetId tweetId
        deriving Show
    TweetUrl
        tweetId TweetId
        tweetUrl URL sqltype=varchar(255)
        finalUrl URL Maybe sqltype=varchar(255)
        UniqueTweetIdTweetUrl tweetId tweetUrl
        deriving Show

When running this I get the following error continuously:

    Migrating: ALTER TABLE `tweet_url` ADD CONSTRAINT `tweet_url_tweet_id_fkey` FOREIGN KEY(`tweet_id`) REFERENCES `tweet`(`id`)
    Left (ConnectionError {errFunction = "query", errNumber = 1215, errMessage = "Cannot add foreign key constraint"})

Clearly it should not be creating a link to the mythical non existing field `tweet.id`. My desire is to link against the existing `tweet.tweet_id` field however using `TweetTweetId` does not yield success ending up in a GHC stage restriction.

I would appreciate advice on how to achieve this table structure, but I also think that it trying to create a foreign key to a non existing field is a bug.

Thanks

MaxGabriel:
* Closes #421

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -378,15 +378,15 @@ findMaxLenOfColumn allDefs name col =
            maxLenAttr <- find ((T.isPrefixOf "maxlen=") . T.toLower) (fieldAttrs fieldDef)
            readMaybe . T.unpack . T.drop 7 $ maxLenAttr
 
--- | Helper for 'AddRefence' that finds out the 'entityId'.
+-- | Helper for 'AddReference' that finds out the which primary key columns to reference.
 addReference :: [EntityDef] -> DBName -> DBName -> DBName -> AlterColumn
-addReference allDefs fkeyname reftable cname = AddReference reftable fkeyname [cname] [id_] 
+addReference allDefs fkeyname reftable cname = AddReference reftable fkeyname [cname] referencedColumns
     where
-      id_ = maybe (error $ "Could not find ID of entity " ++ show reftable
-                         ++ " (allDefs = " ++ show allDefs ++ ")")
-                  id $ do
-                    entDef <- find ((== reftable) . entityDB) allDefs
-                    return (fieldDB $ entityId entDef)
+      referencedColumns = maybe (error $ "Could not find ID of entity " ++ show reftable
+                                  ++ " (allDefs = " ++ show allDefs ++ ")")
+                                id $ do
+                                  entDef <- find ((== reftable) . entityDB) allDefs
+                                  return $ map fieldDB $ entityKeyFields entDef
 
 data AlterColumn = Change Column
                  | Add' Column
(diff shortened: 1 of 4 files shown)


persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -378,15 +378,20 @@ findMaxLenOfColumn allDefs name col =
            maxLenAttr <- find ((T.isPrefixOf "maxlen=") . T.toLower) (fieldAttrs fieldDef)
            readMaybe . T.unpack . T.drop 7 $ maxLenAttr
 
--- | Helper for 'AddRefence' that finds out the 'entityId'.
+dbIdColumns2 :: EntityDef -> [DBName]
+dbIdColumns2 t = case entityPrimary t of
+    Just pdef -> map fieldDB (compositeFields pdef)
+    Nothing   -> [fieldDB (entityId t)]
```
MaxGabriel: @gregwebs I like the `entityKeyFields` function you added in #464 better than this one. Thinking I'll just wait until you merge than, then use `map fieldDB $ entityKeyFields entDef` in my own code.

gregwebs:
looks like a great change. Will merge my PR now

MaxGabriel:
Rebased off master and used Greg's new function; ready for final review.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -378,15 +378,15 @@ findMaxLenOfColumn allDefs name col =
...
-      id_ = maybe (error $ "Could not find ID of entity " ++ show reftable
-                         ++ " (allDefs = " ++ show allDefs ++ ")")
-                  id $ do
-                    entDef <- find ((== reftable) . entityDB) allDefs
-                    return (fieldDB $ entityId entDef)
+      referencedColumns = maybe (error $ "Could not find ID of entity " ++ show reftable
+                                  ++ " (allDefs = " ++ show allDefs ++ ")")
+                                id $ do
+                                  entDef <- find ((== reftable) . entityDB) allDefs
+                                  return $ map fieldDB $ entityKeyFields entDef
```
MaxGabriel: So, this only adds a reference to primary key columns, but based off the Postgresql implementation that's what this function is expected to do I guess?

gregwebs: what else did you want it to do?

MaxGabriel: I didn't know if you were supposed to be able to add foreign keys from one column to another, non-primary key column.

gregwebs: ahh, I see. I was never involved much with the migration code and pretty much just made it pass the tests when adding the more extensible key abilities.

MaxGabriel: Hm, I think adding foreign keys to non-primary-key columns is probably more of a niche use case, anyway.

PR #470, opened 2015-09-06
mockMigration for Sqlite backend

psibi:
Related to #274


gregwebs:
@snoyberg was the one reviewing this previously. I am not a sqlite user. However, I am not worried about merging this since it is a new addition to the API that won't break existing stuff.

PR #477, opened 2015-09-14
Fix persistent-postgresql's treatment of NULLs inside arrays

meteficha:
Trying to fetch an arrays with a `NULL` value would not work.  The problem was that it wasn't obvious how `persistent-postgresql` created `PersistentNull` values, so when the work on arrays got merged 1.5 year ago no one noticed it lacked `NULL` treatment code.

This PR adds a previously-failing test and the fix.


gregwebs:
Looks great. Just needs a #ifdef around the test case

meteficha:
Done @ 79be9a2, tests still running.

PR #478, opened 2015-09-15
Improve failed migration error message

mitchellwrosen:
It would be nice to see the entire migration when an unsafe statement is found, using `runMigration` and friends.

Before:
```
Database migration: manual intervention required.
The following actions are considered unsafe:

    DROP TABLE "player";
```

After:
```
Database migration: manual intervention required.
The unsafe actions are prefixed by '***' below:

    CREATE TEMP TABLE "player_backup"("id" INTEGER PRIMARY KEY,"device_idee" BLOB NOT NULL);
    INSERT INTO "player_backup"("id") SELECT "id" FROM "player";
*** DROP TABLE "player";
    CREATE TABLE "player"("id" INTEGER PRIMARY KEY,"device_idee" BLOB NOT NULL);
    INSERT INTO "player" SELECT "id","device_idee" FROM "player_backup";
    DROP TABLE "player_backup";
```

### DIFF
--- a/persistent/Database/Persist/Sql/Migration.hs
+++ b/persistent/Database/Persist/Sql/Migration.hs
@@ -30,8 +30,6 @@ import Database.Persist.Types
 
 allSql :: CautiousMigration -> [Sql]
 allSql = map snd
-unsafeSql :: CautiousMigration -> [Sql]
-unsafeSql = allSql . filter fst
 safeSql :: CautiousMigration -> [Sql]
 safeSql = allSql . filter (not . fst)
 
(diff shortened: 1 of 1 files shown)


meteficha:
Looks good to me!  I'd change `case filter fst mig` into `if any fst mig` but I reckon it's a matter of style only.

mitchellwrosen:
Sure. Updated.

mitchellwrosen:
`cabal update` failed three times? Am I reading that report correctly?

gregwebs:
Great error message. Looks like you got hit by some build flakiness

Issue #465, opened 2015-09-03
Not ready for aeson 0.10


jcristovao:
Aeson 0.10, while still not released on hackage, will introduce some changes:

https://github.com/bos/aeson/blob/master/changelog.md

And it seems to break persistent-template tests:

    Test suite test: RUNNING...
    
    JSON serialization
      to/from is idempotent FAILED [1]
      decode
    JSON serialization for Entity
      to/from is idempotent FAILED [2]
      decode
    lens operations
    
    Failures:
    
      1) JSON serialization to/from is idempotent
           Falsifiable (after 1 test): 
           Person {personName = "", personAge = Just 0, personAddress = Address {addressStreet = "", addressCity = "", addressZip = Nothing}}
    
      2) JSON serialization for Entity to/from is idempotent
           Falsifiable (after 1 test): 
           Person {personName = "", personAge = Just 0, personAddress = Address {addressStreet = "", addressCity = "", addressZip = Nothing}}
    
    Randomized with seed 934331499

Although the output is not that clear on what is the expected vs what it got...

Note: this is a forward going issue - the new version of aeson is not alllowed by the current cabal file.

MaxGabriel:
I reproduced with this `stack.yaml` file:

```yaml
flags: {}
packages:
- '.'
- location:
    git: https://github.com/bos/aeson.git
    commit: f6006393a663dd1dc391aabf304ef3181fbdead5
extra-deps: ["attoparsec-0.13.0.1"]
resolver: lts-3.0
```

MaxGabriel:
I haven't debugged many QuickCheck tests before, but this would let you see the original value vs what was parsed:

```haskell
main :: IO ()
main = hspec $ do
    describe "JSON serialization" $ do
        prop "to/from is idempotent" $ \person ->
            let cycled = eitherDecode (encode person)
                original = Right (person :: Person)
            in (traceShow ("Original = " ++ show original ++ " cycled = " ++ show cycled) cycled) == original
```

Example output:

```
    "Original = Right (Person {personName = \"\", personAge = Nothing, personAddress = Address {addressStreet = \"\", addressCity = \"\", addressZip = Nothing}}) cycled = Left \"Error in $.age: failed to parse field age: expected Int, encountered Null\""
```

gregwebs:
Seems like a bug in Aeson itself, perhaps in the TH code? A more simplified test case might help

MaxGabriel:
Slightly more simplified test case:

```haskell
[...]

PR #480, opened 2015-09-21
bump aeson dep

bitemyapp:


MaxGabriel:
I don't think we can bump this until #465 is resolved

Also see:
* bos/aeson#83
* bos/aeson#287

bitemyapp:
@MaxGabriel fair 'nuff.

PR #481, opened 2015-09-23
persistent-sqlite.cabal: fix test/sanity.hs test

trofi:
ghc detects build failure as:
  Preprocessing executable 'sanity' for persistent-sqlite-2.2.1...
  [1 of 1] Compiling Main             ( test/sanity.hs, dist/build/sanity/sanity-tmp/Main.dyn_o )

  test/sanity.hs:8:8:
    No instance for (monad-logger-0.3.13.2:Control.Monad.Logger.MonadLogger
                       IO)
      arising from a use of ‘withSqliteConn’
    In the expression: withSqliteConn ":memory:"
    In the expression: withSqliteConn ":memory:" $ const $ return ()
    In an equation for ‘main’:
        main = withSqliteConn ":memory:" $ const $ return ()

Reported-by: Nick Bowler
Gentoo-bug: https://bugs.gentoo.org/show_bug.cgi?id=561274
Signed-off-by: Sergei Trofimovich <siarheit@google.com>

### DIFF
--- a/persistent-sqlite/persistent-sqlite.cabal
+++ b/persistent-sqlite/persistent-sqlite.cabal
@@ -74,4 +74,4 @@ executable sanity
         buildable: False
     main-is: sanity.hs
     hs-source-dirs: test
-    build-depends: base, persistent-sqlite
+    build-depends: base, persistent-sqlite, monad-logger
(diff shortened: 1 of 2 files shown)


gregwebs:
thanks!

PR #487, opened 2015-10-03
Changing id field type for MYSQL

nakaji-dayo:
["Change table Id field type"](https://github.com/yesodweb/persistent/wiki/Persistent-entity-syntax#change-tablecollection-id-field-name-andor-type-persistent--21) with MYSQL.

Since ```default=``` can not be used yet, need to use ```insertKey```.


gregwebs:
Thank you!

PR #489, opened 2015-10-08
Database/Persist/MySQL.hs: fix haddock markup

trofi:
haddock does not allow sticking docstrings to
unnamed constructor fields:
    Database/Persist/MySQL.hs:419:21: parse error on input ‘DBName’

Signed-off-by: Sergei Trofimovich <siarheit@google.com>


snoyberg:
Thanks!

PR #490, opened 2015-10-09
Support http-api-data

fizruk:
This adds `ToHttpApiData` and `FromHttpApiData` instances and replaces `PathPieces` instance bodies with `toUrlPiece` and `parseUrlPieceMaybe`.

From user's POV this changes instance for `CheckMark` to encode in lower-case and parse case-insensitively.

### DIFF
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -132,8 +132,7 @@ import qualified Data.ByteString as BS
 import qualified Data.Text.Encoding as E
 import qualified Data.Serialize as Serialize
 import Web.PathPieces (PathPiece(..))
-import Web.HttpApiData (ToHttpApiData(..), FromHttpApiData(..))
-import Web.HttpApiData.Internal (parseUrlPieceWithPrefix, readEitherTextData, parseUrlPieceMaybe)
+import Web.HttpApiData (ToHttpApiData(..), FromHttpApiData(..), parseUrlPieceMaybe, parseUrlPieceWithPrefix, readTextData)
 import Data.Conduit
 import Control.Monad.IO.Class (liftIO)
 import Data.Aeson (Value (Number), (.:), (.:?), (.!=), FromJSON(..), ToJSON(..), withText, withObject)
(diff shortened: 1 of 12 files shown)


gregwebs:
Thanks! It just needs a few changes for older versions of GHC

fizruk:
@gregwebs you have ideas about GHC 7.6 builds? I am not sure what happens there.

gregwebs:
7.6 was already failing, so don't worry about that. I think cabal is just failing to find an install plan for 7.6 now.
My plan right now is to wait for a 7.6 user to complain that persistent is not installing for them before delving into it.

fizruk:
Maybe you should place GHC 7.6 builds under `allow_failures` then?

gregwebs:
good call. trying that in #491

fizruk:
Is this PR good for merge?

gregwebs:
yeah, I will merge and release it after fixing the Travis build. Thanks a lot!
You might want to put an upper bound on http-api-data, but I will trust your judgement there.

fizruk:
@gregwebs I started to forget about bounds after I've switched to `stack`, sorry :)

gregwebs:
I pushed your branch with a merge to master, will merge the branch after that passes: https://travis-ci.org/yesodweb/persistent/builds/84593423

persistent-mongoDB/Database/Persist/MongoDB.hs
```diff
@@ -131,7 +131,9 @@ import Data.Text (Text)
 import qualified Data.ByteString as BS
 import qualified Data.Text.Encoding as E
 import qualified Data.Serialize as Serialize
-import Web.PathPieces (PathPiece (..))
+import Web.PathPieces (PathPiece(..))
+import Web.HttpApiData (ToHttpApiData(..), FromHttpApiData(..))
+import Web.HttpApiData.Internal (parseUrlPieceWithPrefix, readEitherTextData, parseUrlPieceMaybe)
```
gregwebs: Should this module be `Internal` if it is used by all the persistent libraries? Internal generally means something that should be avoided if possible and not depended upon by a library. A more usual name would be `Utils`, but I don't like generic names like that and tend to look for something more like `Web.HttpApiData.Parse`

fizruk: Thanks for bringing this up! For the first version of `http-api-data` this was intentional, since I didn't know which helpers I should export and by what name (I do not particularly like current naming).
I have opened fizruk/http-api-data#18 to discuss this.

fizruk:
I have updated `http-api-data` dependency. Now `persistent` does not use `Web.HttpApiData.Internal`.

gregwebs:
great! I will

- [ ] Add ChangeLog entries
- [ ] bump versions, push tags, and release to hakage

fizruk:
@gregwebs can I do anything to speed this up?

gregwebs:
Nope. I created #494 for this

fizruk:
@gregwebs Great, thanks!

PR #494, opened 2015-10-19
version bumps and ChangeLogs for http-api-data

gregwebs:


gregwebs:
Ok, most are uploaded. @snoyberg do you have access to upload persistent-redis or persistent-zookeeper?

Issue #390, opened 2015-04-28
check for permissions before starting migration


snoyberg:
See this thread for details: https://groups.google.com/d/msg/yesodweb/txpHU5Xi_-E/h_do1trsLsQJ

Goal: run a query along the following lines to ensure we have permissions to view all information on a schema before starting a migration to ensure more usable error messages:

```sql
SELECT grantee, privilege_type FROM information_schema.role_table_grants WHERE table_name = ? AND grantee = CURRENT_USER;
```

PR #501, opened 2015-11-09
Change the query for checking existence of Table

psibi:
The old query doesn't work in the scenario where the table is present
but under the ownership of a different user. This query fixes it.
Also it improves the error message related to issue #390.

Instead of an error message like this:

```
SqlError {sqlState = "42P07", sqlExecStatus = FatalError, sqlErrorMsg =
"relation \"person\" already exists", sqlErrorDetail = "",
sqlErrorHint = ""}
```

You get this:

```
SqlError {sqlState = "42501", sqlExecStatus = FatalError, sqlErrorMsg =
"must be owner of relation person", sqlErrorDetail = "", sqlErrorHint =
""}```

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -406,7 +406,8 @@ doesTableExist getter (DBName name) = do
     stmt <- getter sql
     with (stmtQuery stmt vals) ($$ start)
   where
-    sql = "SELECT COUNT(*) FROM information_schema.tables WHERE table_name=?"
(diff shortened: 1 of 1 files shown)


psibi:
CC: @gregwebs and others

I will merge this if there is no one against it.

gregwebs:
I am not familiar with Postgres ownership, but I like the new error message!

psibi:
@gregwebs Who else can review this for PostgreSQL then ?

gregwebs:
I released it to hackage. Thanks!

psibi:
Thanks!

PR #510, opened 2015-11-14
add changes to support int8 int16 int32 int64 types in oracle for odbc

gbwey:
Hi,

Can you add these changes to support Oracle when using persistent-odbc?

Thanks,
Grant


gregwebs:
thanks!

PR #513, opened 2015-12-02
Add convenience function insertOrGet to PersistUnique.

vandenoever:


gregwebs:
Sorry for the delay in tending to this.

How about naming it `getOrInsertBy` ?
Does it make sense to define it as `either entityKey id . insertBy` ?

vandenoever:
`either entityKey id . insertBy` is less efficient than what I propose because `insertBy` is more complicated than the proposed implementation.
'getOrInsertBy' is a fine name.

gregwebs:
This is merged now. Thanks!

PR #523, opened 2015-12-25
Use aeson-compat >=0.3.0.0

phadej:

PR #524, opened 2015-12-26
Update to use aeson-compat >= 0.3

felixonmars:
Data.Aeson.Compat has been moved into aeson-compat, and no longer available in aeson-extra since 0.3.0.0. Let's switch to the new package.


felixonmars:
Looks like a duplicate of #523. Sorry for the noise!

gregwebs:
Thanks for tending to persistent!

PR #532, opened 2016-01-22
Add projections to all mongoDB queries

mattjbray:
Here is an attempt to apply projections to all of the queries in the mongoDB backend, so that only the fields defined on the `PersistentEntity` are fetched from the database.

I'm pretty new to persistent so please review carefully :)

Specifically:

* Behaviour of `repsert`, `replace`, `update`, `upsert`. I'm assuming that currently, if there are extra fields in the database document that persistent doesn't know about, those fields will be lost after the document is replaced.  Therefore, adding projections to the queries doesn't change anything.
* Behaviour of `update`, `updateGet`, `updateWhere`. These functions all use the `updatesToDoc` function which makes the changes using `$set`, etc and so adding projections won't result in data loss.
* There is a TODO [here](https://github.com/yesodweb/persistent/blob/5a5590334500a16a93ea60f3cc9bb822cb4af801/persistent-mongoDB/Database/Persist/MongoDB.hs#L994) regarding projections - I am not sure if this change resolves it. I haven't tried to get my head around `orderPersistValues` yet.

If these issues aren't obvious to the maintainers I'll try and resolve/verify them myself.

### DIFF
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -376,7 +376,7 @@ runMongoDBPoolDef = runMongoDBPool defaultAccessMode
 
 queryByKey :: (PersistEntity record, PersistEntityBackend record ~ DB.MongoContext)
            => Key record -> DB.Query
-queryByKey k = DB.select (keyToMongoDoc k) (collectionNameFromKey k)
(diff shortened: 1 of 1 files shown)


gregwebs:
Thanks! Sorry I have been away. This looks like it does in fact fix the TODO

Issue #518, opened 2015-12-14
split PersistStore to distinguish between read-only and read-write access to database


vandenoever:
PersistStore can read and write to a database. Some use-cases can be safer with a read-only handle to the database. Additionally, a backend like Sqlite can benefit in it's concurrently from knowing if an access is for reading or writing. There can be many concurrent readers, but only one concurrent writer.

A solution could be to split up PersistStore into PersistStore and PersistReadOnlyStore.

This distinction makes it possible to make a wrapper around Pool SqlBackend with a RWL.RWLock. Such a lock makes it possible to access sqlite without danger of a busy timeout.

Alternatively, PRAGMA busy_timeout can be set to a high value.

gregwebs:
This is also useful in other dbs for read slaves. I am in favor of this change. Want to take a try at it? Backwards compatibility should be maintained with type PersistStore = (PersistStoreRead, PersistStoreWrite)

vandenoever:
I'm a relative Haskell newbie, but I might have a go over the holidays. It seems something I could do albeit leading to a large patch.

gregwebs:
Cool! Splitting up the typeclass should be easy enough. Properly using the correct connection is where things get interesting.  @pseudonom has done this with an extra layer but might be interested in driving some of this down into persistent itself.

pseudonom:
Yup. I haven't pushed it upstream yet because it seems like it's one point in a pretty big design space. But, if you all like this approach, I can certainly work on a PR:

```haskell
{-# LANGUAGE RankNTypes #-}
{-# LANGUAGE NoImplicitPrelude #-}

module SplitDB
  ( ReadSqlT, WriteSqlT, readToWrite
  , HasWritePool(..), HasReadPool(..), runReadDB, runWriteDB
  , select, selectFirst, update, delete, insert, insertKey, get) where

import ClassyPrelude

import Control.Monad.Trans.Class (MonadTrans)
import Control.Monad.Trans.Control (MonadBaseControl)
import Database.Esqueleto hiding (select, update, delete, insert, insertKey, isNothing, get)
import qualified Database.Esqueleto as E
import qualified Database.Esqueleto.Internal.Sql as EI
import qualified Database.Persist.Class as P
import qualified Database.Persist.Types as P

newtype ReadSqlT m a
  = ReadSqlT { runRead :: SqlPersistT m a }
  deriving (Functor, Applicative, Monad, MonadIO, MonadTrans, MonadThrow)
[...]

PR #534, opened 2016-02-06
Split persistent into read and write classes

pseudonom:
As discussed in https://github.com/yesodweb/persistent/issues/518

### DIFF
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -173,7 +173,8 @@ lookupEnv key = do
     return $ lookup key env
 #endif
 
-instance HasPersistBackend DB.MongoContext DB.MongoContext where
+instance HasPersistBackend DB.MongoContext where
+    type BaseBackend DB.MongoContext = DB.MongoContext
     persistBackend = id
 
 recordTypeFromKey :: Key record -> record
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -540,10 +540,11 @@ instance FromJSON (BackendKey DB.MongoContext) where
 
 -- | older versions versions of haddock (like that on hackage) do not show that this defines
 -- @BackendKey DB.MongoContext = MongoKey { unMongoKey :: DB.ObjectId }@
-instance PersistStore DB.MongoContext where
+instance PersistCore DB.MongoContext where
     newtype BackendKey DB.MongoContext = MongoKey { unMongoKey :: DB.ObjectId }
         deriving (Show, Read, Eq, Ord, PersistField)
 
+instance PersistStoreWrite DB.MongoContext where
     insert record = DB.insert (collectionName record) (toInsertDoc record)
                 >>= keyFrom_idEx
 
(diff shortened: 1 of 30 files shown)


pseudonom:
I assume the next step would be to newtype `SqlBackend` as `ReadSqlBackend` and `WriteSqlBackend`?

erikd:
I'd probably name them `SqlReadBackend` and `SqlWriteBackend`.

pseudonom:
Example of use in https://github.com/pseudonom/splitdb/commit/cf52fd4c94123c2094dd54851742077e086addd2.

Not sure what's going on with Travis CI.

pseudonom:
Can anyone give me hints on Travis? This definitely builds locally and almost all the tests pass.

snoyberg:
Looks like it's because aeson-compat was removed from the test suite dependencies:

https://github.com/yesodweb/persistent/pull/534/files#diff-777410b9e6ac65f07fdf95f2f223f38bL135

pseudonom:
Huh, I removed it because it seemed to be causing problems locally (it was, I think, overlapping with `aeson-extra`) I can try readding it tomorrow.

pseudonom:
Ah, looks like the underlying issue may have been that I was using `stack` and the desired development flow involves `cabal`?

snoyberg:
We test with cabal on Travis, which is probably pulling in newer versions
of dependencies than you're getting with Stack.

pseudonom:
Now it looks like there's a linker error for the new module (which I did add to `persistent`'s cabal file). Is that anything you've encountered before?

snoyberg:
You added it to persistent-test as well?

pseudonom:
Ah, I did not. Thanks, I'll try that tomorrow.

pseudonom:
Hm, some of the runs are timing out during compile of `EmbedTest` now...

gregwebs:
@pseudonom awesome!

I am wondering if there are any shorthands that make sense to add to the code base itself?
I am having a hard time coming up with them though. For example

```
type PersistEntityBaseBackend = (BaseBackend backend ~ PersistEntityBackend val, PersistEntity val)
```

Doesn't seem like a huge win yet.

If the only outstanding issue is getting Travis to build, we can definitely help you though it. Do we need to increase the timeout a bit? persistent may have been on the edge of timing out already.

pseudonom:
Yeah, no shorthands immediately jump out at me. Fortunately, shouldn't be too hard to add them later if we discover we want some?

I'm not sure if increasing the timeout would help. Travis is getting
```
[30 of 50] Compiling EmbedTest        ( EmbedTest.hs, dist/build/EmbedTest.o )
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```
but it doesn't take anything close to 10 minutes to compile `EmbedTest` locally. Is it possible that GHC is hanging in some way that's configuration dependent?

gregwebs:
It could be hitting memory limits now. We have seen cases of that in the past.

pseudonom:
Is bumping the memory limit something I can do?

pseudonom:
Any hints about what I can do here?

pseudonom:
Closing in favor of https://github.com/yesodweb/persistent/pull/546

PR #538, opened 2016-02-09
Determine type based on Length and Type from MySQL Schema

### LINKED ISSUE #535: TINYINT can be more than just Bool
I am working with a mySQL database that uses TINYINT length 2, but Persistent considers all TINYINT to be of type Bool. The problem is, TINYINT could be either Bool, Int8 or Word8.

It would be quite helpful if Persistent determined the actual type when reading it in from the database, not just what the database says is the type.

Skyfold:
This should resolve #535.

I only changed how TINYINT is treated. So if your database structure has a TINYINT of length 2 or 3 it will be treated as a PersistInt64, while TINYINT of length 1 will still be PersistBool. It is worth mentioning that MySQL treats Length as how many digits to display. So a length 1 TINYINT could contain more than 0 and 1. However, distinguishing between a TINYINT that is used a bool or 0-9 is not possible.

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -231,7 +231,7 @@ convertPV f = (f .) . MySQL.convert
 -- the type of the column.
 getGetter :: MySQLBase.Field -> Getter PersistValue
 getGetter field = go (MySQLBase.fieldType field) 
-                        (MySQLBase.fieldMaxLength field)
+                        (MySQLBase.fieldLength field)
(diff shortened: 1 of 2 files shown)


Skyfold:
It looks like the errors Travis is getting are either due to Travis timing out or something to do with the imports. I did not change the imports, but the master branch does not report any errors for Travis builds. Not sure why the Travis builds are stalling sometimes.

Skyfold:
Would you like me to change how I implemented my fix?

gregwebs:
Sorry I have been away. Thanks for the fix!

PR #539, opened 2016-02-10
Support aeson-0.11

phadej:


phadej:
I see this is done already.

PR #541, opened 2016-02-17
Document combinator (=.) along with examples.

psibi:
Adds documentation to basic query combinators with an example schema along with a sample data. I'm planning to add similar line of examples to other combinators as well. So if there is any concern/request/criticism, I would like to hear it in a very early stage! The major motivation for adding this is after my personal experience with haskell newbies who were finding this difficult to use.

CC: @snoyberg , @gregwebs

### DIFF
--- a/persistent/Database/Persist.hs
+++ b/persistent/Database/Persist.hs
@@ -70,7 +70,7 @@ infixr 3 =., +=., -=., *=., /=.
 -- updateAge = updateWhere [UserName ==. \"SPJ\" ] [UserAge =. 45]
 -- @
 --
--- Similar to `updateWhere` you can use other functions present in the module "Database.Persist.Class". Note that the first attribute of `updateWhere` is [`Filter` val] and second attribute is [`Update` val]. By comparing this with the type of `==.` and `=.`, you can see that they match up in the above usage.
+-- Similar to `updateWhere` which is shown in the above example you can use other functions present in the module "Database.Persist.Class". Note that the first parameter of `updateWhere` is [`Filter` val] and second parameter is [`Update` val]. By comparing this with the type of `==.` and `=.`, you can see that they match up in the above usage.
 --
 -- The above query when applied on "dataset-1", will produce this:
 -- 
--- a/persistent/Database/Persist.hs
+++ b/persistent/Database/Persist.hs
@@ -6,7 +6,7 @@ module Database.Persist
     ( module Database.Persist.Class
     , module Database.Persist.Types
 
-    -- * #dataset# Reference Schema & Dataset
+    -- * Reference Schema & Dataset
     -- |
     --
     -- All the combinators present here will be explained based on this schema:
@@ -18,8 +18,7 @@ module Database.Persist
     -- >     deriving Show
     -- > |]
     --
-    -- and this dataset. From here on we will call this as dataset-1
-    -- #foobar#
+    -- and this dataset. The examples below will refer to this as dataset-1.
     --
     -- > +-----+-----+-----+
     -- > |id   |name |age  |
@@ -64,7 +63,7 @@ infixr 3 =., +=., -=., *=., /=.
 
 -- | Assign a field a value.
 --
--- Example usage:
+-- === __Example usage__
 --
 -- @
 -- updateAge :: MonadIO m => ReaderT SqlBackend m ()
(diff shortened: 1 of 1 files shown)


persistent/Database/Persist.hs
```diff
@@ -40,6 +63,26 @@ infixr 3 =., +=., -=., *=., /=.
   forall v typ.  PersistField typ => EntityField v typ -> typ -> Update v
 
 -- | Assign a field a value.
+--
+-- Example usage:
```
MaxGabriel: I think examples are better when marked up as:
```
-- ==== __Example Usage__
```

this causes them to be made into collapsible/expandable sections in the Haddocks, so that you can have lots of examples without making scrolling through functions harder. The docs for `<$>` do this, [for example](https://hackage.haskell.org/package/base-4.8.2.0/docs/Control-Applicative.html).

image

persistent/Database/Persist.hs
```diff
@@ -6,7 +6,30 @@ module Database.Persist
...
+    -- |
+    --
+    -- All the combinators present here will be explained based on this schema:
+    --
+    -- > share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
+    -- > User
+    -- >     name String
+    -- >     age Int
+    -- >     deriving Show
+    -- > |]
```
MaxGabriel: Having a single schema for the examples is a great idea

persistent/Database/Persist.hs
```diff
@@ -6,7 +6,30 @@ module Database.Persist
...
+    --
+    -- and this dataset. From here on we will call this as dataset-1
+    -- #foobar#
```
MaxGabriel: `#foobar#` is maybe a leftover mistake?

persistent/Database/Persist.hs
```diff
@@ -6,7 +6,30 @@ module Database.Persist
...
+    -- All the combinators present here will be explained based on this schema:
+    --
+    -- > share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
+    -- > User
+    -- >     name String
+    -- >     age Int
+    -- >     deriving Show
+    -- > |]
+    --
+    -- and this dataset. From here on we will call this as dataset-1
```
MaxGabriel: > From here on we will call this as dataset-1

"call this as" doesn't sound correct, it'd be better to drop the "as". I think something more explicit would be even better:  "The examples below will refer to this as dataset-1"

MaxGabriel:
iirc there's sometimes initial confusion about these combinators, and these docs will help alot

psibi:
@MaxGabriel Thanks for the feedback. I have pushed new fixes on top of it.

persistent/Database/Persist.hs
psibi: @MaxGabriel This is a nice idea!

persistent/Database/Persist.hs
```diff
@@ -40,6 +62,26 @@ infixr 3 =., +=., -=., *=., /=.
...
 -- | Assign a field a value.
+--
+-- === __Example usage__
+--
+-- @
+-- updateAge :: MonadIO m => ReaderT SqlBackend m ()
+-- updateAge = updateWhere [UserName ==. \"SPJ\" ] [UserAge =. 45]
+-- @
+--
+-- Similar to `updateWhere` you can use other functions present in the module "Database.Persist.Class". Note that the first attribute of `updateWhere` is [`Filter` val] and second attribute is [`Update` val]. By comparing this with the type of `==.` and `=.`, you can see that they match up in the above usage.
```
MaxGabriel: > Similar to `updateWhere` you can use other functions present in the module "Database.Persist.Class".

I get what it means now, but I initially was confused by this sentence and didn't understood how it related to `=.`.

> Note that the first attribute of `updateWhere` is [`Filter` val] and second attribute is [`Update` val].

"attribute" should probably be "parameter" or "argument" here

psibi: > I get what it means now, but I initially was confused by this sentence and didn't understood how it related to =..

I rewrote it to make it more clear. Let me know if it is still confusing.

> "attribute" should probably be "parameter" or "argument" here

You are right. Have been writing too much HTML in my day job!

PR #542, opened 2016-02-21
Document remaining combinators in Persist.hs

psibi:


MaxGabriel:
This LGTM. It's a little hard to look at the new database table and see what's changed sometimes (like on the `[UserAge +=. 1]` query). It'd be slightly nicer if the change was highlighted or made more obvious, or if there was a link back to the original data (I think you were trying to do that with the #foobar# thing you were experimenting with before).

But that's just a minor complaint, this definitely improves the documentation. :+1:

psibi:
> I think you were trying to do that with the #foobar# thing you were experimenting with before

Yes. :) But I couldn't get it to work. It will be nice to have a link to the original table. I will give it a try again tonight. I feel highlighting won't help users because unless they remember the original data they won't know what has changed in the highlighted data: whether it was added, multiplied etc.

MaxGabriel:
Do you think something like this would help?

```
-- > +-----+-----+--------+
-- > |id   |name |age     |
-- > +-----+-----+--------+
-- > |1    |SPJ  |40 -> 45|
-- > +-----+-----+--------+
-- > |2    |Simon|41      |
-- > +-----+-----+--------+
```

```
-- > +-----+-----+-----------+
-- > |id   |name |age        |
-- > +-----+-----+-----------+
-- > |1    |SPJ  |40 × 2 = 80|
-- > +-----+-----+-----------+
-- > |2    |Simon|41         |
-- > +-----+-----+-----------+
```

MaxGabriel:
> But I couldn't get it to work. It will be nice to have a link to the original table.

I had trouble too, but I just figured it out. You create the link just like you did:

```
#dataset#
```

then link to it like this:

```
<#dataset my dataset>
```

psibi:
@MaxGabriel Thanks for figuring out the anchor linking!

Instead of `40 * 2 = 80`, does just showing `40 * 2` is a better idea ? Something like this:

```
-- > +-----+-----+-----------+
-- > |id   |name |age        |
-- > +-----+-----+-----------+
-- > |1    |SPJ  |40 × 2     |
-- > +-----+-----+-----------+
-- > |2    |Simon|41         |
-- > +-----+-----+-----------+
```

MaxGabriel:
I prefer `40 * 2 = 80`, I think it's more explicit, but I think either one works. I would still have something like `->` to handle setting 40 to 45 for example `40 -> 45`

psibi:
I have highlighted the changes like this `40 -> 80`. Now I can do `40 * 2 -> 80`, but since the combinator query is just on top of it, I think the user will be able to figure out why the data changed like that. My reason for not including `* 2` is that this makes the original data `40` more clear. Let me know if you think otherwise. Also the anchor links for the dataset are fixed, thanks!

PR #544, opened 2016-02-21
Possible fix for travis timeout

psibi:


MaxGabriel:
The MySQL output was 754 KB and over 10K lines—maybe adjust verbosity down a little?

psibi:
Was experimenting to see if this will solve the timeout issue, but this doesn't. According to the [cabal documentation](https://downloads.haskell.org/~ghc/7.0.3/docs/html/Cabal/builders.html), the default verbosity is 2. If the verbosity is little, the travis build just times out.

psibi:
Probably a proper fix should be done using [travis_wait](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received), but that didn't seem to work for me.

PR #550, opened 2016-03-20
Privilege check (Fix #390)

### LINKED ISSUE #390: check for permissions before starting migration

psibi:


snoyberg:
What is this providing over the previous approach of simply trying to perform the operation?

psibi:
@snoyberg This checks if the current user has proper privilege. If it doesn't, it exits with this message:

```
Exiting: You have insufficient privilege for migration.
```
Previously, it used to throw up an unhelpful error message. But after this [patch](https://github.com/yesodweb/persistent/pull/501), it used throws a much better error message IMO:

```
SqlError {sqlState = "42501", sqlExecStatus = FatalError, sqlErrorMsg =
"must be owner of relation person", sqlErrorDetail = "", sqlErrorHint =
""}```

Either way, I don't have a strong opinion on this patch. I thought this was the best way to close up the issue #390. :)

snoyberg:
My concern is that:

* The previous error message was a bit daunting, but still gives the information you need (you're not the owner)
* There's a real possibility that the permission check you've written may have corner cases that aren't handled correctly and will require maintenance going forward, especially for new PostgreSQL versions

psibi:
No issues, I will close this and the corresponding issue then.

PR #552, opened 2016-04-11
add a PersistRecordBackend type alias

gregwebs:
@pseudonom I am trying to combat the increased size of type signatures from the read/write split (not released to hackage yet). Do you think it would be good to use this alias everywhere?

### DIFF
--- a/persistent/Database/Persist/Class.hs
+++ b/persistent/Database/Persist/Class.hs
@@ -9,6 +9,7 @@ module Database.Persist.Class
     , PersistStoreRead (..)
     , PersistStoreWrite (..)
     , BaseBackend(..)
+    , PersistRecordBackend
     , getJust
     , belongsTo
     , belongsToJust
--- a/persistent/Database/Persist/Class/PersistStore.hs
+++ b/persistent/Database/Persist/Class/PersistStore.hs
@@ -5,6 +5,7 @@
 module Database.Persist.Class.PersistStore
     ( HasPersistBackend (..)
     , IsPersistBackend (..)
+    , PersistRecordBackend
     , liftPersist
     , PersistCore (..)
     , PersistStoreRead (..)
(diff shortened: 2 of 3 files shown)


snoyberg:
I'm on the fence about this:

* It definitely makes the signatures easier to read
* However, GHC type errors may break apart the constraints to their constituent parts, leading to user confusion
* Using the old empty-typeclass trick instead of constraint kinds can solve that, but always feels more heavyweight

Since the constraint kind approach doesn't have much of a downside and can be easily backed out without breaking backwards compatibility, I'm not at all opposed to giving it a test drive

gregwebs:
It would also be possible to use a more descriptive synonym if the `Persist` prefix is dropped: `EntityBaseBackend`.

Here is what an error looked like when I flipped the type variables:

```
    /home/greg/hs/y/persistent/persistent/Database/Persist/Class/PersistStore.hs:105:22:
        Could not deduce (BaseBackend backend
                          ~ PersistEntityBackend record)
        from the context (MonadIO m, PersistRecordBackend backend record)
          bound by the type signature for
                     insert_ :: (MonadIO m, PersistRecordBackend backend record) =>
                                record -> ReaderT backend m ()
          at Database/Persist/Class/PersistStore.hs:(103,16)-(104,45)
        Relevant bindings include
          record :: record
            (bound at Database/Persist/Class/PersistStore.hs:105:13)
          insert_ :: record -> ReaderT backend m ()
            (bound at Database/Persist/Class/PersistStore.hs:105:5)
        In the first argument of ‘(>>)’, namely ‘insert record’
        In the expression: insert record >> return ()
        In an equation for ‘insert_’:
            insert_ record = insert record >> return ()
```

pseudonom:
It seems like a readability improvement to me too. It also isn't the first/only place to incur the `ConstraintKind` cost, so that cost is incremental rather than totally new.

PR #554, opened 2016-04-13
persistent-template: Make it work with ghc-8.0

erikd:
Ghc 8.0 introduces some TH changes that are detailed here:

    https://ghc.haskell.org/trac/ghc/wiki/Migration/8.0#template-haskell-2.11.0.0

Adapting to these changes while still maintaining compatibility with
ghc-7.10 required some new CPP hackery, but it now compiles and passes
all tests with both ghc-7.10 and ghc-8.0.


erikd:
Just checked with ghc-7.8 as well. All good here.

snoyberg:
Awesome, thanks!

PR #555, opened 2016-04-13
Persist record backend

pseudonom:
Should fix the travis errors in https://github.com/yesodweb/persistent/pull/552.


gregwebs:
Thanks!

PR #561, opened 2016-05-16
Fix the problem caused by the upgrading of SQLite 3.12.1

uuhan:
ghc-mod failed to check haskell source file, which complains:
    
    unknown symbol `stat64'

    unknown symbol `fstat64'

    unknown symbol `lstat64'

### DIFF
--- a/persistent-sqlite/cbits/fix.c
+++ b/persistent-sqlite/cbits/fix.c
@@ -1,0 +1,11 @@
+int stat64(){
+    return 0;
+}
+
+int fstat64(){
+    return 0;
+}
+
+int lstat64(){
+    return 0;
+}
--- a/persistent-sqlite/persistent-sqlite.cabal
+++ b/persistent-sqlite/persistent-sqlite.cabal
@@ -46,7 +46,9 @@ library
         include-dirs: cbits
         cc-options:  -fPIC -std=c99
 
-    c-sources: cbits/config.c
+    c-sources: 
+        cbits/config.c
+        cbits/fix.c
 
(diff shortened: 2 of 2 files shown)


snoyberg:
This isn't the right way to fix this problem, we need ac45d80e28bf3252297291955ecdb9d4c8235884.

snoyberg:
BTW, the GHC issue is marked as resolved since GHC 7.8.1, so if you're still having an issue with this, I recommend commenting on:

https://ghc.haskell.org/trac/ghc/ticket/7072

There's no harm in reapplying the workaround, which I'm doing now.

snoyberg:
Can you test out master before I make a release?

uuhan:
@snoyberg 
I have read the [thread](https://ghc.haskell.org/trac/ghc/ticket/7072) ahead,  it seems not related to my circumstance.  It works well when I run my program through runhaskell method.
However, the master **solved** my state64 missing problem. I just tested and do the fix with  the persistent-sqlite-2.5 in hackage, not test the master. 
**persistent-sqlite-2.5.0.1 works well**.

snoyberg:
OK, new version 2.5.0.1 released.

borsboom:
@snoyberg when Stack is built with persistent-sqlite-2.5.0.1 on OS X, it exits with this error when using Docker integration:

```
SQLite3 returned ErrorCan'tOpen while attempting to perform open "/Users/manny/.stack/docker.db".
```

persistent-sqlite-2.5 does not exhibit this behaviour.

PR #565, opened 2016-05-30
Fix workaround for stat64 bug on OS X (fixes #564)

### LINKED ISSUE #564: persistent-sqlite-2.5.0.1 fails on Mac OS X
Opening a database fails with:
```
sqlite-test: SQLite3 returned ErrorCan'tOpen while attempting to perform open "_test.sqlite3".
```

Minimal reproducing program:
```
module Main where

import qualified Database.Sqlite as Sqlite

main :: IO ()
main = do
  _ <- Sqlite.open "_test.sqlite3"
  return ()
```

It looks like the persistent-sqlite tests don't trigger this issue because they use the `:memory:` connection which doesn't use the `stat` system call.

(previously mentioned in https://github.com/yesodweb/persistent/pull/561#issuecomment-222329087)

borsboom:


snoyberg:
Awesome, good work. I'll cut a release today

PR #566, opened 2016-05-31
travis: don't run zookeeper tests

gregwebs:


gregwebs:
This should speed up Travis a little bit

Issue #569, opened 2016-06-02
unhandled error for composite/non id primary keys


jwaldmann:
I am getting this error 
```
unhandled error for composite/non id primary keys pfilter=In persistList=[] pdef=CompositeDef {compositeFields = [FieldDef {fieldHaskell = HaskellName {unHaskellName = "studentId"}, fieldDB = DBName {unDBName = "SNr"}, fieldType = FTTypeCon Nothing "StudentId", fieldSqlType = SqlOther "SqlType unset for studentId", fieldAttrs = ["sql=SNr"], fieldStrict = True, fieldReference = NoReference},FieldDef {fieldHaskell = HaskellName {unHaskellName = "gruppeId"}, fieldDB = DBName {unDBName = "GNr"}, fieldType = FTTypeCon Nothing "GruppeId", fieldSqlType = SqlOther "SqlType unset for gruppeId", fieldAttrs = ["sql=GNr"], fieldStrict = True, fieldReference = NoReference}], compositeAttrs = []}
```
and I have no idea how to debug this. The error message does not even mention the table's name (but I can guess it from the names of columns - in this case)

model/table looks like
```
Einschreibung sql=stud_grp
  studentId         StudentId            sql=SNr
  gruppeId          GruppeId             sql=GNr
  Primary studentId gruppeId

mysql> describe stud_grp;
+-------+------------+------+-----+---------+-------+
| Field | Type       | Null | Key | Default | Extra |
+-------+------------+------+-----+---------+-------+
| SNr   | bigint(20) | NO   | PRI | NULL    |       |
| GNr   | bigint(20) | NO   | PRI | NULL    |       |
+-------+------------+------+-----+---------+-------+
```
and I confirmed that there are no duplicate rows.

gregwebs:
What is the query that you are running? It looks like you provided an empty list for an in query, can you confirm? I can change persistent to surface a better error message.

jwaldmann:
In this handler: https://gitlab.imn.htwk-leipzig.de/autotool/all/blob/master/yesod/Handler/Gruppen.hs#L15 

Indeed, occurence of this error depends on other data that's possibly combined in the query. The model is: student is associated with a group, group belongs to a lecture (vorlesung). Error occurs when there are no gruppen for a vorlesung.

gregwebs:
I would suggest producing a NonEmpty (I would like to change the API to require that) any time you do an In query. Just call toList on the NonEmpty. A better error message is committed to master now.

[...]

PR #570, opened 2016-06-02
provide a better error message for an empty list

gregwebs:
sorry for the whitespace deletion setting in my editor


gregwebs:
closes #569

PR #580, opened 2016-06-15
Fix #576

### LINKED ISSUE #576: doesTableExist causes problems when there are tables with the same name and different owners.
I have to work with my uni's database where each user has access to his/her own tables only. I run into problem when I wanted to create table with the name someone already used. I am able to create it manually by psql, but specifying entity in persistent causes errors because doesTableExist finds those tables.

For example there is table
tablename=my_table, tableowner=user1
I want to define model in yesod:
MyTable and I am user2, which should produce:
tablename=my_table, tableowner=user2. 
Now doesTableExist would find my_table created by user1, that I have no access to and which should be ignored. 
doesTableExist should use query that check if tableowner="databse user" that I have given in settings file.

psibi:


erikd:
I'd love to see a commit message that was a little more comprehensive than "Fix #576".

psibi:
Thanks, updated the commit message.

gregwebs:
Won't this cause a problem if the user is not the owner but has permission?

psibi:
@gregwebs Yeah, you are right. I didn't think that scenario. A better way would be to do a privilege check like this: https://github.com/yesodweb/persistent/pull/550 .

gregwebs:
I would be supportive of applying that to this case. The other approach would be to use this solution but guard it behind a configuration option

psibi:
Closing my PR as I feel this is not the right way to do it. Free feel to re-open if anybody disagrees.

PR #581, opened 2016-06-15
Support FKs to non-Int PK entities

pseudonom:
Currently, a declaration like

```haskell

import ClassPrelude.Yesod
import Data.UUID (UUID)

  Foo
    Id UUID
  Bar
    fooId FooId
|]
```
will produce bad entity definitions. e.g. The migration will fail with

```
SqlError {sqlState = "42804", sqlExecStatus = FatalError, sqlErrorMsg = "foreign key constraint \"bar_foo_id_fkey\" cannot be implemented", sqlErrorDetail = "Key columns \"foo_id\" and \"id\" are of incompatible types: bigint and uuid.", sqlErrorHint = ""}
```
.

This patch is is currently in the "works for me" stage, though it's pretty ugly. We have to do the fixing at the last minute because otherwise we end up referring to `FooId` before it's been defined.

### DIFF
--- a/persistent-template/Database/Persist/TH.hs
+++ b/persistent-template/Database/Persist/TH.hs
@@ -304,11 +304,12 @@ mkEntityDefSqlTypeExp emEntities entMap ent = EntityDefSqlTypeExp ent
 mkPersist :: MkPersistSettings -> [EntityDef] -> Q [Dec]
 mkPersist mps ents' = do
     x <- fmap mconcat $ mapM (persistFieldFromEntity mps) ents
-    y <- fmap mconcat $ mapM (mkEntity mps) ents
+    y <- fmap mconcat $ mapM (mkEntity entMap mps) ents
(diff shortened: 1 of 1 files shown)


gregwebs:
Hmm, there is a test case for a primary key that works now in PrimaryTest.hs. I think you can do something like this right now

```
Foo
  uuid UUID sql=id
  Primary uuid
Bar
  fooId FooId
```

pseudonom:
Oh, interesting. What are the intended use differences between `Id` and `Primary`?

gregwebs:
`Primary` was designed for specifying a multiple column primary key. I will look at this PR soon, because what you are trying to do should definitely be supported

gregwebs:
I added the test schema definition on this branch: https://github.com/yesodweb/persistent/pull/589

pseudonom:
Great, thanks!

PR #586, opened 2016-07-02
Add upsertBy

jsteiner:
`upsert` can only be used when there is a single uniqueness constraint.
Sometimes you want to upsert on models with more than one uniqueness
constraint. Hence, `upsertBy`.


gregwebs:
Looks good! I am adding a commit for MongoDB on top of this at this PR: https://github.com/yesodweb/persistent/pull/588

PR #587, opened 2016-07-02
Atomic upsert support for Postgresql

psibi:
I have created this PR for getting review.

Things that work now:

* Atomic upsert for postgresql work

Things that have to be done:

* Raise exception when there is more than one uniqueness constraint for an `EntityDef`.
* Fix other backend errors - Right now works only for postgresql

Question I have:

In what scenario will the field `uniqueFields` have more than one element in the list. Why it is modelled as a List data strcuture instead of a tuple like `(HaskellName, DBName)` ?

The `UniqueDef` type for reference:

``` haskell
data UniqueDef = UniqueDef
    { uniqueHaskell :: !HaskellName 
    , uniqueDBName  :: !DBName
    , uniqueFields  :: ![(HaskellName, DBName)] -- why list instead of single tuple ?
    , uniqueAttrs   :: ![Attr]
    }
    deriving (Show, Eq, Read, Ord)
```

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -215,11 +215,15 @@ upsertSql' ent vals updateVal = T.concat
                                 , " where "
                                 , wher
                                 , " returning ??"
-                      ]
+                                ]
     where
       wher = T.intercalate " AND " $ map singleCondition $ entityUniques ent
-      singleCondition :: UniqueDef ->  Text
-      singleCondition udef = escape (entityDB ent) <> "." <> T.concat (map escape (map snd $ uniqueFields udef)) <> " =?"
+
+      singleCondition :: UniqueDef -> Text
+      singleCondition udef = T.intercalate " AND " (map singleClause $ map snd (uniqueFields udef))
+
+      singleClause :: DBName -> Text
+      singleClause field = escape (entityDB ent) <> "." <> (escape field) <> " =?"
 
 -- | SQL for inserting multiple rows at once and returning their primary keys.
 insertManySql' :: EntityDef -> [[PersistValue]] -> InsertSqlResult
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE ViewPatterns #-}
 {-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE DeriveDataTypeable #-}
 
 -- | A postgresql backend for persistent.
 module Database.Persist.Postgresql
@@ -44,7 +45,7 @@ import Control.Monad.Trans.Resource
 import Control.Exception (throw)
 import Control.Monad.IO.Class (MonadIO (..))
 import Data.Data
-import Data.Typeable
+import Data.Typeable (Typeable)
 import Data.IORef
 import qualified Data.Map as Map
 import Data.Maybe
@@ -61,6 +62,7 @@ import qualified Data.IntMap as I
 import Data.ByteString (ByteString)
 import qualified Data.ByteString.Char8 as B8
 import qualified Data.Text as T
+import Data.Text.Read (rational)
 import qualified Data.Text.Encoding as T
 import qualified Data.Text.IO as T
 import qualified Blaze.ByteString.Builder.Char8 as BBB
@@ -77,6 +79,7 @@ import Data.Int (Int64)
 import Data.Monoid ((<>))
 import Data.Pool (Pool)
 import Data.Time (utc, localTimeToUTC)
+import Control.Exception (Exception, throwIO)
 
 -- | A @libpq@ connection string.  A simple example of connection
 -- string would be @\"host=localhost port=5432 user=test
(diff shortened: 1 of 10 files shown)


gregwebs:
For `UniqueDef` that should be for a unique constraint that spans multiple fields

gregwebs:
Looking great! Once you write a test case everything else will fall into place.

gregwebs:
This will need to be re-based to the latest master now.

psibi:
@gregwebs Any idea on why the tests are failing: https://travis-ci.org/yesodweb/persistent/jobs/145239853 ?

The tests seem to be failing for `insert` sql query. I know that the TRAVIS postgresql version is 9.3 and we need a minimum of 9.5 to make this upsert sql feature work. But even then, the tests should fail with a different error message. Here it seems to get caught in a increasing series of `select` and `insert` queries.

gregwebs:
Looks like Travis only supports 9.4 right now.

I just pushed your branch with a merge to master which gets rid of sql debug logging on travis. I will look a the error there
https://travis-ci.org/yesodweb/persistent/builds/145356673

psibi:
There seems to be some workaround now for bringing 9.5: https://github.com/travis-ci/travis-ci/issues/4264#issuecomment-155404724

gregwebs:
I would rather stick with what Travis supports than start installing things. It avoids slowing down the build. The test on Travis will be good to prove that things still work as expected on < 9.5

psibi:
Okay. I see two options there:
* Wait till Postgresql 9.5 support comes in Travis. And then merge this PR.
* Merge the PR but pass `connUpsertSql = Nothing` in Postgresql configuration, so that the native upsert isn't used yet.

What do you recommend ? (Note that this PR has still some minor work left, so it's not ready for merge yet.)

gregwebs:
The persistent code needs to do a version check and only do a 9.5 upsert if postgresql is at version 9.5

gregwebs:
This is a great change! Should just need a squash and rebase at this point, which can be done from github if we just want 1 commit

psibi:
I have made some minor changes. I think this is good to merge now but I would like another review.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -148,16 +149,38 @@ open' modConn cstr logFunc = do
...
+upsertFunction :: Double -> Maybe (EntityDef -> Text -> Text)
+upsertFunction version = if (version >= 9.5)
+                         then Just upsertSql'
+                         else Nothing
 
 -- | Generate a 'Connection' from a 'PG.Connection'
 openSimpleConn :: (IsSqlBackend backend) => LogFunc -> PG.Connection -> IO backend
 openSimpleConn logFunc conn = do
     smap <- newIORef $ Map.empty
+    serverVersion <- getServerVersion conn
```
gregwebs: If `getServerVersion` should always work, then it is best to throw an exception here. And the result should be `IO (Either String Double)`, where the error case shows the version.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -199,6 +222,32 @@ insertSql' ent vals =
...
+
+upsertSql' :: EntityDef -> Text -> Text
+upsertSql' ent updateVal = T.concat
+                           [ "insert into "
+                           , escape (entityDB ent)
+                           , "("
+                           , T.intercalate "," $ map (escape . fieldDB) $ entityFields ent
+                           , ") VALUES ("
+                           , T.intercalate "," $ map (const "?") (entityFields ent)
+                           , ") on conflict ("
```
gregwebs: Not sure if the rest of SQL is consistently using upper case.

persistent/Database/Persist/Types/Base.hs
```diff
@@ -218,8 +218,20 @@ toEmbedEntityDef ent = embDef
                         _ -> Nothing
                     }
 
+-- Type for storing the Uniqueness constraint in the Schema.
```
gregwebs: great docs, thanks!

persistent-postgresql/Database/Persist/Postgresql.hs
psibi: I'm throwing an exception now. But I'm letting the type of `getServerVersion` be `IO (Maybe Double)`. I don't find any advantage of it being the `Either` data type as the exception is thrown when we encounter `Left` data.

persistent-postgresql/Database/Persist/Postgresql.hs
psibi: Thanks for the feedback. I made it upper case.

persistent-postgresql/Database/Persist/Postgresql.hs
gregwebs: Where are you throwing an exception? It looks to me that if the version doesn't parse nothing will happen and a proper upsert will silently be unavailable. The point of using Either is so that the version string can be used in an error message.

gregwebs: oh, sorry, I see it is on a later commit. In general I like to have base functions that return Either instead of throw an exception, but this will work.

persistent/Database/Persist/Class/PersistUnique.hs
```diff
@@ -78,9 +78,6 @@ class (PersistUniqueRead backend, PersistStoreWrite backend) => PersistUniqueWri
            -- unique key)
            -> ReaderT backend m (Entity record)
            -- ^ the record in the database after the operation
-    upsert record updates = do
```
gregwebs: Does this need to be removed?

gregwebs:
Good to merge now? Do you want to do a rebase -i first? I see there is already a Changelog entry :)
I think it will be best to bump things to 2.6 since there is a change to the base persistent package affecting others.

persistent/Database/Persist/Class/PersistUnique.hs
psibi: No. I have added them back.

psibi:
I have rebased the last 10 commits. But I think you can squash the remaining one from the github web interface itself. Free feel to do it. I have also changed the Changelog entry to point to 2.6. Haven't bumped the cabal version to 2.6. Let me know, if you want me to do that.

psibi:
@gregwebs Is any work still pending in this PR ? Can this be mergd.

gregwebs:
AWESOME! I will let this mellow on master for a little bit and then release by Monday.

gregwebs:
This is on hackage now

psibi:
:+1: Thanks!

Issue #536, opened 2016-02-07
Primary key column cannot be nullable


rahulmutt:
I'm using the yesod-postrgres template. Suppose you have

config/models

<pre>
Test
  field1 Text
  field2 Text Maybe
  field3 Text
  Primary field1 field3
</pre>

Compiling this gives:

<pre>
Model.hs:13:7:
    Exception when trying to run compile-time code:
      primary key column cannot be nullable: "field3"
    Code: persistFileWith lowerCaseSettings "config/models"
    In the splice: $(persistFileWith lowerCaseSettings "config/models")
</pre>

Rearranging,
config/models

<pre>
Test
  field1 Text
  field2 Text Maybe
  field3 Text
  Primary field1 field3
</pre>

Error:

<pre>
Model.hs:13:7:
    Exception when trying to run compile-time code:
      primary key column cannot be nullable: "field1"
    Code: persistFileWith lowerCaseSettings "config/models"
    In the splice: $(persistFileWith lowerCaseSettings "config/models")
</pre>

config/models

<pre>
Test
  field1 Text
  field3 Text
  field2 Text Maybe
  Primary field1 field3
</pre>

Compiled fine.

As can be seen, "field3" is certainly not nullable in case 1 and "field1" is not nullable in case 2. 

It seems there is an assumption in the implementation that the primary key columns must be defined consecutively starting from the top - which is a fine assumption to make. But I feel the user should have to at least be shown a more accurate error message - that he or she has not followed the assumption.

hdgarrood:
This might be the same as #780.

parsonsmatt:
closing in favor of #780

PR #590, opened 2016-07-11
Allow composite Primary keys for tables that contain nullable fields.

cdepillabout:
The `takeComposite` function in `persistent/Database/Persist/Quasi.hs` was incorrectly disallowing composite Primary keys for tables that contain nullable fields.

A definition like this:

```
PrimaryCompositeWithOtherNullableFields
   foo String
   baz String Maybe
   bar String
   Primary foo bar
```

Would generate an error:

```
/path/to/some/haskell/file/Foo.hs:15:23:
    Exception when trying to run compile-time code:
      primary key column cannot be nullable: "bar"
    Code: template-haskell-2.10.0.0:Language.Haskell.TH.Quote.quoteExp
            persistLowerCase
            "\n\
            \    Primary CompositeWithOtherNullableFields\n\
            \        foo String\n\
            \        baz String Maybe\n\
            \        bar String\n\
            \        Primary foo bar\n\
```

This PR fixes the logic in `takeComposite` so that it only requires the columns being used in the primary key be non-nullable.

A small test table is also added.

### DIFF
--- a/persistent-test/src/CompositeTest.hs
+++ b/persistent-test/src/CompositeTest.hs
@@ -85,6 +85,13 @@ share [mkPersist persistSettings { mpsGeneric = False }, mkMigrate "compositeMig
     address AddressId
     Primary citizen address
     deriving Eq Show
+
+  PrimaryCompositeWithOtherNullableFields
+    foo String
+    bar String
+    baz String Maybe
+    Primary foo bar
+    deriving Eq Show
 |]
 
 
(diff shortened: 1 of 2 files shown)


cdepillabout:
Fixes #536.

psibi:
The tests seem to be failing for MySQL backend.

gregwebs:
great! Lets see if #591 fixes MySQL

gregwebs:
new PR works. Thanks for the fix!

PR #591, opened 2016-07-13
Composite primary key with nullable fields

gregwebs:
Add a MySQL fix on top of #590


cdepillabout:
@gregwebs Thanks for getting this working on MySQL!  I was planning on doing it yesterday, but I ran out of time!  Looks like you beat me to it :-)

PR #597, opened 2016-07-30
Add MonadBaseControl as part of type signature

psibi:
This is because of change in the package mongoDB where type signature
has changed from `MonadIO m => Selection -> Action m ()` to `(MonadIO m,
MonadBaseControl IO m) => Selection -> Action m ()`. This leads to the
additon of extra type constraint to make it compile.

Also helps in fixing the nix error:
https://hydra.nixos.org/build/37252928/nixlog/1/raw

Also Update mongoDB version constraint to 2.1.0
Bump aeson-compat version constraint.
Remove aeson-extra from persistent-test package.

### DIFF
--- a/persistent-mongoDB/persistent-mongoDB.cabal
+++ b/persistent-mongoDB/persistent-mongoDB.cabal
@@ -27,7 +27,7 @@ library
                    , bytestring         >= 0.9
                    , conduit            >= 0.5.3
                    , resourcet          >= 0.3
-                   , mongoDB            >= 2.0.3   && < 3.0
+                   , mongoDB            >= 2.1.0   && < 3.0
                    , bson               >= 0.3.1   && < 0.4
                    , network            >= 2.2.1.7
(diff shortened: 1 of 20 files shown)


gregwebs:
Thanks! I will try to get this merged tomorrow. Only thing from looking at this is I am wondering why some APIs require `MonadBaseControl` and others do not. It seems like we should add this constraint at a base-level if possible so that it applies everywhere (either in persistent or just in persistent-mongoDB)

psibi:
Yeah, this change is pretty invasive just because of an external package. I personally don't like this patch. :) But I don't know if it is possible to just make the change in `persistent-mongoDB` to achieve this.

gregwebs:
@snoyberg is there an alternative to forcing `MonadBaseControl` on all of persistent?

snoyberg:
I've opened #599 with an alternative.

psibi:
Closing this in favor of #599

PR #599, opened 2016-08-02
Support for mongoDB 2.1

snoyberg:
Pinging @gregwebs and @psibi, relevant to #597. The fact that I could work around this in this manner means to me that the mongoDB library made a mistake in their API change.

### DIFF
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -173,6 +173,12 @@ lookupEnv key = do
     return $ lookup key env
 #endif
 
+-- | Helper function to deal with MonadBaseControl constraints
+runMongo :: Trans.MonadIO m => DB.Action IO a -> DB.Action m a
+runMongo action = do
+    env <- ask
+    liftIO (runReaderT action env)
+
 instance HasPersistBackend DB.MongoContext where
     type BaseBackend DB.MongoContext = DB.MongoContext
     persistBackend = id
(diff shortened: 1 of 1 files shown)


psibi:
This looks way better. Have you intentionally not added the `MonadBaseControl` constraint to the `runMongo` function ? Will the tests work if mongoDB version constraint is bumped to 2.1.0. ? The Travis test for this repository was passing successfully previously. The nix hydra servers were failing for this package because it was picking the new version of `mongoDB`.

snoyberg:
I haven't experimented with the tests, but I did locally test against the new version of mongoDB. We can always bump the version number in the stack.yaml file, I have no opinion on that.

The trick going on here is that, instead of having a `MonadBaseControl` constraint, I've just constrained the monad in the argument (i.e., in negative position) to `IO`. This fulfills the `MonadBaseControl` constraint, and then I can just `liftIO` the resulting action back into the external monad.

I don't have time for it, but ideally I would have sent a PR upstream to mongoDB to make the change there. Ideally, all functions would have a signature like one of the following:

```haskell
foo :: MonadIO m => arg -> Action m res
foo :: (MonadIO m, MonadReader MongoContext m) => arg -> m res
```

The reason this works is that, AFAICT, there is never a reason to have the base monad in negative position (such as with a callback function or an exception handler).

gregwebs:
@VictorDenisov lets look at fixing this upstream

VictorDenisov:
Ok. Looking into it.

VictorDenisov:
I think I understood the concept. I can do this trick in mongodb. I'll release another version of mongodb by the end of this week.

VictorDenisov:
Published https://hackage.haskell.org/package/mongoDB-2.1.1. It shouldn't have MonadBaseControl for functions modified in 2.1.0.1. I also see that some of functions still have MonadBaseControl. We can remove those restrictions in later versions.

PR #604, opened 2016-08-14
Travis Postgres server update to 9.4

psibi:
Once the CI officially supports 9.5, we can switch to that. Switching to
9.5 will verify us that the atomic upsert feature of postgres works as
expected.


gregwebs:
great! For now we get a good test that older versions still work :)

PR #605, opened 2016-08-14
monad-control CPP pre-processor cleanup

psibi:


psibi:
I will merge this if there is no objections.

gregwebs:
thanks!

Issue #558, opened 2016-05-03
Non-existent withStmt referenced in documentation


nd2s:
I was trying to find a way of selecting a varying number of columns from an SQL query only known at runtime. I found documentation about `withStmt` in [Database.Persist.Sql:ConnectionPool](http://haddock.stackage.org/lts-5.15/persistent-2.2.4.1/Database-Persist-Sql.html#t:ConnectionPool) but this seems to have gone in v2?

It probably wouldn't have helped me anyway because this apparently only worked with tuples not lists, but looks like a bug in documentation to me.

snoyberg:
@gregwebs Do you know anything about this?

gregwebs:
This is not familiar to me. Is there is any use case for `withStmt` instead of using `rawSql`?

Tritlo:
Maybe to create extensions? Like `CREATE EXTENSION hstore;` in `Postgresql`.

psibi:
@chreekat removed the dangling do in https://github.com/yesodweb/persistent/pull/610. Closing the issue. Please re-open if necessary.

PR #610, opened 2016-09-19
Remove dangling doc reference to R.withStmt

chreekat:
Whatever module synonym "R" pointed to is gone, and "withStmt" may be lost to time as well.

### DIFF
--- a/persistent/Database/Persist/Sql/Types.hs
+++ b/persistent/Database/Persist/Sql/Types.hs
@@ -80,16 +80,8 @@ type ConnectionPool = Pool SqlBackend
 -- some complex @JOIN@ query, or a database-specific command
 -- needs to be issued.
 --
--- To issue raw SQL queries you could use 'R.withStmt', which
--- allows you to do anything you need.  However, its API is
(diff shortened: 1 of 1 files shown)


psibi:
Related to this: https://github.com/yesodweb/persistent/issues/558

I would still like to know in which module the function was present. Maybe, it was hidden by mistake ?

chreekat:
@psibi I pick-axed through git history using `git log -S withStmt`. It looks like it used to be a record in a config parameter. And `R` was ` Database.Persist.GenericSql.Raw`, which is long gone.

Examples:
1. [withStmt field is removed](https://github.com/yesodweb/persistent/commit/3820295a7b92f3f591e17944d1f6b04abdc93f26#diff-4bd524cab55f45a450b3ab01dc7890f2L124)
2. [`R` goes away](https://github.com/yesodweb/persistent/commit/615418f3776cd18df29e5eec69ef915d11ab5235#diff-226b1af25f094fc806a79f127899f234L55)

psibi:
Thanks for digging up the history!

PR #614, opened 2016-10-04
Reimplement upsert and tests

### LINKED ISSUE #613: Persistent does not implement upsert
Upsert is an either/or. Either the row exists and is updated, or it does not exist and is inserted.

Persistent, however, implements a function that inserts _and then updates_ a row that does not exist. The tests expect that behavior, even though it is incorrect. What's more, Persistent's version of upsert _replaces_ a row that exists, rather than just updating it![*] Nowhere have I read that to be the desired behavior.

[*] An existing row is only replaced if upsert's second argument is `[]`, but I still think that's a bug.

In short:

|  | **Expected** | **Actual** |
| --- | --- | --- |
| **Row Exists** | updated | replaced if updates == `[]` |
| **Row Doesn't Exist** | inserted | inserted and then updated |

Correct me if I'm wrong. I'll be sending along a patch soon in the meanwhile.

---

> Upsert: An operation that inserts rows into a database table if they do not already exist, or updates them if they do.
> — https://en.wiktionary.org/wiki/upsert

\

> A relational database management system uses SQL MERGE (also called upsert) statements to INSERT new records or UPDATE existing records depending on whether [the given] condition matches.
> — https://en.wikipedia.org/wiki/Merge_(SQL)

\

> "UPSERT" is a DBMS feature that allows a DML statement's author to atomically either insert a row, or on the basis of the row already existing, UPDATE that existing row instead....
> — https://wiki.postgresql.org/wiki/UPSERT

\

Illustrative (questionable) test:

> ``` haskell
> it "upsert with updates" $ db $ do
>     deleteWhere ([] :: [Filter Upsert])
>     let email = "dude@example.com"
>     Nothing :: Maybe (Entity Upsert) <- getBy $ UniqueUpsert email
>     let up0 = Upsert email 0
>     Entity _ up1 <- upsert up0 [UpsertCounter +=. 1]
>     upsertCounter up1 @== 1
> ```
> 
> — https://github.com/yesodweb/persistent/blob/master/persistent-test/src/PersistentTest.hs#L509

`upsertCounter up1` should be 0.

chreekat:
Fixes #613, making `upsert` do what is expected.

Tests were rewritten for clarity as well as behavior.

### DIFF
--- a/persistent-test/src/Init.hs
+++ b/persistent-test/src/Init.hs
@@ -135,7 +135,7 @@ infix 1 @/= --, /=@
 actual @/= expected = liftIO $ assertNotEqual "" expected actual
 
 infix 1 @==, ==@
-expected @== actual = liftIO $ expected @?= actual
+actual @== expected = liftIO $ actual @?= expected
 expected ==@ actual = liftIO $ expected @=? actual
 
 {-
--- a/persistent-test/src/PersistentTest.hs
+++ b/persistent-test/src/PersistentTest.hs
@@ -59,6 +59,7 @@ import Data.Aeson
 
 import Data.Conduit
 import qualified Data.Conduit.List as CL
+import Data.Function (on)
 import Data.Functor.Identity
 import Data.Functor.Constant
 import PersistTestPetType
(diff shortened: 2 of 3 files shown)


chreekat:
Help running the tests on other CPP combos would be appreciated. Not sure what the error is, since the message is truncated.

psibi:
@chreekat I'm closing this PR in favor of this https://github.com/yesodweb/persistent/pull/618 as that already contains your commits. Re-open if you disagree.

PR #618, opened 2016-10-06
Fix upsert column error (For postgres 9.5 + server)

psibi:


persistent/Database/Persist/Sql/Orphan/PersistUnique.hs
```diff
@@ -18,43 +21,75 @@ import qualified Data.Conduit.List as CL
...
+    uniqueKey <- onlyUnique record
+    upsertBy uniqueKey record updates
 
-instance PersistUniqueWrite SqlBackend where
+escape :: DBName -> T.Text
+escape (DBName s) = T.pack $ '"' : go (T.unpack s) ++ "\""
+  where
+    go "" = ""
+    go ('"':xs) = "\"\"" ++ go xs
+    go (x:xs) = x : go xs
```
chreekat: I think hlint would say to "avoid explicit recursion" here. I'm bad about "pointless" code golf, but since I'm never getting that time back anyway:

```haskell

escape :: DBName -> T.Text
escape = quote . T.foldr esc "" . unDBName
  where
    quote = T.cons '"' . flip T.snoc '"'
    esc '"' = mappend "\"\""
    esc c   = T.cons c
```

Feel free to ignore. :cat2:

persistent/Database/Persist/Sql/Orphan/PersistUnique.hs
```diff
@@ -18,43 +21,75 @@ import qualified Data.Conduit.List as CL
...
+                                    , n
+                                    , "-?"]
+                            go'' n Multiply =
+                                T.concat
+                                    [ n
+                                    , "="
+                                    , escape (entityDB t) <> "."
+                                    , n
+                                    , "*?"]
+                            go'' n Divide = T.concat [n, "=", n, "/?"]
```
chreekat: Does `Divide` not need the same treatment?

chreekat: Reducing duplication might make it easier to spot differences.

```haskell
go'' n Add      = goOp n "+"
go'' n Subtract = goOp n "-"
...
goOp n op = T.concat [n, "=", escape (entityDB t), ".", n, op, "?"]
```

psibi: Good catch, I will fix this and probably make it generic as you have suggested. The previous code, I think was adapted from somewhere else in the same codebase.

psibi:
Ah sorry, I didn't notice that it included a fix for that bug also. I will merge your changes with this branch and try to fix the failing tests there. Thanks!

persistent/Database/Persist/Sql/Orphan/PersistUnique.hs
gregwebs: mostly I am concerned about readability. I like to avoid naming things `go` if possible.

psibi: Thanks for the feedback, I will change it to `escapeQuote`. Actually, I didn't write the `escape` function itself. I took it from some other parts in `persistent`.

chreekat:
@psibi woot! Thanks for clearing all this up.

psibi:
Right now the tests fail for MongoDB. @chreekat I changed your implementation slightly to make it compile with other versions of ghc.

psibi:
Any idea/pointers on how to fix the Mongodb tests ? I did try changing the implementation of it's upsert like this (locally - not in the PR) for mongodb:

```
    upsertBy uniq newRecord upds = do
        let uniqueDoc = toUniquesDoc uniq
        let selection = DB.select uniqueDoc $ collectionName newRecord
        mdoc <- getBy uniqhttps://news.ycombinator.com/
        case mdoc of
          Nothing -> insert_ newRecord 
          Just _ -> if null upds
                    then return ()
                    else DB.modify selection $ updatesToDoc upds
        newMdoc <- getBy uniq
        case newMdoc of
          Nothing -> err "possible race condition: getBy found Nothing"
          Just doc -> return doc
      where
        err = Trans.liftIO . throwIO . UpsertError
```
While it did produce correct output for the sample programs, I was trying locally - The tests are still failing. CC: @gregwebs and others

chreekat:
@psibi There's some copypasta in what you posted. Assuming that's not the problem, I don't know enough about Mongo to have any input, myself.

psibi:
@chreekat Ah sorry, fixed that. It was a copy paste error. I looked into this issue today and I think the reasons the test are failing are because the uniqueness constraints are not enforced for `mongoDB`. I have opened a relevant issue here: https://github.com/yesodweb/persistent/issues/622

psibi:
I have fixed MongoDB tests by introducing additional CPP for it specifically. The reasons why the tests were failing was because the `insertEntity` function was colliding with the previous rows which was inserted as part of the test. To avoid that, I introduced specific CPP for them. @chreekat Can you review the upsert tests for them ?

psibi:
@gregwebs and others Can this be reviewed and possibly merged ? This fixes the upsert behavior. I have also changed mongodb upsert related code which I'm not super familiar with but have sort of got it to work.

gregwebs:
I think it would be better to add the unique index for mongoDB instead of adding in a bunch of CPP

psibi:
@gregwebs Add unique indexes for it in the test using the function supplied by the package `mongoDB` ?

I think I tried doing it, but it was failing for some reason I cannot remember. I will probably try it again and let you know.

chreekat:
I might be missing something, but it looks like the tests are doing the same thing in both branches of the #ifdef!...

psibi:
@gregwebs I added unique index like this, but it doesn't help:
  
```             
#ifdef WITH_MONGODB
        let index = Index { iColl = collectionName record, iKey = ["email" MongoDB.=: (1 :: Int32)], iName = "uniqueEmail", iUnique=True, iDropDups = True, iExpireAfterSeconds = Nothing}
        MongoDB.liftDB $ MongoDB.createIndex index
#endif
```

The reason is that in the test code there is a usage of the `insertEntity` function. Between two `it` blocks in the test code, the `insertEntity` has same unique field which leads to the test failure. For other backends, apparently it doesn't matter. One way of fixing this would be to drop the mongo collection after each `it` block is executed. Another way is to not use the same keys between `it` blocks which I have done presently.

psibi:
@gregwebs @snoyberg Can this be reviewed and merged ?

gregwebs:
Awesome, thanks!

gregwebs:
It seems like there are no breaking changes here, but I wanted to double check with you if it needs any version bumps? What is your hackage handle? I would like to add you to the hackage maintainer list.

psibi:
@gregwebs 

For those who relied on the previous buggy behavior of `upsert`: This will be a breaking change.
So, I think we can do a major release here. I can add a relevant haddock documentation for upsert in another PR and release a new version.

My Hackage handle is  `psibi`

gregwebs:
Thanks! You should be able to upload to hackage now.

psibi:
@gregwebs 

I have uploaded new version of `persistent`. Note that a new version of `persitent-mongoDB` also needs to be released as it involved some changes in it's upsert code.

ygale:
@psibi @chreekat @gregwebs This breaking change just caused a major regression in our deployed application in production. That's fine if this was really needed - but could you please update the description with the details of how this changes the behavior, beyond what is described in #613? Right now all we have here is a huge pile of code patches that are really hard to follow. Thanks!

psibi:
@ygale Sorry for the regression you are facing. That's why persistent was released with a major version bump in `2.7.0`.  Nothing is changed beyond what is explained in #613, so only the upsert function's behaviour is changed. You can see the comprehensive test suite for `upsert` to get a more clear idea. Let me know if you need more clarifications or if any parts of #613 is unclear.

ygale:
@psibi No problem we're really glad for your work on fixing this problem, thanks! If the description in #613 covers it in terms of behavior, then we're good. Thanks!

chreekat:
@ygale just for reference, #668 also provides some context for the change in #618.

To summarize what is new with #618 / 2.7.0 :

1. `upsert` does not replace existing rows, it only updates them. (It used to do both.)
2. `upsert` does not replace an existing row when the list of updates is empty -- it does nothing. (It used to act like `replace`).

ygale:
@chreekat Great thanks! Sounds like if you could just write this in the (currently empty) description field of this PR:

> Implements #613 

that will save a lot of time and worry for anyone else who is affected by the change. Thanks again to everyone for fixing this.

chreekat:
Cool, I took the liberty of putting words in @psibi's ~~mouth~~ textbox. :)

PR #626, opened 2016-10-24
Relax upper bound on mysql-simple

paul-rouse:
This is best done before we try to get persistent-mysql back into stackage.

The changed types/instances in mysql-simple don't seem to used at all by persistent-mysql, so I believe we only need a minor version bump here (or even a hackage revision) - there are extra tuple instances for QueryResults and an extra field in ResultError.


snoyberg:
Thanks! I've edited the cabal file on Hackage.

PR #627, opened 2016-10-25
Relax upper bound on mysql-simple - again

paul-rouse:
I'll release mysql-simple-0.4.0.0 soon, but I'll let you do this first to avoid unnecessary bounds issues in the next stackage build.

The theoretically-breaking change is another addition to ResultError, and as before it shouldn't affect persistent-mysql versioning.  I had missed it while working through the old PRs :-(


snoyberg:
OK, merged and updated on Hackage.

PR #629, opened 2016-11-01
NUMERIC in MySQL is implemented as DECIMAL

qrilka:
See e.g. http://dev.mysql.com/doc/refman/5.7/en/fixed-point-types.html
Having `NUMERIC` here leads to unnecessary no-op migration from `DECIMAL` to `NUMERIC`.


psibi:
What's the advantage of using `DECIMAL` as compared to `NUMERIC` ? I cannot find anything in the pointed link that tells about it.

> Having NUMERIC here leads to unnecessary no-op migration from DECIMAL to NUMERIC.

Where does it lead to unnecessary migration ?

qrilka:
That link describes that in MySQL `NUMERIC` is implemented as `DECIMAL` thus when persistent creates a new `NUMERIC` that gives new `DECIMAL` field in MySQL and when persistent migrations are used they do `ALTER TABLE` trying to convert `DECIMAL` to `NUMERIC` which actually doesn't change  table.
So I'd rather ask you what is the advantage of using `NUMERIC` in persistent when you'll have a `DECIMAL` in MySQL and thus a minor inconsistency between persistent and MySQL?

psibi:
>  when persistent migrations are used they do ALTER TABLE trying to convert DECIMAL to NUMERIC which actually doesn't change table.

Yeah, that should not happen. I guess that's because of this: https://github.com/yesodweb/persistent/blob/8c9bb04cb617ee7f1c5f2a0275b283b04aa989b6/persistent-mysql/Database/Persist/MySQL.hs#L579  You are right that it should be a no-op migration.

> So I'd rather ask you what is the advantage of using NUMERIC in persistent when you'll have a DECIMAL in MySQL and thus a minor inconsistency between persistent and MySQL?

I don't know! I'm just trying to see what's the best and adapt it to that in either the `show` or the `parsing` stage. I will wait for others to review this. Thanks for the patch.

gregwebs:
I think we should we be fixing `parseType` to parse both types (regardless of which one persistent prefers to use). This will help persistent work with schemas it didn't create.

qrilka:
@gregwebs that concern seems to be a bit larger than this PR - as in 3f700630e803790ff64e1f5c36e18b60e0f34bf7 `parseType` was made almost equal to `SqlOther . T.decodeUtf8`.
I wish there were some more information about the motivation behind that commit

gregwebs:
I think the concern is relevant because someone may be expecting the current behavior for inter-op with external migration tooling.

I believe MySQL always had some issues with migrations and that the change mentioned simplified things. I think that change makes sense: why bother adding intelligence if it isn't needed? In this case it seems like some intelligence is needed.

qrilka:
BTW in wiki it's documented that `DECIMAL` is used in MySQL - https://github.com/yesodweb/persistent/wiki/Persistent-entity-syntax :)

Issue #625, opened 2016-10-20
persistent-mongoDB: inconsistent data on write actions leading to inaccurate reads


dfithian:
In persistent-mongoDB `insertKey` is implemented using `toInsertDoc`, which omits "null" values from the write action (empty collections and Nothing, for example): https://github.com/yesodweb/persistent/blob/master/persistent-mongoDB/Database/Persist/MongoDB.hs#L452

`insert` also leverages `toInsertDoc`.

However, `repsert` serializes using `recordToDocument`, which does not filter "null" values: https://github.com/yesodweb/persistent/blob/master/persistent-mongoDB/Database/Persist/MongoDB.hs#L483

Inconsistency aside, the behavior of the `insert` functions affects select actions in that selecting for an empty list or Nothing will fail to find such a document.

I propose rewriting the `insert`-like functions to use the same serialization logic as `repsert`.

gregwebs:
I thought it was bad practice to write queries that are only satisfied by an empty list or a null. Persistent queries will check for not-existence.

I wrote this before MongoDB's newer storage engine when it was well known that their storage was very inefficient, so it seemed like a win to not write anything to the document. If you could point out that the cost of writing nulls is very low than I don't think I would have an objection to making the change.

gregwebs:
Definitely any inconsistencies should be addressed.

dfithian:
The *computed* value for a query might be an empty list, but in the current state of things an empty list would be omitted from the document when writing. As a result, the empty list is not equal to a missing value, and the query fails to satisfy that case. I'll look into storage.

TomMD:
Is this related to the fact that `insert (Just []) >> lookup ~~> Nothing` or is that a separate bug?

PR #630, opened 2016-11-06
yesodweb/persistent#625 Use same serialization logic for insertion

dfithian:
yesodweb/persistent#625 Use same serialization logic for insertion


gregwebs:
Take a look at the Travis builds
* some changes to get rid of warnings that broke backwards compatibility:https://travis-ci.org/yesodweb/persistent/jobs/173751674
* There is a failing test case for MongoDB: https://travis-ci.org/yesodweb/persistent/jobs/173751685

PR #632, opened 2016-11-10
Implement proper column type parsing for MySQL

qrilka:
We need to parse correctly column type information from MySQL
so persistent types will be properly mapped to the ones
persistent-mysql creates in MySQL

Closes #629


gregwebs:
awesome, thank you!

qrilka:
@gregwebs is there any chance to see this on hackage?
Ideally I'd like to see #641 included into that new release of `persistent-mysql` too

PR #636, opened 2016-11-18
Doc comments to the Migration module

parsonsmatt:
I studied the source code of the Migrations module and wrote up my findings. If these are correct and welcome, then please accept this token of documentation. Otherwise, I'm happy to edit/update/clarify. Thanks!

### DIFF
--- a/persistent/Database/Persist/Sql/Migration.hs
+++ b/persistent/Database/Persist/Sql/Migration.hs
@@ -33,6 +33,8 @@ allSql = map snd
 safeSql :: CautiousMigration -> [Sql]
 safeSql = allSql . filter (not . fst)
 
+-- | Given a 'Migration', this parses it and returns either a list of
+-- errors associated with the migration or a list of migrations to do.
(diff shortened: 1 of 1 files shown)


persistent/Database/Persist/Sql/Migration.hs
```diff
@@ -42,26 +44,34 @@ parseMigration =
...
       Left errs -> error $ unlines $ map unpack errs
       Right sql -> return sql
 
+-- | Prints a migration.
 printMigration :: MonadIO m => Migration -> ReaderT SqlBackend m ()
 printMigration m = showMigration m
                >>= mapM_ (liftIO . Data.Text.IO.putStrLn)
 
+-- | Convert a 'Migration' to a list of 'Text' values corresponding to their
+-- 'SQL' statements.
```
MaxGabriel: Minor, but should this be `@SQL@` since it's not a Haskell type?

parsonsmatt: Ah I meant to refer to the `Sql` Haskell type.

snoyberg:
Thanks!

Issue #634, opened 2016-11-16
Bug in wiki?


qrilka:
https://github.com/yesodweb/persistent/wiki/Persistent-entity-syntax#nullable-fields shows `default=Nothing` as a way to set `NULL` by default but in the section below it's said that:
```
The DEFAULT attribute is set with the exact SQL entered in your models file—Persistent
performs no analysis of default values, and simply passes them onto the DBMS verbatim.
```
and http://hackage.haskell.org/package/persistent-2.6/docs/src/Database.Persist.Sql.Internal.html#defaultAttribute shows that it's correct

P.S. that em dash without surrounding spaces looks quite odd in that sentence

MaxGabriel:
That does sound like a bug to me. Would you be up for running a quick test to see if default=Nothing works, and updating the documentation accordingly?

I think the use of spaces around em dashes is a regional thing, or just a font thing. Wikipedia says [different style guides](https://en.wikipedia.org/wiki/Dash#Spacing_and_substitution) have different recommendations:

> According to most American sources (such as The Chicago Manual of Style) and some British sources (such as The Oxford Guide to Style), an em dash should always be set closed, meaning it should not be surrounded by spaces. But the practice in some parts of the English-speaking world, including the style recommended by The New York Times Manual of Style and Usage for printed newspapers and the AP Stylebook, sets it open, separating it from its surrounding words by using spaces or hair spaces (U+200A) when it is being used parenthetically.[

Here are some quotes from Butterick's Practical Typography on it:

> Em and en dashes are typ­i­cally set flush against the sur­round­ing text. Some fonts in­clude a lit­tle white space around the em dash; some don’t. If your em dashes look like they’re be­ing crushed—par­tic­u­larly if you’re set­ting type on screen—it’s fine to add word spaces be­fore and after.

> Fans of Robert Bringhurst’s book The El­e­ments of Ty­po­graphic Style (and I am among them—see bib­li­og­ra­phy) may know that he rec­om­mends us­ing en dashes with spaces rather than em dashes. Per­haps this prac­tice is com­mon in Mr. Bringhurst’s na­tive Canada. But in Amer­i­can ty­pog­ra­phy, it’s not.

qrilka:
@MaxGabriel the reason of this issue was a migration failure because of that `default=Nothing` which was fixed by `default=NULL` :)

MaxGabriel:
Ah, nice! I've updated the wiki with that fix!

PR #641, opened 2016-12-12
Proper mysql default migrations

qrilka:
Current `peristent-mysql` code treats `DEFAULT NULL` in DB as no default value so as a result we get noop migrations setting it to `DEFAULT NULL` for columns with `default=NULL` in entity declaration.
This PR fixes it

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -456,7 +456,7 @@ getColumns connectInfo getter def = do
     stmtIdClmn <- getter "SELECT COLUMN_NAME, \
                                  \IS_NULLABLE, \
                                  \DATA_TYPE, \
-                                 \COLUMN_DEFAULT \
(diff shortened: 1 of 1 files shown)


gregwebs:
Thanks for the contribution! Is my understanding correct that this a bug in MySQL that we have to work around where COLUMN_DEFAULT does not return NULL?

qrilka:
@gregwebs it appears to be the other way around - MySQL does return `NULL` but `persistent-mysql` sees it as `Nothing` and assumes it as "no default", see e.g. #634 
At the same time for non-nullable fields that `NULL` means exactly what `master` version of the code expects.
So it's about `NULL` in MySQL which is used for 2 things at the same time (depending on `IS_NULLABLE`):
* no default
* default of `NULL`

snoyberg:
Thanks!

qrilka:
thanks @snoyberg for merging this - are there any plans for new releases of `persistent` and `persistent-mysql`?

snoyberg:
I spent the last day updating the Travis config to be more comprehensive, and I'm going through the PR backlog right now. I'll likely be making releases towards the end of the week.

PR #645, opened 2016-12-19
Fix unexpected letter case

qrilka:


psibi:
@qrilka That's intended. You can see the comment above the code.

qrilka:
Ouch, sorry...

PR #646, opened 2016-12-28
Enable foreign key constraints by default for Sqlite backend.

rdnetto:
Sqlite has foreign key constraints [disabled by default](https://www.sqlite.org/foreignkeys.html#fk_enable). This PR enables foreign key constraints by default, and provides a mechanism for restoring the previous behaviour.

As this is a breaking change, I'm not sure if we'll want to bump the version number by more.

### DIFF
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,3 +1,7 @@
+## 2.6.1.0
+
+* Removed the behaviour of toggling write-ahead log (WAL) based on a prefix in the connection string, in favour of using a record.
+
 ## 2.6
 
 Compatibility for backend-specific upsert functionality.
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,6 +1,7 @@
 ## 2.6.1.0
 
 * Removed the behaviour of toggling write-ahead log (WAL) based on a prefix in the connection string, in favour of using a record.
+* Turned on foreign key constraints [#646](https://github.com/yesodweb/persistent/issues/646)
 
 ## 2.6
 
--- a/persistent-sqlite/Database/Persist/Sqlite.hs
+++ b/persistent-sqlite/Database/Persist/Sqlite.hs
@@ -4,20 +4,21 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE TemplateHaskell #-}
 -- | A sqlite backend for persistent.
---
--- Note: If you prepend @WAL=off @ to your connection string, it will disable
--- the write-ahead log. For more information, see
--- <https://github.com/yesodweb/persistent/issues/363>.
 module Database.Persist.Sqlite
     ( withSqlitePool
     , withSqliteConn
     , createSqlitePool
     , module Database.Persist.Sql
     , SqliteConf (..)
+    , SqliteConnectionInfo
+    , mkSqliteConnectionInfo
+    , sqlConnectionStr
+    , walEnabled
     , runSqlite
     , wrapConnection
+    , wrapConnection'
     , mockMigration
     ) where
 
@@ -49,6 +50,7 @@ import Control.Monad.Trans.Resource (ResourceT, runResourceT)
 import Control.Monad (when)
 import Control.Monad.Trans.Reader (ReaderT, runReaderT)
 import Control.Monad.Trans.Writer (runWriterT)
+import Lens.Micro.TH (makeLenses)
 
 -- | Create a pool of SQLite connections.
 --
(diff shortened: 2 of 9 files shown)


MaxGabriel:
I'm a huge fan of foreign key constraints and I think most Haskell developers would be too. 

#### Concerns:
1. This is a breaking change for developers whose application-level SQL statements don't respect foreign keys. Based on testing, it wouldn't break connecting to a database that's inconsistent, or updating rows that violate a foreign key constraint, as long as you don't update the violating column. Note: Persistent's auto-migration already sets up foreign keys for SQLite (they just aren't enforced by it), so users will have foreign keys set up unless they're manually configuring their database. However, people connecting to existing, inconsistent databases are likely to *not* have foreign keys set up, so it could be less of an issue for them. Ultimately I'm in favor of making this breaking change because I think it's much better for users, and it should be easy to turn off.

2. I think the current behavior of configuring usage of the write-ahead log (WAL) in the connection string is really bad, as evidenced by things like needing to append the space at the end of `FK=off ` for things to work. It's undiscoverable, not typed, not well documented. I would be strongly in favor of adding a configuration record with fields for enabling/disabling the WAL as well as foreign keys as the way to implement this PR, and this could be a good time because it bundles the breaking changes.

#### Possible alternative:

1. Change the SQLite Yesod scaffolding to enable foreign keys. This would give new users foreign key checks but not affect older users.

Downsides: 

* Non-Yesod users (notably scripts) wouldn't get FKs by default
* More boilerplate in the scaffolding

#### Related documentation:

https://github.com/yesodweb/yesod-cookbook/blob/master/cookbook/Activate-foreign-key-checking-in-Sqlite.md

Thoughts?

rdnetto:
(Sorry for my delay in replying - I've been away over the last week and haven't been checking my email.)

> I think the current behavior of configuring usage of the write-ahead log (WAL) in the connection string is really bad, as evidenced by things like needing to append the space at the end of FK=off for things to work. It's undiscoverable, not typed, not well documented. I would be strongly in favor of adding a configuration record with fields for enabling/disabling the WAL as well as foreign keys as the way to implement this PR, and this could be a good time because it bundles the breaking changes.

Agreed. 

> Change the SQLite Yesod scaffolding to enable foreign keys. This would give new users foreign key checks but not affect older users.

I'm not familiar with Yesod, but going by [this example](http://www.yesodweb.com/book/persistent#persistent_integration_with_yesod) it looks like the normal way to integrate persistent with it is to use `withSqlitePool` (or similar) to create the connection pool. Based on that, it seems like the right way to do it would be to replace all the instances of the connection string argument with `SqliteConf`, and add the WAL/FK flags to that. This would provide the same interface to Yesod and non-Yesod users alike. The only downside is that it also has a `sqlPoolSize` field, which would be ignored by `withSqliteConn` and `runSqlite`.

MaxGabriel:
Ok, so let's assume we'll use `SqliteConf` to hold the new options (and leave the pool size unused in certain methods). Then we make these changes (might be easier to just read the code):

1. Functions that currently take a `Text` for the database name take a `SqliteConf`.
2. Add a `Default` instance for the `SqliteConf`. This gives us the ability to add new SQLite configuration options as people need them without making breaking changes, since we can direct people to use the `Default` instance.
3. Add an `IsString` instance for `SqliteConf` where the string specifies the database name, with the defaults being used for the other values. This gives us an easy migration path, since people are probably already using `OverloadedStrings` to specify the database name, they should notice no changes.
4. If `WAL=on/off` is present in the connection string, a deprecation warning will be printed recommending that the option in `SqliteConf` be used instead. In the meantime the connection string value can override the `SqliteConf` value until it's removed.
5. Unclear what to do with the pool size parameter in functions like `withSqlitePool`, since that would be duplicated in `SqliteConf`. Possibly these functions could be deprecated, new functions could be added that just take the `SqliteConf`, and later the new functions replace the existing functions. Alternatively, we could keep the SQLite options in a separate data type (say, `SqliteConnectionOptions`). This is logically nice because the pool size is at a higher level than an individual connection, and it doesn't necessitate changing existing functions. It might be more complex though.

Here's what the new types and typeclasses would look like:

```haskell
-- | Options for connecting to a SQLite database. 
--
-- Since additional fields are very likely to be added to accommodate different use cases, you should instantiate this record using the 'Default' instance to avoid breaking changes.
--
-- Example usage of 'def' here
--
-- If you have the @OverloadedStrings@ extension enabled, you can also just pass a literal string to specify a database name, using the defaults for the remaining options:
--
-- Example usage of the overloaded strings option here
--
-- @since 2.7
data SqliteConf = SqliteConf
    { sqlDatabase :: Text -- ^ The connection string to use, typically the database name, or @:memory:@ for an in-memory database. Default: @:memory:@
    , sqlPoolSize :: Int -- ^ The number of connections to keep in the pool. This value is ignored by functions that do not maintain a pool of connections. Do not use the @:memory:@ database when pooling connections, since the pool will regularly remove connections, destroying your database. Default: 1
    , sqliteEnableWal :: Bool -- ^ Whether or not to enable the write ahead log (WAL). See <http://www.sqlite.org/wal.html>. Default: @True@
    , sqliteEnableForeignKeys :: Bool -- ^ Whether or not to enable foreign key checks. See <https://www.sqlite.org/foreignkeys.html>. Default: @True@
    } deriving (Show)

instance Default SqliteConf where
  def = SqliteConf 
        { sqliteConnectionString = ":memory:"
        , sqlPoolSize = 1
        , sqliteEnableWal = True
        , sqliteEnableForeignKeys = True
        }

instance IsString SqliteConf where
  fromString s = def { sqliteConnectionString = pack s }
```

<hr>

@rdnetto Tell me what you think? Maybe cc @snoyberg on these changes?

snoyberg:
I'd lean towards having new functions that work with this new datatype.
Also, instead of a Default instance, why not just rely on the IsString
instance and then provide setter functions to modify various fields? I'm
basically getting at this design:
http://www.snoyman.com/blog/2016/11/designing-apis-for-extensibility

On Wed, Jan 4, 2017 at 12:30 AM, Maximilian Tagher <notifications@github.com

MaxGabriel:
Oh, just exporting the setters is a good idea, but I'm not huge on relying on the `IsString` instance. I would prefer the `Default` instance or a function like `mkSqliteSettings` that accomplishes the same thing. `mkSqliteSettings` would be more explicit, documentable and could take required parameters that we don't want handled by default (the database name?), so it might be preferable. Reasoning:

1. The `Default` typeclass matches Yesod-related packages like [cookie](https://www.stackage.org/haddock/lts-7.14/cookie-0.4.2.1/Web-Cookie.html) and the `mkSqliteSettings` option matches packages like [auto-update](https://www.stackage.org/package/auto-update). I feel like these two patterns must be more widespread than using an `IsString` instance for the default values, then overriding them with setters.
2. `"test" { sqliteEnableWal = True }` looks weird to me
3. Just using a normal function or typeclass just seems a lot simpler. A beginner would know about functions and typeclasses, but maybe not overloaded strings (or their use of them is purely to use `Text`/`ByteString` literals). They could also search for documentation on `def` or `mkSqliteSettings`, but they couldn't for an overloaded string.

<hr>

> I'd lean towards having new functions that work with this new datatype.

That sounds good to me!

MaxGabriel:
Cool, I think that fleshes out the implementation details. @rdnetto is this something you want to implement? If not I can make these changes and incorporate your specific work on foreign keys

rdnetto:
Sure, no problem.

rdnetto:
I've rewritten the patch use the record-based approach. I've added a dependency on `microlens-th` to generate the lenses for export - I think that makes more sense than writing them by hand, but wanted to validate that assumption.

It's worth noting that merging this will exacerbate #647, but since that's already broken for all non-Sqlite relational databases, I don't think it's a blocker.

rdnetto:
I can't seem to make any further progress with this. I've got the code itself working, but `persistent-test` fails with compilation errors for the GHC 7.10.2 builds (but not the 7.8.4 ones), which I can't reproduce locally. Given that most of them are unable to find `getEntity`, which isn't something I touched, I'm wondering if I've messed up the `#ifdef`s somehow

@MaxGabriel, would you mind taking a look?

merijn:
@rdnetto The test output contains persistent-sqlite-2.6 rather than persistent-sqlite-2.6.1.0 which you changed the version number too. Isn't it simply running tests against an old version of persistent-sqlite for some odd reason?

rdnetto:
@merijn Good catch - putting in an explicit dependency on it did the trick. Looks like it was using the older version because of conflicting constraints on a library.

rdnetto:
Bump. :)

snoyberg:
For some reason, the patch is no longer available. @rdnetto do you know what's going on here?

rdnetto:
Oops, that's my fault. I forgot this PR was open and deleted the forked repo. I've since recreated it, but Github hasn't restored the association between it and the PR.

Since there's a merge conflict and I can no longer update it, the best option is to just create a new PR.

PR #651, opened 2017-02-23
Added functions to monitor and control SQLite memory usage

anton-dessiatov:
Also updated changelog and made a version bump in .cabal file.

No changes to existing API were made, just added 2 data structures and 2 functions. This stuff served a good job for me debugging application memory usage so it might be useful to others.

### DIFF
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,4 +1,4 @@
-## 2.6.0.2
+## 2.6.1
 
 * Added functions to monitor (status) and control (softHeapLimit) process-wide SQLite memory usage.
 
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,7 +1,6 @@
 ## 2.6.0.2
 
-* Added functions to monitor (status) and control (softHeapLimit) process-wide
-  SQLite memory usage.
+* Added functions to monitor (status) and control (softHeapLimit) process-wide SQLite memory usage.
 
 ## 2.6.0.1
 
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,3 +1,8 @@
+## 2.6.0.2
+
+* Added functions to monitor (status) and control (softHeapLimit) process-wide
+  SQLite memory usage.
+
 ## 2.6.0.1
 
 * Ensure connection is closed if wrapConnectionWal fails
--- a/persistent-sqlite/Database/Sqlite.hs
+++ b/persistent-sqlite/Database/Sqlite.hs
@@ -630,9 +630,10 @@ status verb reset = alloca $ \pCurrent -> alloca $ \pHighwater -> do
   let (code, hasCurrent, hasHighwater) = statusVerbInfo verb
   e <- decodeError <$> statusC code pCurrent pHighwater (if reset then 1 else 0)
   case e of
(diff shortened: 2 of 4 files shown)


persistent-sqlite/Database/Sqlite.hs
```diff
@@ -532,4 +536,101 @@ config c = case c of
...
+  -- | This parameter records the largest memory allocation request handed to
+  -- scratch memory allocator. Only the value returned in the 'sqliteStatusHighwater'
+  -- field of 'SqliteStatus' record is of interest. The value written into the
+  -- 'sqliteStatusCurrent' field is Nothing.
+  | SqliteStatusScratchSize
+  -- | This parameter records the number of separate memory allocations currently
+  -- checked out.
+  | SqliteStatusMallocCount
+
+statusVerbInfo :: SqliteStatusVerb -> (CInt, Bool, Bool)
```
snoyberg: Could we have a comment explaining what these values mean?

persistent-sqlite/ChangeLog.md
```diff
@@ -1,3 +1,7 @@
+## 2.6.0.2
```
snoyberg: With API extension, this should be a minor version bump (2.6.1).

persistent-sqlite/Database/Sqlite.hs
```diff
@@ -532,4 +536,101 @@ config c = case c of
       ErrorOK -> return ()
       _ -> sqlError Nothing "sqlite3_config" e
 
-
+-- | Return type of the 'status' function
+--
+-- Since 2.6.0.2
```
snoyberg: Can you switch this to `@since 2.6.1` (and elsewhere)?

persistent-sqlite/cbits/config.c
```diff
@@ -4,3 +4,7 @@
 int persistent_sqlite_set_log(void (*logFn)(void*, int, const char*), void* arg) {
   return sqlite3_config(SQLITE_CONFIG_LOG, logFn, arg);
 }
+
+long long int persistent_sqlite3_soft_heap_limit64(long long int N) {
```
snoyberg: What is this function adding versus just calling `sqlite3_soft_heap_limit64` directly?

persistent-sqlite/ChangeLog.md
snoyberg: Can you include a bump in the .cabal file?

persistent-sqlite/cbits/config.c
anton-dessiatov: It's just a layer of protection from future versions of SQlite doing something weird with sqlite3_int64 type. Currently that type is just a typedef for long long it so this function really does nothing. If, however, SQLite ever changes sqlite3_int64 to something different than long long int then this function will either do implicit conversion (if we are sort of lucky) or at least gives us a C compiler error.

snoyberg:
I think it's better if we drop it here. SQLite has shown a remarkable level
of API stability, I'm not worried about them suddenly changing that.

snoyberg:
Code looks good, just a few Travis failures that need to be addressed, e.g.:

```
Database/Sqlite.hs:634:74: Not in scope: ‘pure’
Database/Sqlite.hs:635:7:
    Not in scope: ‘<*>’
    Perhaps you meant ‘<$>’ (imported from Control.Applicative)
```

https://travis-ci.org/yesodweb/persistent/jobs/205780522

snoyberg:
Thanks!

PR #655, opened 2017-03-01
Enable foreign key constraints by default for Sqlite backend

rdnetto:
This is the PR from #646 re-raised.

CC: @snoyberg

### DIFF
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,3 +1,7 @@
+## 2.6.1.0
+
+* Removed the behaviour of toggling write-ahead log (WAL) based on a prefix in the connection string, in favour of using a record.
+
 ## 2.6
 
 Compatibility for backend-specific upsert functionality.
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,4 +1,4 @@
-## 2.6.2.0
+## 2.6.1.1
 
 * Removed the behaviour of toggling write-ahead log (WAL) based on a prefix in the connection string, in favour of using a record.
 * Turned on foreign key constraints [#646](https://github.com/yesodweb/persistent/issues/646)
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,6 +1,7 @@
 ## 2.6.1.0
 
 * Removed the behaviour of toggling write-ahead log (WAL) based on a prefix in the connection string, in favour of using a record.
+* Turned on foreign key constraints [#646](https://github.com/yesodweb/persistent/issues/646)
 
 ## 2.6
 
--- a/persistent-sqlite/Database/Persist/Sqlite.hs
+++ b/persistent-sqlite/Database/Persist/Sqlite.hs
@@ -6,10 +6,16 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell #-}
 -- | A sqlite backend for persistent.
+--
+-- Note: If you prepend @WAL=off @ to your connection string, it will disable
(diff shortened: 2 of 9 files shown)


snoyberg:
This constitutes significantly breaking changes to the API. Personally, I'm very averse to such changes unless they are absolutely warranted. I thought we'd decided on keeping the existing API and extending with new functions. Am I misremembering?

rdnetto:
I think some of the details might have gotten lost in the discussion.

Reading over it again, we agreed on the `mkSqliteSettings` approach (which I ended up calling `mkSqliteConnectionInfo`), but there was also a comment about having new functions to avoid breaking API which I missed.

I'll add a commit to fix this up so that it doesn't break API.

snoyberg:
Awesome, thanks!

rdnetto:
OK, that should do it. We're now exporting new functions, with the old ones marked as deprecated.

persistent-sqlite/ChangeLog.md
```diff
@@ -1,3 +1,8 @@
+## 2.6.1.1
```
snoyberg: This is going to need to be 2.6.2, but I can change that during merging.

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -57,7 +68,17 @@ import Control.Monad.Trans.Writer (runWriterT)
 -- Instead, use 'withSqliteConn'.
 createSqlitePool :: (MonadIO m, MonadLogger m, MonadBaseControl IO m, IsSqlBackend backend)
                  => Text -> Int -> m (Pool backend)
-createSqlitePool s = createSqlPool $ open' s
+createSqlitePool = createSqlitePoolFromInfo . conStringToInfo
+{-# DEPRECATED createSqlitePool "Use createSqlitePoolFromInfo" #-}
```
snoyberg: I'm not sure what the motivation is for deprecating these, I don't think we discussed that either. These string-based functions are likely sufficient for most use cases.

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -429,9 +482,9 @@ escape (DBName s) =
     go '"' = "\"\""
     go c = T.singleton c
 
--- | Information required to connect to a sqlite database
+-- | Information required to setup a connection pool.
 data SqliteConf = SqliteConf
-    { sqlDatabase :: Text
+    { sqlConnInfo :: SqliteConnectionInfo
```
snoyberg: This is a breaking change.

* [ ] Resolve breaking change

persistent-sqlite/Database/Persist/Sqlite.hs
rdnetto: - [ ] Remove deprecations

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -484,19 +481,29 @@ escape (DBName s) =
...
     , sqlPoolSize :: Int
     } deriving Show
 
 instance FromJSON SqliteConf where
-    parseJSON v = modifyFailure ("Persistent: error loading Sqlite conf: " ++) $
-      flip (withObject "SqliteConf") v $ \o -> SqliteConf
-        A.<$> o .: "database"
-        A.<*> o .: "poolsize"
+    parseJSON v = modifyFailure ("Persistent: error loading Sqlite conf: " ++) $ flip (withObject "SqliteConf") v parser where
+        parser o = if HashMap.member "database" o
```
snoyberg: This works fine, but for future reference the `Alternative` instance would probably be easier:

```haskell
(SqliteConf <$>o .: "database" ...) <|> (SqliteConfInfo <$> o.: "connInfo" ...)
```

snoyberg:
Thanks!

PR #665, opened 2017-04-13
Allow separate version check for PostgreSQL

dfithian:
I'm using persistent-postgresql with Amazon Redshift. There's a bug in Redshift that doesn't allow any user to call `show server_version`, so all of my connections fail. I was able to workaround this initially by removing all checks for the server version (my first commit) and now I would like to pass in my own function for providing the version.

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -136,7 +136,7 @@ createPostgresqlPool = createPostgresqlPoolModified (const $ return ())
 --
 -- <https://groups.google.com/d/msg/yesodweb/qUXrEN_swEo/O0pFwqwQIdcJ>
 --
--- Since 2.1.3
+-- @since 2.1.3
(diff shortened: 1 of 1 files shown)


gregwebs:
Its great to have Redshift support! The only problem with this change is that it breaks the API for `openSimpleConn` which is exported.

psibi:
Can you add the haddock markup with `since` for the function you are introducing ?

dfithian:
Done. I used 2.6.2 since the last released version was 2.6.1.

psibi:
I don't think the current haddock annotation for `since` that has been used is correct.

dfithian:
Then can you help me with a link? I have never heard of this feature.

psibi:
https://github.com/haskell/haddock/blob/master/doc/markup.rst#since

psibi:
Thanks!

dfithian:
Is there a next planned minor release? Would it be possible to release this change so I can get rid of my extra stack config?

PR #620, opened 2016-10-11
Add getEntity helper method.

cdepillabout:
This method is similar to `insertEntity`.

Closes #616.

PR #670, opened 2017-05-09
Relax constraint on `getEntity`

JaSpa:
This was probably a copy & paste error in #620. There's no need to require a `PersistStoreWrite` when only doing a get operation.


psibi:
Thanks!

PR #672, opened 2017-05-15
Fix #671 - revise handling of 'default null' in mysql

### LINKED ISSUE #671: persistent-mysql: #641 introduced a bug for text/blob fields
The change made in #641 created a problem: migration now attempts to alter the default value of `text` (and presumably `blob`) fields, but mysql does not allow these types to have defaults.  Rather distressingly, the migration is actually trying to *drop* the default (even though there isn't one) but it is still a fatal error!

It's easy to reproduce this using an out-of-the-box yesod-mysql scaffolding:
```
stack --resolver lts-8.13 new foo yesod-mysql
cd foo
<create a clean mysql database, and edit config/settings.yml accordingly>
stack build
stack exec foo    # First run sets up the database without problem
stack exec foo    # Second run attempts a spurious migration and fails with an error from mysql
```

The error message is:
```
Migrating: ALTER TABLE `user` ALTER COLUMN `password` DROP DEFAULT
foo: ConnectionError {errFunction = "query", errNumber = 1101, errMessage = "BLOB/TEXT column 'password' can't have a default value"}
```

Reverting the two lines changed in #641 fixes this completely.  The `password` field in the scaffolding is indeed nullable, so the new logic applies to it, but it is a text field, so the null in COLUMN_DEFAULT really does mean "unset", because it is not allowed to a default.

@qrilka I can see what you you wanted to achieve, but I am not sure how to fix it.  What do you think?

paul-rouse:
This PR undoes #641, which broke text/blob fields (#671), and fixes the handling of "default=NULL" by silently ignoring it.  Giving this default explicitly never makes any difference, since it is the implicit default for any nullable field.  The original problem addressed by #641 is that MySQL does not retain the distinction between the implicit and explicit versions, so migration could get confused, but #641 resolved the ambiguity in the direction of the explicit declaration, which causes an error for text/blob fields.  This PR goes in the other direction, and makes it implicit.

@snoyberg This is relatively serious, because the Yesod MySQL scaffolding is currently broken, but perhaps @qrilka would like a little while to confirm that the original problem is fixed.

### DIFF
--- a/persistent-mysql/ChangeLog.md
+++ b/persistent-mysql/ChangeLog.md
@@ -1,3 +1,7 @@
+##2.6.0.2
+
+Prevent spurious no-op migrations when `default=NULL` is specified - revised version [#672](https://github.com/yesodweb/persistent/pull/672) (which fixes bug [#671](https://github.com/yesodweb/persistent/issues/671) introduced by the earlier attempt [#641](https://github.com/yesodweb/persistent/pull/641))
+
 ## 2.6
 
 Compatibility for backend-specific upsert functionality.
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -457,7 +457,7 @@ getColumns connectInfo getter def = do
     stmtIdClmn <- getter "SELECT COLUMN_NAME, \
                                  \IS_NULLABLE, \
                                  \DATA_TYPE, \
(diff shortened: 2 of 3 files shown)


snoyberg:
Awesome, thank you! Could you add a ChangeLog note explaining the bug that this fixes? Once that's in, I'll merge and release.

paul-rouse:
Thanks for the reminder, here it is!  I had meant to put it in once the discussion was finished...

snoyberg:
Thanks! New version is now on Hackage.

PR #674, opened 2017-05-31
MySQL on duplicate key update

parsonsmatt:
MySQL has a native upsert in `INSERT ... ON DUPLICATE KEY UPDATE ...`. This PR introduces that functionality to `persistent-mysql`.

It also introduces a `bulkInsertOnDuplicateKeyUpdate` function, which has an API I'd like feedback on.

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -14,6 +15,9 @@ module Database.Persist.MySQL
     , MySQLBase.defaultSSLInfo
     , MySQLConf(..)
     , mockMigration
+    , insertOnDuplicateKeyUpdate
+    , bulkInsertOnDuplicateKeyUpdate
+    , SomeField(..)
     ) where
 
 import Control.Arrow
@@ -720,6 +724,7 @@ showColumn (Column n nu t def _defConstraintName maxLen ref) = concat
         Just s -> -- Avoid DEFAULT NULL, since it is always unnecessary, and is an error for text/blob fields
                   if T.toUpper s == "NULL" then ""
                   else " DEFAULT " ++ T.unpack s
+                  {-# LANGUAGE GADTs #-}
     , case ref of
         Nothing -> ""
         Just (s, _) -> " REFERENCES " ++ escapeDBName s
@@ -1026,7 +1031,7 @@ insertOnDuplicateKeyUpdate
        )
     => record
     -> [Update record]
-    -> ReaderT backend m ()
+    -> SqlPersistT m ()
 insertOnDuplicateKeyUpdate record =
   bulkInsertOnDuplicateKeyUpdate [record] []
 
@@ -1047,8 +1052,7 @@ data SomeField record where
 -- the value that is provided. You can use this to increment a counter value.
 -- These updates only occur if the original record is present in the database.
 bulkInsertOnDuplicateKeyUpdate
-    :: forall record m backend.
-       ( PersistEntityBackend record ~ BaseBackend backend
+    :: ( PersistEntityBackend record ~ BaseBackend backend
        , backend ~ SqlBackend
        , PersistEntity record
        , MonadIO m
@@ -1057,7 +1061,7 @@ bulkInsertOnDuplicateKeyUpdate
     => [record] -- ^ A list of the records you want to insert, or update
     -> [SomeField record] -- ^ A list of the fields you want to copy over.
     -> [Update record] -- ^ A list of the updates to apply that aren't dependent on the record being inserted.
-    -> ReaderT backend m ()
+    -> SqlPersistT m ()
 bulkInsertOnDuplicateKeyUpdate [] _ _ = pure ()
 bulkInsertOnDuplicateKeyUpdate records [] [] = insertMany_ records
 bulkInsertOnDuplicateKeyUpdate records fieldValues updates =
@@ -1067,8 +1071,7 @@ bulkInsertOnDuplicateKeyUpdate records fieldValues updates =
 -- garbage results if you don't provide a list of either fields to copy or
 -- fields to update.
 mkBulkInsertQuery
-    :: forall record m backend.
-    ( PersistEntityBackend record ~ BaseBackend backend
+    :: ( PersistEntityBackend record ~ BaseBackend backend
     , backend ~ SqlBackend
     , PersistEntity record
     , PersistStore backend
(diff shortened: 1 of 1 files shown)


persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1014,3 +1019,107 @@ mockMigration mig = do
...
+-- These updates only occur if the original record is present in the database.
+bulkInsertOnDuplicateKeyUpdate
+  :: ( PersistEntityBackend record ~ SqlBackend
+     , PersistEntity record
+     , MonadIO m
+     )
+  => [record] -- ^ A list of the records you want to insert, or update
+  -> [SomeField record] -- ^ A list of the fields you want to copy over.
+  -> [Update record] -- ^ A list of the updates to apply that aren't dependent on the record being inserted.
+  -> SqlPersistT m ()
```
parsonsmatt: This function would be used like:

```haskell
bulkInsertOnDuplicateKeyUpdate
  [ {- a big ol' list of records you want to insert/update -} ]
  [ SomeField UserName, SomeField UserEmail ] -- this copies the values that are being inserted to existing records
  [ UserModified =. now, UserEncounted +=. 1 ] -- this update is performed for any field that matches the inserted records
```

gregwebs:
Travis CI is pointint to some import issues on older versions

parsonsmatt:
@gregwebs Alright, CI is building now :smile: Any comments on the API or implementation?

psibi:
Minor comment: I would prefer the name inserManyOnDuplicateKeyUpdate instead of the current one as that is roughly equivalent to the current naming convention if I'm not wrong.

parsonsmatt:
:+1: Good call @psibi. I've implemented that change.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1014,3 +1019,107 @@ mockMigration mig = do
...
+-- that a key conflicts with a record currently in the database, the second and
+-- third parameters determine what will happen.
+--
+-- The second parameter is a list of fields to copy from the original value.
+-- This allows you to specify that, when a collision occurs, you'll just update
+-- the value in the database with the field values that you inserted.
+--
+-- The third parameter is a list of updates to perform that are independent of
+-- the value that is provided. You can use this to increment a counter value.
+-- These updates only occur if the original record is present in the database.
```
MaxGabriel: The documentation for the third parameter is kind of confusing to me

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1014,3 +1019,107 @@ mockMigration mig = do
...
+mkUpdateText :: PersistEntity record => Update record -> Text
+mkUpdateText x =
+  case updateUpdate x of
+    Assign -> n <> "=?"
+    Add -> T.concat [n, "=", n, "+?"]
+    Subtract -> T.concat [n, "=", n, "-?"]
+    Multiply -> T.concat [n, "=", n, "*?"]
+    Divide -> T.concat [n, "=", n, "/?"]
+    BackendSpecificUpdate up ->
+      error . T.unpack $ "BackendSpecificUpdate" <> up <> "not supported"
```
MaxGabriel: Should there be a space after `BackendSpecificUpdate` and before `not supported`?

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1014,3 +1019,107 @@ mockMigration mig = do
...
+  where
+    n = T.pack . escapeDBName . fieldDB . updateFieldDef $ x
+
+commaSeparated :: [Text] -> Text
+commaSeparated = T.intercalate ", "
+
+parenWrapped :: Text -> Text
```
MaxGabriel: These functions make the code much more readable, nice idea

persistent-mysql/Database/Persist/MySQL.hs
parsonsmatt: Probably! It's vendored from part of the `update` function in `persistent`. I could push it further upstream and refactor that code to use it as well.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1014,3 +1019,107 @@ mockMigration mig = do
...
+parenWrapped = ("(" <>) . (<> ")")
```
MaxGabriel: Not that it's a huge deal since this function is so small, but is point free really better here? It's longer than just `parenWrapped t = "(" <> t <> ")"`

persistent-mysql/Database/Persist/MySQL.hs
parsonsmatt: It's a little difficult to explain. If you pass something like

```
insertManyOnDuplicateKeyUpdate _ [SomeField UserName] [UserAge +=. 1]
```
then it generates the following clause in the Update:

```
INSERT INTO ... VALUES ...
ON DUPLICATE KEY UPDATE ... 
  `user`.`name` = VALUES(`user`.`name`)
  `user`.`age` = `user`.`age` + 1
```

So the name gets copied from whatever record was being inserted and had a key collision, and the age gets incremented by 1.

Since the `Update` field requires a specific value, that means that we can't make it dependent on the record that we're updating. MySQL would support something like

```sql
ON DUPLICATE KEY UPDATE
  user.age = VALUES(user.age) + 2
```

which would add 2 to the record that hit the duplicate key. There's not a good way to make the update have more complicated values, unfortunately, so this feature would be difficult to incorporate as-is.

persistent-mysql/Database/Persist/MySQL.hs
parsonsmatt: Whenever I get the chance, I try to make smiley faces in code. Tbh `T.concat ["(", t, ")"]` is the tiniest bit cleaner and probably more efficient.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -720,6 +724,7 @@ showColumn (Column n nu t def _defConstraintName maxLen ref) = concat
         Just s -> -- Avoid DEFAULT NULL, since it is always unnecessary, and is an error for text/blob fields
                   if T.toUpper s == "NULL" then ""
                   else " DEFAULT " ++ T.unpack s
+                  {-# LANGUAGE GADTs #-}
```
MaxGabriel: Is this intentionally here?

parsonsmatt: oh, no, that is definitely a mistake

persistent-mysql/Database/Persist/MySQL.hs
parsonsmatt: Thanks! I can use them in the other instances of these if you'd like.

MaxGabriel:
LGTM

gregwebs:
awesome! Can this be used to fill out `upsert` in the typeclass?

parsonsmatt:
Unfortunately not, as this doesn't return the updated records. We'd need something like `upsert_ :: record -. [Update record] -> SqlPersistT m ()` for this to make it into the class.

gregwebs:
That could make sense to add `upsert_` then. Could you also add a SELECT in the same transaction to get the `upsert` behavior though?

parsonsmatt:
Ooh, I think that would work. I'm not sure it'd be any more efficient than the current default implementation which does a `SELECT` and then an `UPDATE` if present and `INSERT` if missing -- still two queries.

psibi:
@parsonsmatt I just looked on the code. Is there any reason why native upsert for mysql has been done like this instead of going via the typeclass route: https://www.stackage.org/haddock/lts-8.17/persistent-2.6.1/Database-Persist-Sql.html#t:SqlBackend

The function `connUpsertSql` there has been specifically created for implementing native upsert feature for backends. In fact, Postgres right now uses that. Sorry for commenting about this so late.

parsonsmatt:
@psibi The type class function `upsert` returns the updated or inserted record. MySQL's `ON DUPLICATE KEY UPDATE` doesn't have any way of returning a value, so we can't do that. If the type class had `upsert_ :: rec -> [Update rec] -> SqlPersistT m ()`, then this would fit that signature.

gregwebs:
@psibi I am in favor of adding `upsert_`.

psibi:
Yeah, `upsert_` would be good. It can be probably added as a typeclass method and a custom function be passed as a part of `SqlBackend` type which will generate the appropriate query for the specific backend. For those backends, for which the function hasn't been passed, `upsert_` can simply be defined as `upsert >> return ()`.

gregwebs:
@psibi can you merge and release this as is?

psibi:
`persistent-mysql-2.6.1` has been released with the relevant CHANGELOG in Hackage.

PR #677, opened 2017-06-15
Add detail error message while invoking sqlite step.

highfly22:


MaxGabriel:
This ended up going in in #743. Thanks for your contribution @highfly22!

PR #678, opened 2017-06-16
Added support of MEDIUM- and LONG text and blob types in MySQL

qrilka:


qrilka:
@snoyberg any updates on this one?

snoyberg:
I didn't see this one, I'm not really maintaining persistent day-to-day anymore.

That said: I don't really think _anyone_ is stepping up to maintain the MySQL backend itself very much right now (maybe I'm wrong? speak up if this isn't true). If you're interested in taking an active role in maintainership here, I think it would be appreciated.

I don't see a problem with this, so I'll merge it in.

snoyberg:
Thanks!

qrilka:
thanks @snoyberg 
Cabal file lists @meteficha as a maintainer (and you between the authors) but as I remember he wasn't available for contact for quite some time...
I'm not sure about actively maintaining this package as of this moment.
@paul-rouse you were the author of latest fixes to `persistent-mysql` as far as I remember and also you maintain `mysql-simple` - would it be much of a burden to play some role here too?

paul-rouse:
Yes, I do have an interest in keeping `persistent-mysql` running, so I will try to step in when things crop up.

paul-rouse:
@qrilka What was the reason for this change?  We are considering reverting it because it introduced a migration problem (#752), and, as I note in that issue, I have had no problem using `longtext` etc without this change.

qrilka:
@paul-rouse could you explain how could you use `longtext` without this change? Specify manually everywhere? Probably the reason of that PR was the fact that I was either not aware of `sqltype` option in model definitions or I just forgot about it

MaxGabriel:
Even with this change, you'd still need to specify `longtext` using the `sqltype` option

paul-rouse:
Yes @qrilka, I am talking about adding `sqltype=longtext` to the model in places where you need to change the default mapping.

PR #681, opened 2017-07-04
Update comment to reflect the state of things

dfithian:
The comment on `mpsGeneric` in `MkPersistSettings` says it defaults to `True`, but [here](https://github.com/tvision-insights/persistent/blob/mps-generic-comment/persistent-template/Database/Persist/TH.hs#L373) we can see that it defaults to `False`.


psibi:
Thanks!

PR #686, opened 2017-07-22
[persistent-postgresql] Recognize column-widths

JaSpa:
Before this change: If a custom `PersistFieldEntity` specifies a type of `bit(3)` for example, its type was migrated on every app start.

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -644,6 +644,7 @@ getColumns getter def = do
                           ,",column_default "
                           ,",numeric_precision "
                           ,",numeric_scale "
+                          ,",character_maximum_length "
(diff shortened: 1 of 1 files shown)


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -742,24 +743,27 @@ getAlters defs def (c1, u1) (c2, u2) =
...
+                  Right t -> do
+                      let cname = DBName x
+                      ref <- getRef cname
+                      return $ Right Column
+                          { cName = cname
+                          , cNull = y == "YES"
+                          , cSqlType = t
+                          , cDefault = fmap stripSuffixes d''
+                          , cDefaultConstraintName = Nothing
+                          , cMaxLen = Nothing
```
MaxGabriel: Should this remain `Nothing`?

PR #688, opened 2017-07-28
remove obsolete mongo demo

gelisam:
This code is bitrotten, and does more harm than good to those stumbling
upon it while trying to find an example of how to use persistent-mongoDB.

According to [this video](https://begriffs.com/posts/2014-11-18-typesafe-db-persistent.html) at 6:43, the best way to figure things out in persistent is to look at the test suite. So I looked at the persistent repo, saw that there was both a `persistent-test` and a `persistent-mongoDB/test`, and decided to look at the latter because I was interested in the tests for persistent-mongoDB. I wasted a lot of time trying to understand why inserting that code in my code base didn't work, until I eventually checked out the persistent repo to double-check what I took for granted, that persistent's official test suite was indeed compiling with the latest version of persistent. And that's when I finally realized that `mongoapp.hs` is outside of persistent's official testsuite.

The file is terribly bitrotten. It uses Control.Monad.Context, which only existed until mongoDB-0.10.0, and it uses a version of `mkPersist` from persistent-template-0.5.1 at the latest because it doesn't take an `MkPersistSettings` argument. These are so old that and even lts-0.0 has a more modern version of those packages than that. This example file is providing negative value to your users, please delete it (or modernize it and put it in your test suite to make sure it remains up to date).


parsonsmatt:
I'm in favor of this change. Examples are great, but the worst thing an example can do is lead someone astray.

psibi:
Can we instead fix this ? I can volunteer to fix this if nobody comes forward.

gregwebs:
lets merge now, That doesn't prevent fixing this or providing better examples.

PR #690, opened 2017-08-16
A few markup improvements

iustin:
Fixes a few wrong haddock `@since` markups, moves to always use `@since` (vs. mixed Since/`@since`), and cleans up some whitespace issues.


psibi:
Thanks!

PR #693, opened 2017-09-01
[persistent-mysql] Add support for conditional copying

parsonsmatt:
This PR extends the `insertManyOnDuplicateKeyUpdate` function to allow conditional copying of data from the input record.

Use case: Suppose you're importing a report that sometimes has partial data, and you want to copy the report into the database. The result record may have many `Nothing` fields. If we copy this using the current `SomeField`, then we will overwrite the existing data with `NULL`s. Oops.

The new constructor and functions permit us to avoid copying bad data, including `NULL` and empty strings, into the database, potentially overriding already present data.

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -18,6 +18,8 @@ module Database.Persist.MySQL
     , insertOnDuplicateKeyUpdate
     , insertManyOnDuplicateKeyUpdate
     , SomeField(..)
+    , copyUnlessNull
+    , copyUnlessEmpty
     ) where
 
 import Control.Arrow
@@ -27,11 +29,11 @@ import Control.Monad.Trans.Class (lift)
 import Control.Monad.Trans.Except (runExceptT)
 import Control.Monad.Trans.Reader (runReaderT)
 import Control.Monad.Trans.Writer (runWriterT)
+import Data.Either (partitionEithers)
 import Data.Monoid ((<>))
 import Data.Aeson
 import Data.Aeson.Types (modifyFailure)
 import Data.ByteString (ByteString)
-import Data.Either (partitionEithers)
 import Data.Fixed (Pico)
 import Data.Function (on)
 import Data.IORef
@@ -1034,10 +1036,25 @@ insertOnDuplicateKeyUpdate
 insertOnDuplicateKeyUpdate record =
   insertManyOnDuplicateKeyUpdate [record] []
 
--- | This wraps values of an Entity's 'EntityField', making them have the same
--- type. This allows them to be put in lists.
+-- | This type is used to determine how to update rows using MySQL's
+-- @INSERT ON DUPLICATE KEY UPDATE@ functionality, exposed via
+-- 'insertManyOnDuplicateKeyUpdate' in the library.
 data SomeField record where
   SomeField :: EntityField record typ -> SomeField record
+  -- ^ Copy the field directly from the record.
+  CopyUnlessEq :: PersistField typ => EntityField record typ -> typ -> SomeField record
+  -- ^ Only copy the field if it is not equal to the provided value.
+
+-- | Copy the field into the database only if the value in the
+-- corresponding record is non-@NULL@.
+copyUnlessNull :: PersistField typ => EntityField record (Maybe typ) -> SomeField record
+copyUnlessNull field = CopyUnlessEq field Nothing
+
+-- | Copy the field into the database only if the value in the
+-- corresponding record is non-empty, where "empty" means the Monoid
+-- definition for 'mempty'. Useful for 'Text', 'String', 'ByteString', etc.
+copyUnlessEmpty :: (Monoid typ, PersistField typ) => EntityField record typ -> SomeField record
+copyUnlessEmpty field = CopyUnlessEq field mempty
 
 -- | Do a bulk insert on the given records in the first parameter. In the event
 -- that a key conflicts with a record currently in the database, the second and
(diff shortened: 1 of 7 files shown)


psibi:
Sorry, but can you actually expand the use case you have explained with an example ? I'm having a difficult time following the problem this solves.

Also in the code, you seem to have deleted the existing haddock comment for the type `SomeField`. Has it become outdated or changed with this feature you have added ?

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1034,10 +1037,25 @@ insertOnDuplicateKeyUpdate
...
+-- | This type is used to determine how to update rows using MySQL's
+-- @INSERT ON DUPLICATE KEY UPDATE@ functionality, exposed via
+-- 'insertManyOnDuplicateKeyUpdate' in the library.
 data SomeField record where
   SomeField :: EntityField record typ -> SomeField record
+  -- ^ Copy the field directly from the record.
+  CopyUnlessEq :: PersistField typ => EntityField record typ -> typ -> SomeField record
+  -- ^ Only copy the field if it is not equal to the provided value.
+
+-- | Copy the field into the database only if the value in the
```
psibi: Can you bump the cabal version, update changelog and add the `@since` haddock syntax for all the functions you have added here.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1034,10 +1037,25 @@ insertOnDuplicateKeyUpdate
...
+-- corresponding record is non-@NULL@.
+copyUnlessNull :: PersistField typ => EntityField record (Maybe typ) -> SomeField record
+copyUnlessNull field = CopyUnlessEq field Nothing
+
+-- | Copy the field into the database only if the value in the
+-- corresponding record is non-empty, where "empty" means the Monoid
+-- definition for 'mempty'. Useful for 'Text', 'String', 'ByteString', etc.
+copyUnlessEmpty :: (Monoid typ, PersistField typ) => EntityField record typ -> SomeField record
+copyUnlessEmpty field = CopyUnlessEq field mempty
```
psibi: Both of the introduced functions are only seem to be applicable for `insertManyOnDuplicateKeyUpdate`. Can you mention that in the haddock comments? Also can you add the example you mentioned here: https://github.com/yesodweb/persistent/pull/693#issuecomment-326688351 and explain it's usecase in  a similar way we have done it here: [persistent doc](https://www.stackage.org/haddock/lts-7.0/persistent-2.6/Database-Persist.html#v:-43--61-..)

psibi:
Thanks for the explanation, that was helpful.

> 3. If the new value to be inserted is non-null, then we copy the value into the existing database table.

You mean update the existing row ?

persistent-mysql/Database/Persist/MySQL.hs
parsonsmatt: Sure thing! Is adding a constructor like this a minor version bump?

psibi: You need to bump it to `2.6.2`. More info here: http://www.snoyman.com/blog/2017/06/how-to-send-me-a-pull-request

parsonsmatt:
> You mean update the existing row ?

Yes, that's correct.

I've added the documentation requested along with version bump. Thanks!

psibi:
LGTM. I will merge this once it is reviewed by one more person. CC: @gregwebs @snoyberg and others.

parsonsmatt:
The travis build is failing with

```
persistent-test: ConnectionError {errFunction = "connect", errNumber = 1044, errMessage = "Access denied for user 'travis'@'%' to database 'persistent'"}
Process exited with ExitFailure 1: /home/travis/build/yesodweb/persistent/.stack-work/install/x86_64-linux/lts-2.22/7.8.4/bin/persistent-test
```

Any idea what the fix for that might be?

paul-rouse:
About travis: the last successful run seems to have been on *precise*, but they have recently finished changing the default to *trusty*: https://blog.travis-ci.com/2017-08-31-trusty-as-default-status.

I'm guessing we are suffering a trysty/mysql problem along the lines of https://github.com/travis-ci/travis-ci/issues/8331. The images are being updated on Wednesday - https://docs.travis-ci.com/user/build-environment-updates/2017-09-06/ - so perhaps it will work then!

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1034,29 +1039,129 @@ insertOnDuplicateKeyUpdate
...
--- type. This allows them to be put in lists.
+-- | This type is used to determine how to update rows using MySQL's
+-- @INSERT ON DUPLICATE KEY UPDATE@ functionality, exposed via
+-- 'insertManyOnDuplicateKeyUpdate' in the library.
 data SomeField record where
   SomeField :: EntityField record typ -> SomeField record
+  -- ^ Copy the field directly from the record.
+  CopyUnlessEq :: PersistField typ => EntityField record typ -> typ -> SomeField record
+  -- ^ Only copy the field if it is not equal to the provided value.
+  -- /Since 2.6.2/
```
snoyberg: I'd recommend using the `@since 2.6.2` syntax

parsonsmatt: TIL, thanks!

parsonsmatt:
I'll implement the documentation example as a test.

parsonsmatt:
I made a change to the implementation. I removed the `CopyUnlessEq` data constructor from the export list and am exporting a function instead.

I'm pretty sure I'm going to want to expand this to `copyUnlessIn :: PersistField rec typ -> [typ] -> SomeField rec` in the near future, and then `CopyUnlessEq` becomes a special case of that.

parsonsmatt:
Good call, I didn't understand exactly what was going on there. I've reverted that change and used `round` instead of `truncate`, which should give us the right answers, as MySQL appears to be doing a `round` instead of `truncate`.

I can also revert the unrelated test changes and leave that for another PR, if you're OK with an unrelated red test.

parsonsmatt:
`et tu, sqlite`? :stuck_out_tongue: 

Well, MySQL passes, and SQLite is just dying for no reason now.

paul-rouse:
I should leave the tests as they were, and we'll live with them failing for now.  I'd like a little while to think about `persistent-mysql` itself, now that fractional seconds could be supported.  If we (optionally?) changed the mapping to SQL types so as to include the microseconds, the problem would go away.  It might cause too much breakage, though, and that's what I would like to think through - any comments are welcome, of course!

parsonsmatt:
I've reverted the test changes.

parsonsmatt:
Is there anything I can do to help get this merged? :)

paul-rouse:
OK by me - are you happy too @psibi ?  We know why the test fails - for unrelated reasons - and I am addressing that separately.

psibi:
@paul-rouse LGTM. Please go ahead and merge.

paul-rouse:
Thanks both of you!

PR #695, opened 2017-09-07
expose parseReferences for the creation of custom QuasiQuoters

sbditto85:
Exposed parseReferences so that custom QuasiQuoters could be made. If there is a reason not to do this please reject the PR.


sbditto85:
Looks like the build failures are independent of what I did.

paul-rouse:
I don't see any objection to this, and no-one else has chimed in with one.  The test failure is a separate issue.

sbditto85:
I have attempted to do what you've asked, if there is any problems let me know!

paul-rouse:
Thanks!

Issue #698, opened 2017-09-12
Question: Type safety with multiple `SqlBackend`s


parsonsmatt:
I'm working on a library [persistent-typed-db](https://github.com/parsonsmatt/persistent-typed-db) which is an attempt to provide type-safe `persistent` access to multiple SQL databases that are managed by the same app. This is accomplished by wrapping `SqlBackend` with a newtype and a phantom type parameter that indicates the database. This would ensure that you only use a `SqlBackend` with the appropriate models.

Motivation: Our application architecture uses a main database along with an account-specific shard. The main database and account-specific databases have totally different models. Running a `SqlPersistT` action on the wrong database has bitten us a few times, and we're about to spin up an admin interface that will be able to connect to at least four (and probably more) databases. I want type safety around this.

At a first pass, I attempted to use the `BaseBackend` and `IsSqlBackend` classes for the `SqlReadBackend` and `SqlWriteBackend`, but that ended up failing.

```haskell
-- imagine this:
newtype SqlBackendFor db = SqlBackendFor { unSqlBackendFor :: SqlBackend }

instance BaseBackend (SqlBackendFor db) where
  type BaseBackend (SqlBackendFor db) = SqlBackend
  persistBackend = unSqlBackendFor
```

This is the definition I used. We come across the problem in the way `PersistRecordBackend` requires the backend of the record to be equal to the *`BaseBackend`* of the backend, not the backend itself:

```haskell
-- | A convenient alias for common type signatures
type PersistRecordBackend record backend 
  = ( PersistEntity record
    , PersistEntityBackend record ~ BaseBackend backend
    )

-- snip

    -- | Get a record by identifier, if available.
    get :: (MonadIO m, PersistRecordBackend record backend)
        => Key record -> ReaderT backend m (Maybe record)
```

So I can't implement `PersistStoreRead` and have `ReaderT (SqlBackendFor MainDb) m (Maybe record)`.

At this point, I don't anticipate my library being anything more than vendoring Persistent class instances for my newtype. We're going to hammer this design out at work, but if this is something that ends up being useful, it might make more sense to upstream it directly to Persistent.

parsonsmatt:
It does seem like the newtype wrapper around `SqlBackend` is preferable for a variety of reasons. I'm working on figuring out the best way to fix compatibility with Esqueleto, and I believe it'll come in changing the `PersistEntityBackend record ~ BaseBackend backend` clause to something a little more flexible.

parsonsmatt:
[...]

PR #699, opened 2017-09-13
WIP/RFC: Add a phantom type to SqlBackend to track which database it is for

parsonsmatt:
This PR adds a phantom type to the SqlBackend type, addressing #698 . See that issue for motivation etc.

# Why?

I already wrote a library for this. Why am I upstreaming it? My library is 99% vendored code from Persistent. Additionally, Esqueleto has a hard constraint on `SqlBackend`, so the library version of the code can't work without *entirely forking Esqueleto*. Upstreaming the code allows for Esqueleto to continue working with Persistent in a way that's nice

# Why not?

This is a breaking change, and also probably not the best way to approach it. I would like to improve the design so that the extension isn't noisy or unnecessary for the common/sane case (a single schema), but as-is, it's going to be pretty heavy weight.

# Your thoughts?

I'm curious if 1) there are better ways to approach this, 2) the likelihood of this getting merged, and 3) any objections or work to be done that would make merging this easier.

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -69,7 +69,7 @@ import Control.Monad.Trans.Resource (runResourceT)
 -- The pool is properly released after the action finishes using
 -- it.  Note that you should not use the given 'ConnectionPool'
 -- outside the action since it may be already been released.
-withMySQLPool :: (MonadIO m, MonadLogger m, MonadBaseControl IO m, IsSqlBackend backend)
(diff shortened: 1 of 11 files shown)


parsonsmatt:
I've made a fork of Esqueleto that uses the typed-db backend, and *actually* the fundeps on the Esqueleto, FromPreprocess, Select etc. classes wig out in the presence of a type variable. Essentially, changing the instance to:

```haskell
instance Esqueleto SqlQuery SqlExpr (SqlFor db) where
  ...
```

causes the types to not be fully determined since `db` is a type variable. So this PR would actually 100% break esqueleto. I'm going to try another route on this.

PR #701, opened 2017-09-15
Introduce `BackendCompatible` class

parsonsmatt:
This PR introduces some machinery for generalizing the SQL backend requirement, and uses that to generalize the `insertOnDuplicateKeyUpdate` functions.

This PR enables support for the [`persistent-typed-db`](https://github.com/parsonsmatt/persistent-typed-db) library which wraps the `SqlBackend` with a phantom type to indicate which database models are associated with. This turns mixing up your connection and models in a multi-db environment into a compile time error.

This PR is based on [work done in a PR to `esqueleto`](https://github.com/bitemyapp/esqueleto/pull/53). I was able to get esqueleto queries working by incorporating the `BackendCompatible` class to Esqueleto and instantiating it appropriately in the application.

The `BackendCompatible` class and the `HasPersistBackend` / `IsPersistBackend` classes are very similar, and it's likely that the functionality can be merged in some way.

PR #702, opened 2017-09-15
[persistent-mysql] Use an "INSERT IGNORE" query if the updates are empty.

parsonsmatt:
The current behavior of `insertManyOnDuplicateKeyUpdate` will perform an `insertMany_` if there are no updates to perform. `insertMany_` will then throw an error if the key already exists in the database.

This is surprising -- typically, when you're doing a bulk upsert like this, the desired behavior is "Try to insert these rows into the database. If there's a key collision, then perform the set of updates." Given an empty set of updates, I would expect nothing to be done to the preexisting record, not throwing an exception and rolling back the entire insert.

MySQL's `INSERT IGNORE` is a little unfortunate: [docs](https://dev.mysql.com/doc/refman/5.5/en/insert.html)

> If you use the IGNORE modifier, errors that occur while executing the INSERT statement are ignored. For example, without IGNORE, a row that duplicates an existing UNIQUE index or PRIMARY KEY value in the table causes a duplicate-key error and the statement is aborted. With IGNORE, the row is discarded and no error occurs. Ignored errors may generate warnings instead, although duplicate-key errors do not.
IGNORE has a similar effect on inserts into partitioned tables where no partition matching a given value is found. Without IGNORE, such INSERT statements are aborted with an error. When INSERT IGNORE is used, the insert operation fails silently for rows containing the unmatched value, but inserts rows that are matched. For an example, see Section 19.2.2, “LIST Partitioning”.
Data conversions that would trigger errors abort the statement if IGNORE is not specified. With IGNORE, invalid values are adjusted to the closest values and inserted; warnings are produced but the statement does not abort. You can determine with the mysql_info() C API function how many rows were actually inserted into the table.

I believe the unfortunate bits of this API are hidden behind the guarantees that Persistent can make about the query we're generating.

### DIFF
--- a/persistent-mysql/ChangeLog.md
+++ b/persistent-mysql/ChangeLog.md
@@ -2,6 +2,7 @@
 
 * Extend the `SomeField` type to allow `insertManyOnDuplicateKeyUpdate` to conditionally copy values.
 * Depend on `mysql-simple >= 0.4.3` to fix encoding and decoding of date/time values with fractional seconds (when a column is specified using something like `sqltype=TIME(6)`).  See also [#705](https://github.com/yesodweb/persistent/issues/705)
+* Fix behavior of `insertManyOnDuplicateKeyUpdate` to ignore duplicate key exceptions when no updates specified.
 
 ## 2.6.1
 
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -1026,7 +1026,7 @@ mockMigration mig = do
 -- | MySQL specific 'upsert'. This will prevent multiple queries, when one will
 -- do.
 insertOnDuplicateKeyUpdate
-  :: (PersistEntity record , MonadIO m)
+  :: (PersistEntity record, MonadIO m)
   => record
   -> [Update record]
   -> SqlPersistT m ()
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -1058,30 +1058,8 @@ insertManyOnDuplicateKeyUpdate
   -> [Update record] -- ^ A list of the updates to apply that aren't dependent on the record being inserted.
   -> SqlPersistT m ()
 insertManyOnDuplicateKeyUpdate [] _ _ = return ()
-insertManyOnDuplicateKeyUpdate records [] [] =
-  uncurry rawExecute $ mkInsertIgnoreQuery records
 insertManyOnDuplicateKeyUpdate records fieldValues updates =
   uncurry rawExecute $ mkBulkInsertQuery records fieldValues updates
-
--- | This makes a special query that inserts the given rows into the
--- database without performing any updates. If there are duplicate keys,
--- then it just ignores that.
-mkInsertIgnoreQuery :: PersistEntity record => [record] -> (Text, [PersistValue])
-mkInsertIgnoreQuery records = (q, concat vals)
-  where
-    entityDef' = entityDef records
-    tableName = T.pack . escapeDBName . entityDB $ entityDef'
-    vals = map (map toPersistValue . toPersistFields) records
-    entityFieldNames = map (T.pack . escapeDBName . fieldDB) (entityFields entityDef')
-    recordPlaceholders = commaSeparated $ map (parenWrapped . commaSeparated . map (const "?") . toPersistFields) records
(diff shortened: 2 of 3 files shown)


parsonsmatt:
This PR is a subset of #701 and can be closed without merging if that PR is merged. PR #701 needed to incorporate these changes in order to unentangle the `PersistEntityBackend record ~ BaseBackend backend` constraint arising from `insertMany_`, which would block the use in `persistent-typed-db`.

If you choose not to merge #701, then merging this PR and Esqueleto merging [the BackendCompatible class](https://github.com/bitemyapp/esqueleto/pull/53) will work to get [`persistent-typed-db`](https://github.com/parsonsmatt/persistent-typed-db) fully compatible.

parsonsmatt:
I'm getting a weird test failure on the MySQL tests in an unrelated module.

```
 src/DataTypeTest.hs:66: 
  1) data type specs handles all types
       expected: ("time",14:19:40)
        but got: ("time",14:19:41)
```

The test handles MySQL specially:

```haskell
roundTime :: TimeOfDay -> TimeOfDay
#ifdef WITH_MYSQL
roundTime (TimeOfDay h m s) = TimeOfDay h m (fromIntegral (truncate s :: Integer))
#else
roundTime = id
#endif
```

Seems like updating the Travis image might have changed an expectation on how MySQL does times. 

When inserting values into the database, we use `truncateTimeOfDay`, but then we use `roundTime` when we compare. I suspect that this might be causing some of the difference.

paul-rouse:
The mysql test seems to have been broken by Travis recently upgrading their default Ubuntu images to "trusty".  As a result, MySQL will have been upgraded from 5.5.X to 5.7.X, assuming travis uses the standard versions for the distro.

Before MySQL 5.6.4, fractional seconds were not supported, and, if I understand correctly, any fractional part in a literal was simply ignored, thereby *truncating* to the whole second below - [documented here.](https://dev.mysql.com/doc/refman/5.5/en/fractional-seconds.html)

Since 5.4.6 the default is to use whole-second precision, but, crucially, excess precision in literals is now *rounded*, as described in the [5.7 documentation here.](https://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html)

So, MySQL itself has made a totally incompatible change.  The `persistent` test did the right thing for the old beviour, but now fails in the 50% of cases when the rounding goes upward.  Great:-(

I'll see if I can come up with a way of fixing the test...

parsonsmatt:
MySQL, you bring me so much joy :rage: 

Thanks for posting the rationale there. I figured it was a Travis change.

I replaced `truncate` with `round` in my other branch, which appears to have fixed the test failure for MySQL.

paul-rouse:
I realise #701 includes this change, but I feel more competent to comment on the MySQL part of it for now.

`INSERT IGNORE` is indeed a can of worms, and one thing about it does bother me: it suppresses errors you would expect in [strict mode](https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-strict) when invalid values are inserted.  Persistent does not protect against all of these.  A nasty example is trying to insert data which is too long for the default `text` or `blob` fields (which are only 64k) - I was bitten by that particular one in the days before the Yesod scaffolding added the code to set the MySQL session mode to `STRICT_ALL_TABLES`.

It is a little worrying that a behind-the-scenes use of `INSERT IGNORE` could nullify an explicit strict-mode setting!

Is it possible to use `INSERT ... ON DUPLICATE KEY UPDATE` instead?  You have to supply an update, but it can be a no-op.  For example, if the table `foo` has a column `bar` you can say `INSERT INTO foo ... ON DUPLICATE KEY UPDATE bar = bar`.

BTW I don't understand the statement in the [docs](https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html) about InnoDB tables and auto-increment columns - I have just tried it and I don't see that behaviour!

parsonsmatt:
That's a fantastic point that I hadn't considered. Thanks! I'll modify the PR to use `ON DUPLICATE KEY UPDATE` with a dummy field set.

parsonsmatt:
Hooray for spurious CI failures :joy: 

This is passing all tests now, using the dummy update.

paul-rouse:
Thanks for including the changelog note as well!  I suggest I merge this now, and then if you could rebase #701 on this, we can look at the new type class separately, and ask one of the others for comments (I can't see any objections myself).

parsonsmatt:
Thanks! Will do :)

PR #265, opened 2014-08-08
Adds support for PostgreSQL's enumerated types

danclien:
This was my attempt at fixing Issue #264. I doubt it's as clean as it should be, but let me know if you have any suggestions.
### Changes
- Generalized `derivePersistField` into `derivePersistFieldGeneric` that accepts a `Q Exp` for the `PersistFieldSql` instance.
- Added `derivePersistFieldSqlTypeOther` which creates the `PersistFieldSql` instance with `SqlOther`.
- This fixes migrations always using `VARCHAR` in #264 .
#### PostgreSQL

```
CREATE TYPE color AS ENUM ('Red', 'Blue', 'Yellow');
```
#### Sum type

```
data Color = Red | Yellow | Blue deriving (Show, Read, Eq)
derivePersistFieldSqlTypeOther "Color" "color"
                           --     ^       ^
                           --     |       +--- Name of type in PostgreSQL
                           --     +----------- Name of type in Haskell
```
#### Table definition

```
share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
Person sql=people
    name String
    favorite_color Color default="'Blue'::color"
    deriving Show
|]
```
### Changes
- Adds a check into `getGetter` that sees if the incoming type in a PostgreSQL Enum.
- If it is an Enum, it converts the `ByteString` to a `PersistText` via UTF8.
- This fixes reading the Enum in #246.

PR #706, opened 2017-09-19
Add support for PostgreSQL's custom enumerated types.

### LINKED ISSUE #264: Add support for PostgreSQL's custom enumerated types
# Add support for PostgreSQL's custom enumerated types

PostgreSQL has ability to declare custom enumeration that's only accepts a predetermined set of strings. I have a potential solution which I'll send a pull request for in a bit.

[Official documentation](http://www.postgresql.org/docs/9.2/static/datatype-enum.html)
[Example from Postgres Guide](http://postgresguide.com/sexy/enums.html)
### Migrations

When using `derivePersistField`, the migration always wants to use a `varchar` which loses type safety in PosgreSQL.

```
Migrating: ALTER TABLE "people" ALTER COLUMN "favorite_color" TYPE VARCHAR
```
### Reading the value

Since the `Oid`is different for each custom enumerated typed , the call to `getGetter` in `Postgresql.hs` always throw the following error because it does not exist in `builtinGetters`:

```
Postgresql.getGetter: type ... not supported.
```
#### PostgresSQL

```
CREATE TYPE color AS ENUM ('Red', 'Blue', 'Yellow');

CREATE TABLE person (
  id SERIAL PRIMARY KEY,
  name varchar(256) NOT NULL,
  favorite_color color DEFAULT 'Blue'::color
);
```
#### Haskell

Main.hs

```
{-# LANGUAGE EmptyDataDecls    #-}
{-# LANGUAGE FlexibleContexts  #-}
{-# LANGUAGE GADTs             #-}
{-# LANGUAGE QuasiQuotes       #-}
{-# LANGUAGE TemplateHaskell   #-}
{-# LANGUAGE TypeFamilies      #-}

import Control.Monad.IO.Class (liftIO)
import Database.Persist
import Database.Persist.Postgresql
import Database.Persist.TH
import Color

Person
    name String
    favorite_color Color default="'Blue'::color"
    deriving Show
|]

connStr = "host=localhost dbname=schoolobjects user=root password=root"

main :: IO ()
main = withPostgresqlPool connStr 10 $ \pool -> do
    flip runSqlPersistMPool pool $ do
        runMigration migrateAll
        pid <- insert $ Person "Bob" Yellow
        mp <- get pid
        case mp of
            Nothing -> liftIO $ putStrLn "Person not found"
            Just person -> liftIO $ print person
```

Color.hs

```
{-# LANGUAGE TemplateHaskell #-}

module Color where

import Database.Persist.TH

data Color = Red | Yellow | Blue deriving (Show, Read, Eq)
derivePersistField "Color"
```

eugenk:
This shall fix #264 and replace #265. It builds on that pull request and uses @danclien's changes and rebases them on master.

Please tell me if I violate any code conventions of the projects or if other improvements should be made. :)

I'm not sure what's wrong with the build on Travis, but it seems to be the same on master.

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -760,6 +760,7 @@ showSqlType SqlString  Nothing    False = "TEXT"
 showSqlType SqlString  (Just i)   True  = "VARCHAR(" ++ show i ++ ") CHARACTER SET utf8"
 showSqlType SqlString  (Just i)   False = "VARCHAR(" ++ show i ++ ")"
 showSqlType SqlTime    _          _     = "TIME"
+showSqlType (SqlEnum t)  _        _     = T.unpack t
 showSqlType (SqlOther t) _        _     = T.unpack t
 
 -- | Render an action that must be done on the database.
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -917,6 +917,7 @@ showSqlType SqlTime = "TIME"
 showSqlType SqlDayTime = "TIMESTAMP WITH TIME ZONE"
 showSqlType SqlBlob = "BYTEA"
 showSqlType SqlBool = "BOOLEAN"
+showSqlType (SqlEnum t) = t
 
 -- Added for aliasing issues re: https://github.com/yesodweb/yesod/issues/682
 showSqlType (SqlOther (T.toLower -> "integer")) = "INT4"
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -34,6 +34,7 @@ import Database.Persist.Sql.Types.Internal (mkPersistBackend)
 import Data.Fixed (Pico)
 
 import qualified Database.PostgreSQL.Simple as PG
+import qualified Database.PostgreSQL.Simple.TypeInfo as PGT
 import qualified Database.PostgreSQL.Simple.TypeInfo.Static as PS
 import qualified Database.PostgreSQL.Simple.Internal as PG
 import qualified Database.PostgreSQL.Simple.ToField as PGTF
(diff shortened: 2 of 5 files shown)


psibi:
@eugenk I would still prefer this to be fixed in the way Greg Weber has proposed here: https://github.com/yesodweb/persistent/issues/264#issuecomment-51641354

I have a feel that the current PR misuses `SqlOther`.

eugenk:
Then, this is out of my abilities. I'm not familiar with the codebase of Persistent and do not want to mess things up. https://github.com/yesodweb/persistent/pull/265#issuecomment-283256005 made me think that a successful rebase might be enough :)

psibi:
Haven't looked much deeply, but I think it should be easy to adapt your code.

eugenk:
Adding and using the new SqlType is not a problem at all. What I'm concerned about are these two bullets:
> * modify the postgres backend migration code to use Enum
> * modify MySQL and sqlite migration code to use an Enum, but it could just be treated as a String for now.

psibi:
@eugenk Okay, so I took time to look into this issue in more detail. Apart from Greg Weber's comment, this PR lacks the creation logic for enumerated types in PostgreSQL. So, the PR will likely break all the existing code with sum types. So for this PR to get merged following things to be done:

* Table creation logic needs to be done for Enumerated types
* Migration logic needs to be added when new values are added to the enum, when the type name of the enum is changed etc.
* As far as possible, implement this without breaking backward compatibility.

eugenk:
I'm afraid, I'm not able to do this. Maybe this PR could be used as a starting point for someone else (@psibi?) to finish it, but I'm not sure what exactly to change in the migration code. Especially, the compatibility part is very difficult for me because I'm new to persistent.

psibi:
@eugenk Thanks for making this PR and making it alive again. I'm sorry for the earlier mis-communication that happened which made you work so far.

eugenk:
No worries ;)
I hope this can serve as a starting point (or at least inspiration) for a complete implementation.

eugenk:
@psibi, if you could point me to the lines that tie together something like
```haskell
data Enum = A | B deriving (Show, Read)
derivePersistField "Enum"
```
and
https://github.com/eugenk/persistent/blob/bbc92bb4a29f21729b0fba8a552537b77f43e031/persistent-postgresql/Database/Persist/Postgresql.hs#L562-L566
then I could try to write some migration code.

I mean, how does the migration know about all the `derivePersistField`s? And what exactly is a [`Q [Dec]`](https://github.com/eugenk/persistent/blob/bbc92bb4a29f21729b0fba8a552537b77f43e031/persistent-template/Database/Persist/TH.hs#L1298)?

MaxGabriel:
I think that `SqlEnum` will need to be:

```haskell
SqlEnum T.Text [T.Text] -- ^ Name of the enum, and its values.
```

Since to adequately describe the type of the enum, you need the name and every value. This will allow creating migrations from the type as well.

eugenk:
But how do I add an argument to https://github.com/eugenk/persistent/blob/bbc92bb4a29f21729b0fba8a552537b77f43e031/persistent-template/Database/Persist/TH.hs#L1331-L1332 to reflect this change? I have no Idea how TemplateHaskell works...

bitemyapp:
I'm going to close this without prejudice as I believe it has stalled out for now. If you'd like to resume work on this PR or someone else is prepared to help you with it please feel free to reopen.

Issue #705, opened 2017-09-17
Placeholder: MySQL fractional seconds in time/datetime


paul-rouse:
This is just a placeholder arising from the recent discussions about the test failures caused by MySQL changing its handling of fractional seconds in date/time types (https://github.com/yesodweb/persistent/pull/702#issuecomment-329965229, https://github.com/yesodweb/persistent/pull/693#issuecomment-329979698).

The idea is that `persistent-mysql` might move towards using SQL types like `datetime(6)`, now that MySQL itself supports types with fractional seconds (since 5.6.4).  Almost certainly an explicit option will be needed to enable this, and possibly a runtime test as well: the behaviour would be changed on those MySQL versions which do support it, and older versions of MySQL, which don't, are still in use (eg ["trusty"](https://packages.ubuntu.com/trusty/mysql-server)).

It is also an option to do nothing.  The documentation could be improved to explain the default behaviour and how it can be modified by using `sqltype=...`.

This issue is here while I am thinking about the options, and to invite comments!

paul-rouse:
I am tending towards the "do nothing" approach, together with a documentation improvement.  I think it may be better for the developer to be explicit about the storage types when faced with interrelated considerations including: changing existing application behaviour, migrating existing databases, and whether the application will always run against a MySQL server version which supports fractional seconds.

The downside is that any use of `sqltype` makes a `persistent` model less portable between different backends.  It is not obvious to me that people will use fractional seconds often enough for this to matter much.  However, I think the question is still open...

psibi:
It would be nice if you can link this issue on the changelog. Sorry for saying it late.

paul-rouse:
Good point!  I'll push an edit

paul-rouse:
Closing as there have been no comments for a month.  I have put a note in the documentation (#721).

PR #707, opened 2017-09-20
Update tests to allow for MySQL behaviour changes

paul-rouse:
This handles tests of fractional seconds in date/time types (introduced in MySQL 5.6.4) and a change to a strict SQL mode (introduced as the default in MySQL 5.7.5).  A lower-bound update on `mysql-simple` is needed for the fractional seconds.

### DIFF
--- a/persistent-mysql/ChangeLog.md
+++ b/persistent-mysql/ChangeLog.md
@@ -1,6 +1,7 @@
 ## 2.6.2
 
 * Extend the `SomeField` type to allow `insertManyOnDuplicateKeyUpdate` to conditionally copy values.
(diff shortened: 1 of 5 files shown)


paul-rouse:
@psibi would you mind casting an eye over this, in case I've done something silly (I do have a clean run of the tests at https://travis-ci.org/paul-rouse/persistent).  I think I'm heading for the "do nothing" approach in #705 , but I'll write a comment there explaining.

psibi:
LGTM. Thanks for taking care of this.

paul-rouse:
Thanks!

PR #709, opened 2017-09-22
Switch Travis to use VM environments to solve random test timeouts

paul-rouse:
See issue #710

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: enabled
 language: generic
 
 addons:


psibi:
IIRC, Michael moved travis from sudo enabled to container one explicitly to decrease build times.

paul-rouse:
I think this was a question of whether you can cache builds, which I believe used to be available only in the container environment.  As far as I can tell, this has changed, and caching now works in sudo-enabled builds as well - I did test a repeated build from my fork of persistent, and the second build really did go a lot faster.

Does that answer your concern?

paul-rouse:
Yes, a repeat run does go a good deal faster (3 or 4 times), so caching is certainly working.  The actual times look OK to me, too - do you agree @psibi?

psibi:
Yeah, the travis output looks good and caching is actually working. Free feel to merge.

paul-rouse:
Thanks for looking at it!

Issue #704, opened 2017-09-16
Proposal: Evolve away from existing persistent wiki


psibi:
Motivation: http://www.yesodweb.com/blog/2015/08/thoughts-on-documentation

There are two different options we can choose for migrating the existing persistent wiki:
1. Create a subfolder named `docs` inside `persitent` repository and move all the wiki content inside it.
2. Move the wiki to a new repository named `persistent-docs`

I personally prefer method one but it has a disadvantage that the wiki git history will be lost after migration.
What do others think ? I can volunteer to perform this migration if there is no concerns.

(Note that we previously migrated yesod-cookbook here: https://github.com/yesodweb/yesod-cookbook and we are doing good.)

paul-rouse:
The general idea looks good.  Option 1 probably makes the documentation more obvious to a newcomer, but I don't have a strong reason for preferring one option over the other.

psibi:
Wiki has been migrated and the old one now points to the new location.

PR #711, opened 2017-09-23
Migrate Persistent wiki

psibi:
See https://github.com/yesodweb/persistent/issues/704 for the motivation behind it.


paul-rouse:
I'm happy to look through this properly later if you want me to review it - can't give it enough time at the moment!

paul-rouse:
I spotted one small error, a general point about the page titles, and some outdated links.  Once you have decided what to do about them, I should merge it!

* There is still a wiki-style link in `Persistent-entity-syntax.md` in the note below the first table ("the section below about timezone support")
* I would suggest making the original wiki page title a first-level header, eg in `Persistent-entity-syntax.md` add `#Persistent entity syntax` at the start, and similarly in the other pages.

Some of the links refer to outdated/moved material.  I just mention them now in case you want to repair them at the same time as this migration:

* There are two links to specific package versions in the `JSON instances` section of `Persistent-entity-syntax.md` - the version numbers could be deleted so as to always get the latest version.
* There's a link to the cookbook at the start of the `SQLite` section of `Database-Configuration.md`, and it ought to be changed to point to the new version `https://github.com/yesodweb/yesod-cookbook/blob/master/cookbook/Activate-foreign-key-checking-in-Sqlite.md`

psibi:
Thanks for the detailed feedback. I have addressed all your comments in the latest push.

paul-rouse:
I can't see what's wrong but that link you corrected ("the section below about timezone support") doesn't seem to display properly when I view [your branch](https://github.com/psibi/persistent/blob/persistent-docs/docs/Persistent-entity-syntax.md).  Maybe because the url fragment contains spaces (and doesn't actually correspond to a defined anchor)?

psibi:
Sorry, my bad. It should be fixed right now.

paul-rouse:
Fantastic!  I don't have any more comments, so go ahead and merge!

psibi:
Thanks for the review!

PR #712, opened 2017-09-30
Link against sqlite using pkgconfig by default.

QuLogic:
There are two reasons for this:
1. The cabal user guide recommends [preferring pkg-config dependencies if available](https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-pkgconfig-depends)
2. Building a dependent package against persistent-sqlite that uses system libraries fails to fine `libsqlite3.so.0`. I don't quite understand the internals of why this occurs, but it works when using pkg-config.

I also added an extra flag to switch back to the previous behaviour, though I'm not so sure it's even necessary.

### DIFF
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,3 +1,7 @@
+## 2.6.3
+
+* Add 'use-pkgconfig' flag to use pkg-config to find system SQLite library.
+
 ## 2.6.2.1
 
 * Update `sqlite` cbit sources to 3.19.3 from 3.12.1
--- a/persistent-sqlite/persistent-sqlite.cabal
+++ b/persistent-sqlite/persistent-sqlite.cabal
@@ -1,5 +1,5 @@
 name:            persistent-sqlite
-version:         2.6.2.1
+version:         2.6.3
 license:         MIT
 license-file:    LICENSE
 author:          Michael Snoyman <michael@snoyman.com>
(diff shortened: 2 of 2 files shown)


psibi:
We use a slightly forked sqlite for persistent which is bundled with it. So, I'm not sure if this is a good idea.

paul-rouse:
@psibi this change is effective only when the flag `systemlib` is set `true` - as you say, that is not the default situation.

However, I am not convinced the default behaviour should change when someone *has* explicitly set `systemlib`.  My preference would be to keep `systemlib` working the way it does now by default, and only use `pkg-config` when explicitly set by another flag - so the extra flag would work the other way round from the one in this PR.

QuLogic:
Though I prefer pkg-config, I was 50/50 on which would be preferable in Go. I chose to default to pkg-config because the cabal user guide says to do so.

paul-rouse:
This change only applies to someone who has looked hard enough at the the build to know that they want to use `systemlib` (instead of the normal bundled C code).  They are probably not going to be happy with a sudden change to the meaning of their carefully chosen setting!

QuLogic:
So it turns out that the second "benefit" is actually due to [a ghc insufficiency](https://ghc.haskell.org/trac/ghc/ticket/9498) but I still think this is a good idea. I changed the flag to work in the opposite manner.

paul-rouse:
OK, could you please add a version bump in the cabal file, add a Changelog entry, and fix the `description` of the new flag in the cabal file?

paul-rouse:
Thanks!

Issue #716, opened 2017-10-11
Feature request: Named parameter substitutions in rawSql


felixSchl:
Have you considered adding support for named substitutions for `rawSql` at all? Complications with position based approach arise when parameter is used more that once etc. I guess I could use `CREATE FUNCTION` or something like that, but it'd be nice if I didn't have to...

```haskell
      rawSql
        [s|
          DELETE FROM category WHERE lft BETWEEN ? AND ?;
          UPDATE category SET rgt = rgt - ? WHERE rgt > ?;
          UPDATE category SET lft = lft - ? WHERE lft > ?;
        |] [ toPersistValue 10
           , toPersistValue 20
           , toPersistValue 50
           , toPersistValue 20
           , toPersistValue 50
           , toPersistValue 20 ]
```

Versus

```hsakell
      rawSql
        [s|
          DELETE FROM category WHERE lft BETWEEN ?lft AND ?rgt;
          UPDATE category SET rgt = rgt - ?width WHERE rgt > ?rgt;
          UPDATE category SET lft = lft - ?width WHERE lft > ?rgt;
        |] [ ("lft", toPersistValue 10)
           , ("rgt", toPersistValue 20)
           , ("width", toPersistValue 50)
           ]
```

psibi:
The idea sounds good. Would you be interested in creating a PR for that ? Note that the old behavior should still be retained for backwards compatibility.

felixSchl:
I had a quick look `stmtQuery` runs pretty deep and is used virtually everywhere using positional arguments. I think to achieve what I have described above, we'd implement a function that creates a normal `stmtQuery` compatible text and set of values. Doees that sound right?

felixSchl:
I wonder if this was possible:

```  
let lft = 10
     rgt = 20
     width = 50
rawSql
  [rawsql|
    DELETE FROM category WHERE lft BETWEEN #{lft} AND #{rgt};
    UPDATE category SET rgt = rgt - #{width} WHERE rgt > #{rgt};
    UPDATE category SET lft = lft - #{width} WHERE lft > #{rgt};
|]
```

The transformation would keep note of the order of the occurrence and then construct a string and replacement values as expected by the current `rawSql`

PR #717, opened 2017-10-12
Implement rawSql as a QuasiQuoter

felixSchl:
I'm having a go at #716, now you can do:

```haskell
let lft = 10 :: Int
     rgt = 20 :: Int
     width = 50 :: Int
[executeQQ|
    DELETE FROM category WHERE lft BETWEEN #{lft} AND #{rgt};
    UPDATE category SET rgt = rgt - #{width} WHERE rgt > #{rgt};
    UPDATE category SET lft = lft - #{width} WHERE lft > #{rgt};
|]
```

Would be great if you could help me figure out how to pass a function into `makeExpr` so that I don't have to duplicate the whole block of code:

https://github.com/felixSchl/persistent/blob/7856bd3878aadb9a703360d76934a7b7e78c66a4/persistent/Database/Persist/Sql/Raw/QQ.hs#L37-L49

I've also updated the test bed to include a few basic tests for this:

https://github.com/felixSchl/persistent/blob/7856bd3878aadb9a703360d76934a7b7e78c66a4/persistent-test/src/PersistentTest.hs#L891-L913

### DIFF
--- a/persistent/Database/Persist/Sql/Raw/QQ.hs
+++ b/persistent/Database/Persist/Sql/Raw/QQ.hs
@@ -1,9 +1,64 @@
+{-|
+Module: module Database.Persist.Sql.Raw.QQ
+Description: QuasiQuoters for performing raw sql queries
+
+This module exports convenient QuasiQuoters to perform raw SQL queries.
+All QuasiQuoters follow them same pattern and are analogous to the similar named
+functions exported from 'Database.Persist.Sql.Raw'. Neither the quoted
+function's behaviour nor it's return value is altered during the translation and
+all documentation provided with these functions holds.
+
+The QuasiQuoters in this module perform a simple substitution on the query text,
+where interpolated values of the form @#{foo}@ are replaced with a single
+question mark ('?') and collected into a list. The targeted function is then
+applied to the modified query text as well as a list of all found values.
+Please note that it is not required to call 'toPersistValue' on each
+interpolated value, as this conversion is done automatically.
+
+Here is a small example:
+
+@
+   let lft = 10 :: Int
+       rgt = 20 :: Int
+       width = 50 :: Int
+   in [sqlQQ|
+           DELETE FROM category WHERE lft BETWEEN #{lft} AND #{rgt};
+           UPDATE category SET rgt = rgt - #{width} WHERE rgt > #{rgt};
+           UPDATE category SET lft = lft - #{width} WHERE lft > #{rgt};
+           |]
+@
+
+This directly translates to this:
+
+@
+   let lft = 10 :: Int
+       rgt = 20 :: Int
+       width = 50 :: Int
+   in rawSql (
+        "DELETE FROM category WHERE lft BETWEEN ? AND ?;"  <>
+        "UPDATE category SET rgt = rgt - ? WHERE rgt > ?;" <>
+        "UPDATE category SET lft = lft - ? WHERE lft > ?;"
+        ) [ toPersistValue lft
+          , toPersistValue rgt
+          , toPersistValue width
+          , toPersistValue rgt
+          , toPersistValue width
+          , toPersistValue rgt
+          ]
+@
+
+-}
+
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE OverloadedStrings #-}
 
 module Database.Persist.Sql.Raw.QQ (
-      sqlQQ
+      -- * Sql QuasiQuoters
+      queryQQ
+    , queryResQQ
+    , sqlQQ
     , executeQQ
+    , executeCountQQ
     ) where
 
 import Control.Arrow (first, second)
@@ -13,7 +68,7 @@ import Language.Haskell.TH.Quote
 import Language.Haskell.Meta.Parse
 
 import Database.Persist.Class (toPersistValue)
-import qualified Database.Persist.Sql.Raw as Raw
+import Database.Persist.Sql.Raw (rawSql, rawQuery, rawQueryRes, rawExecute, rawExecuteCount)
 
 data StringPart
   = Literal String
@@ -35,7 +90,7 @@ parseStr a ('#':'{':xs) = Literal (reverse a) : parseHaskell [] xs
 parseStr a (x:xs)       = parseStr (x:a) xs
 
 makeExpr :: TH.Q TH.Exp -> [StringPart] -> TH.ExpQ
-makeExpr n s = TH.appE (TH.appE [| flip (.) (first pack) |] (TH.appE [| uncurry |] n)) (go s)
+makeExpr f s = TH.appE [| uncurry $(f) . first pack |] (go s)
     where
     go [] = [| (mempty, []) |]
     go (Literal a:xs)   = TH.appE [| first (a ++) |] (go xs)
@@ -47,16 +102,34 @@ reify s =
         Left e -> TH.reportError e >> [| mempty |]
         Right v -> return v
 
-sqlQQ :: QuasiQuoter
-sqlQQ = QuasiQuoter
-    (makeExpr [| Raw.rawSql |] . parseStr [] . filter (/= '\r'))
+makeQQ :: TH.Q TH.Exp -> QuasiQuoter
+makeQQ x = QuasiQuoter
+    (makeExpr x . parseStr [])
     (error "Cannot use qc as a pattern")
     (error "Cannot use qc as a type")
     (error "Cannot use qc as a dec")
 
+-- | Analoguous to 'Database.Persist.Sql.Raw.rawSql'
+-- @since 2.7.2
+sqlQQ :: QuasiQuoter
+sqlQQ = makeQQ [| rawSql |]
+
+-- | Analoguous to 'Database.Persist.Sql.Raw.rawExecute'
+-- @since 2.7.2
 executeQQ :: QuasiQuoter
-executeQQ = QuasiQuoter
-    (makeExpr [| Raw.executeSql |] . parseStr [] . filter (/= '\r'))
-    (error "Cannot use qc as a pattern")
-    (error "Cannot use qc as a type")
-    (error "Cannot use qc as a dec")
+executeQQ = makeQQ [| rawExecute |]
+
+-- | Analoguous to 'Database.Persist.Sql.Raw.rawExecuteCount'
+-- @since 2.7.2
+executeCountQQ :: QuasiQuoter
+executeCountQQ = makeQQ [| rawExecuteCount |]
+
+-- | Analoguous to 'Database.Persist.Sql.Raw.rawQuery'
+-- @since 2.7.2
+queryQQ :: QuasiQuoter
+queryQQ = makeQQ [| rawQuery |]
+
+-- | Analoguous to 'Database.Persist.Sql.Raw.rawQueryRes'
+-- @since 2.7.2
+queryResQQ :: QuasiQuoter
+queryResQQ = makeQQ [| rawQueryRes |]
--- a/persistent/Database/Persist/Sql/Raw/QQ.hs
+++ b/persistent/Database/Persist/Sql/Raw/QQ.hs
@@ -62,7 +62,7 @@ import Control.Monad.Reader (ask)
 import Control.Monad.Fail (fail)
 import Data.Text (pack, unpack)
 import Data.Maybe (fromMaybe, Maybe(..))
-import Data.Monoid (mempty)
+import Data.Monoid (mempty, (<>))
 import qualified Language.Haskell.TH as TH
 import Language.Haskell.TH.Quote
 import Language.Haskell.Meta.Parse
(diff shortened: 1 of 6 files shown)


cdepillabout:
@felixSchl Something like this should work:

```haskell
makeExpr :: TH.ExpQ -> [StringPart] -> TH.ExpQ
makeExpr f s = TH.appE [| uncurry $(f) . first pack |] (go s)
    where
    go [] = [| (mempty, []) |]
    go (Literal a:xs)   = TH.appE [| first (a ++) |] (go xs)
    go (AntiQuote a:xs) = TH.appE [| first ("?" ++) . second (toPersistValue $(reify a) :) |] (go xs)

sqlQQ :: QuasiQuoter
sqlQQ = QuasiQuoter
    (makeExpr [| Raw.rawSql |] . parseStr [] . filter (/= '\r'))
    (error "Cannot use qc as a pattern")
    (error "Cannot use qc as a type")
    (error "Cannot use qc as a dec")

executeQQ :: QuasiQuoter
executeQQ = QuasiQuoter
    (makeExpr [| Raw.rawExecute |] . parseStr [] . filter (/= '\r'))
    (error "Cannot use qc as a pattern")
    (error "Cannot use qc as a type")
    (error "Cannot use qc as a dec")
```

The problem is that you effectively want to pass a template haskell `Name` to `makeExpr`, instead of a regular function.  The code is arguably the cleanest if you package this `Name` up in an `Q Exp` within `executeQQ` and `sqlQQ` (instead of just passing a raw `Name`).  `makeExpr` can easily make use of this `Q Exp` representing the function you want to call.

Also, it would be really nice to have some documentation for the `sqlQQ` and `executeQQ` functions.  As an end-user, it's hard to use quasiquoters (and template-haskell stuff in general) if you don't know the types of the code being generated.  That is to say, the haddocks should explicitly state that `rawSql` produces a value of type `(RawSql a, MonadIO m) => ReaderT SqlBackend m [a]` (I think?).

Also a small example in the haddocks would probably help.

felixSchl:
Thanks @cdepillabout I managed to implement your advice, but I have doubts about the readability of this

https://github.com/yesodweb/persistent/pull/717/files#diff-7f47ab6d5194f8a630eab553fbf91a38R38

Still better than duplicate code I guess.

The transform is really, really simple. It just substitutes all `#{foo}` with question marks and collects them into  a list. For simplicity it also applies `toPersistValue` on each value, but that could be removed. It then simply applies `rawSql` or `rawExecute` to the substituted string and collected values

felixSchl:
@cdepillabout This is my first feature contribution to a Haskell project. Would you be able to advise me how / where to put that documentation?

persistent/Database/Persist/Sql/Raw/QQ.hs
```diff
@@ -0,0 +1,62 @@
...
+
+parseStr :: String -> String -> [StringPart]
+parseStr a []           = [Literal (reverse a)]
+parseStr a ('\\':x:xs)  = parseStr (x:a) xs
+parseStr a ['\\']       = parseStr ('\\':a) []
+parseStr a ('#':'{':xs) = Literal (reverse a) : parseHaskell [] xs
+parseStr a (x:xs)       = parseStr (x:a) xs
+
+makeExpr :: TH.Q TH.Exp -> [StringPart] -> TH.ExpQ
+makeExpr n s = TH.appE (TH.appE [| flip (.) (first pack) |] (TH.appE [| uncurry |] n)) (go s)
```
cdepillabout: > Thanks @cdepillabout I managed to implement your advice, but I have doubts about the readability of this

Yeah, I think you're right.  I haven't taken a hard look at the code, but it seems like you might not need to split the `(frist pack)` and `uncurry` stuff into two separate expressions?

Would something like the following work?  I've checked to make sure that this compiles, although I'm not sure if it passes the tests:

```haskell
makeExpr :: TH.ExpQ -> [StringPart] -> TH.ExpQ
makeExpr f s = TH.appE [| uncurry $(f) . first pack |] (go s)
    where
    go [] = [| (mempty, []) |]
    go (Literal a:xs)   = TH.appE [| first (a ++) |] (go xs)
    go (AntiQuote a:xs) = TH.appE [| first ("?" ++) . second (toPersistValue $(reify a) :) |] (go xs)

```

Or maybe you could even throw the `go` function inside the quasiquotes:

```haskell
makeExpr :: TH.ExpQ -> [StringPart] -> TH.ExpQ
makeExpr f s = [| uncurry $(f) $ first pack ( $(go s) ) |]
    where
    go [] = [| (mempty, []) |]
    go (Literal a:xs)   = TH.appE [| first (a ++) |] (go xs)
    go (AntiQuote a:xs) = TH.appE [| first ("?" ++) . second (toPersistValue $(reify a) :) |] (go xs)
```

cdepillabout:
@felixSchl 

> The transform is really, really simple. It just substitutes all #{foo} with question marks and collects them into a list. For simplicity it also applies toPersistValue on each value, but that could be removed. It then simply applies rawSql or rawExecute to the substituted string and collected values

Yeah, from looking at your example in the first comment, it does look like a very simple transformation.  I was trying to say that it would be nice to have haddocks showing the type of the value produced by `sqlQQ`.  You'll have to use the [haddock syntax]() to write this.

Here's an example of what I'm thinking about (using your explanation):

```haskell
-- | This is a quasiquoter for creating raw SQL expressions.
--
-- It substitutes all @#{foo}@ with question marks and collects them into a list. It applies
-- 'toPersistValue' on each value. It then simply applies 'rawSql' or 'rawExecute' to the
-- substituted string and collected values.
--
-- It produces a value of type 
-- @('RawSql' a, 'MonadIO' m) => 'ReaderT' 'SqlBackend' m [a]@.
--
-- Here is a small example:
--
-- @
--    let lft = 10 :: Int
--        rgt = 20 :: Int
--        width = 50 :: Int
--    in [sqlQQ|
--            DELETE FROM category WHERE lft BETWEEN #{lft} AND #{rgt};
--            UPDATE category SET rgt = rgt - #{width} WHERE rgt > #{rgt};
--            UPDATE category SET lft = lft - #{width} WHERE lft > #{rgt};
--            |]
-- @
sqlQQ :: QuasiQuoter
sqlQQ = QuasiQuoter ...
```

> This is my first feature contribution to a Haskell project.

Thanks a lot for your contribution!  It is very much appreciated. :+1:

felixSchl:
I am still keen to work on this, but I am also somewhat complacent because I was able to simple copy that code into a module for my project, so before proceeding would be great to hear the maintainers' opinions on where this PR should go.

paul-rouse:
Sorry, I'll help with it - we were probably all waiting to see if one of the other maintainers was more interested than ourselves:smile:

I don't see any problem with this, especially since it is almost all in a new module.  Could you please have a look at these things - I know a list looks daunting, but they should all be easy:

*   There is still an outstanding comment in the code diffs about `first pack` and `uncurry` - I think you should make that change, assuming it works.
*   Add haddock documentation.  At the module level, it would be good to show the intended usage, including something like your example from above.  Please also include `@since` comments in both the module and function docs (assume the new version will be 2.7.2).
*    Bump the version in the cabal file, and add a ChangeLog entry.

I'll check it again when you've done those things.  Thanks for the PR!

felixSchl:
I've made all the changes and simplified even further. I hope documenting each function becomes somewhat irrelevant now:

```haskell
-- | Analoguous to 'Database.Persist.Sql.Raw.rawSql'
-- @since 2.7.2
sqlQQ :: QuasiQuoter
sqlQQ = makeQQ [| rawSql |]
```

I haven't had time to run the tests locally, let's see what CI says. Also, @paul-rouse, I was not sure where to put the `@since` attribute in the module docs, top, bottom, after description? Seems random

paul-rouse:
Thanks for continuing to work on this!  The test failure is due to a weakness in the tests themselves, but I would like to see a clean run - give me a day or two, and I'll try to sort them out.  So far my attempts have only managed to produce a bug report against the latest pre-release version of `stack` ...

felixSchl:
Awesome, thanks. I actually hit another uncovered usecase just minutes ago:

```haskell
updateItemsInCart products cartPk = do
  escape <- connEscapeName <$> ask

  let tbl          = escape $ entityDB $ entityDef (Nothing :: Maybe CartItem)
      cartIdCol    = escape $ fieldDB  $ persistFieldDef CartItemCartId
      productIdCol = escape $ fieldDB  $ persistFieldDef CartItemProductId
      qtyCol       = escape $ fieldDB  $ persistFieldDef CartItemQuantity

  -- update the items
  forM_ products $ \(productPk, qty) ->
    if N.unpack qty <= 0
      then deleteBy $ CartItemUniqueCartIdProductId cartPk productPk
      else
        [executeQQ|
          INSERT INTO #{tbl} (#{cartIdCol}, #{productIdCol}, #{qtyCol})
          VALUES (#{cartPk}, #{productPk}, #{qty})
          ON CONFLICT (#{cartIdCol}, #{productIdCol})
          DO UPDATE SET
            #{qtyCol} = EXCLUDED.#{qtyCol}
        |]
```

This generates the following query:

```
[Debug#SQL] 
          INSERT INTO ? (?, ?, ?)
          VALUES (?, ?, ?)
          ON CONFLICT (?, ?)
          DO UPDATE SET
            ? = EXCLUDED.?
```

But my intention was:

```
[Debug#SQL] 
          INSERT INTO "cart_item" ("cart_id", "product_id", "quantity")
          VALUES (?, ?, ?)
          ON CONFLICT ("cart_id", "product_id", "quantity")
          DO UPDATE SET
            ? = EXCLUDED.?
```

felixSchl:
I think it would be cool if this was possible:

```haskell
  forM_ products $ \(productPk, qty) ->
    if N.unpack qty <= 0
      then deleteBy $ CartItemUniqueCartIdProductId cartPk productPk
      else
        [executeQQ|
          INSERT INTO ^{tbl} (@{CartItemCardId}, @{CartItemProductId}, @{CartItemQuantity})
          VALUES (#{cartPk}, #{productPk}, #{qty})
          ON CONFLICT (@{CartItemCardId}, @{CartItemProductId})
          DO UPDATE SET
            @{CartItemQuantity} = EXCLUDED.@{CartItemQuantity}
        |]
```

So `^{...}` does table references, `@{...}` does column references and `#{...}` continues to do value substitutions. It's all possible because we are in the `ReaderT`, so `connEscapeName <$> ask` should be possible to generate, but I ran into staging issues when trying to pass the escape function to `go`..

paul-rouse:
I have fixed the travis tests in master to work around the problem meeting the requirement for `happy` in the lts-2 case, so I think they should work next time you push a commit in this PR.  *However*, when I ran them in my own fork, plus your patch, I got a different error in the lts-2 case, and this *is* one you need to look at: [example log](https://travis-ci.org/paul-rouse/persistent/jobs/295678148).

felixSchl:
Hm, not sure how to fix the `mempty` bug. If I include it explicitly `import Data.Monoid (mempty)`, it fails with lts 6 and says "Perhaps you meant this: import Data.Monoid ()".

paul-rouse:
The compiler's suggestion is misleading in this case - you really do need `import Data.Monoid (mempty)` for lts-2.

In this project, all compiler warnings are switched on and are treated as errors, which is quite strict!  Of course the problem in your QQ.hs module is that the Prelude in base-4.7 didn't export the Monoid stuff, but from base-4.8 onwards it does, making the explicit import redundant.  Since we know what we are doing (:smile:) it is simplest to just silence the warning:

```
{-# OPTIONS_GHC -fno-warn-unused-imports #-}
...
import Data.Monoid (mappend)
```

felixSchl:
Does that suggest I could do `import Prelude hiding (mempty)` and later `import Data.Monoid (mempty)`. A bit awkward though without the context :)

paul-rouse:
I think `import Prelude hiding (mempty)` will complain in lts-2, when `mempty` is *not* exported by `Prelude`

felixSchl:
Ah yeah, of course. Hopefully this round of CI will pass. Did you have any response to the previous comments I made (interpolating table and column names)? I think that would be really cool to have, but I still haven't worked out how to do to the `fmap connEscapeName ask >>= \escape ->` dance inside TH.

paul-rouse:
I can certainly see the benefit of having a syntax to interpolate table and column names.  It is up to you - if you want to work that out now, then I'll wait for it; otherwise, if you prefer, I can merge this version and you can submit a separate PR when you are ready to extend it with the other types of substitution.  Of course that assumes that the extended version would not break anything you have already done here.

felixSchl:
I don't see how it would break anything as the types don't have to change and the syntax is only extended. But if you bear with me, I'd like to give it another go before hand off

paul-rouse:
OK, I'll wait for you to say!

felixSchl:
I've pretty much got it going, but I've met an issue cannot work out, any help would be really appreciated. I am trying to add tests that use the new quasi quoting feature to escape column names, but somehow it fails to deduce "PersistStoreWrite backend".

Here's the error message:

```
/home/felix/projects/persistent/persistent-test/src/PersistentTest.hs:1013:20: error:
    • Could not deduce (PersistStoreWrite backend0)
        arising from a use of ‘persistFieldDef’
      from the context: (RawSql a,
                         MonadIO m,
                         PersistStoreWrite backend,
                         backend ~ SqlBackend)
        bound by the type signature for:
                   runQuery :: (RawSql a, MonadIO m, PersistStoreWrite backend,
                                backend ~ SqlBackend) =>
                               Int64 -> ReaderT backend m [a]
        at src/PersistentTest.hs:(1006,11)-(1011,36)
      The type variable ‘backend0’ is ambiguous
      These potential instances exist:
        instance PersistStoreWrite SqlBackend
          -- Defined in ‘persistent-2.7.0:Database.Persist.Sql.Orphan.PersistStore’
        instance PersistStoreWrite SqlWriteBackend
          -- Defined in ‘persistent-2.7.0:Database.Persist.Sql.Orphan.PersistStore’
    • In the second argument of ‘($)’, namely
        ‘(persistFieldDef $ PersonAge)’
      In the first argument of ‘escape_aJV7’, namely
        ‘(fieldDB $ (persistFieldDef $ PersonAge))’
      In the first argument of ‘Control.Arrow.first
                                . ((++) . T.unpack)’, namely
        ‘escape_aJV7 (fieldDB $ (persistFieldDef $ PersonAge))’
```

paul-rouse:
It looks as if some extra constraint or functional dependency is needed, but I'd need to look harder.  I would isolate it in a minimal example, and if you still can't see what's going on, perhaps dump the template haskell expansion (-ddump-splices).

However, I am becoming a little concerned.  If it is hard for the user to get the types right, the API is probably wrong!  Should we, after all, separate out this extra table/column name interpolation into another PR?

felixSchl:
I went back on using `persistFieldDef` in favor of `getFieldName` and `getTableName` which means I did not have to grab the escape function, but which also means the query construction itself runs in `ReaderT SqlBackend m`. Even thought the TH code now evaluates to running in a monad, I think this version is by far the cleanest and most readable.

With respect to fundeps and error messages, after the change above I found it intuitive to work with downstream when fixing up the tests. For example, passing a wrong table type, it would say "Type not in scope 'Name'", if you pass anything but the expected type you would get "Couldn't match type ‘PersistEntityBackend "Type"’ with ‘SqlBackend’". For columns you get a "DataConstructor not in scope" compile error.

felixSchl:
@paul-rouse, Have you had any chance to look over this since? I've been using this successfully on a project of mine for a week now without problems.

paul-rouse:
Apologies - I saw the latest commit while the tests were still running, so I waited for them to complete:smile:  Sorry!

Could you please add an entry in the changelog (including a link to this PR).  You can still change the entry which has been started for version 2.7.2, because it is not yet released.  It looks as if you will need to rebase your branch to master so that we don't get a merge conflict in ChangeLog.md

Then we should be good to go.

felixSchl:
Not sure what's going on there but it seems the tests are failing for reasons other than the actual code changes, looks it's having trouble fetching some dependencies for the old lts2.2 https://travis-ci.org/yesodweb/persistent/jobs/299921263

paul-rouse:
I restarted that test, and it is OK now.

Sorry to be a pain - could you add a link to the PR in ChangeLog.md, please, to be consistent with other recent entries?

felixSchl:
Oh sorry, yeah will do

felixSchl:
Alright, updated and CI still OK :)

paul-rouse:
Great!  Thanks for all of your work on this!

PR #719, opened 2017-10-22
Cache .stack-work in Travis script

parsonsmatt:
I noticed that  significant amount of the time on Travis is spent building library dependencies. This line should allow `stack` to cache the `extra-deps` libraries.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ script: travis/run.sh
 cache:
   directories:
   - $HOME/.stack
+  - .stack-work
 
 
 # zookeeper doesn't seem to work any more


paul-rouse:
I worry slightly about adding to the amount kept between tests, but I think this is safe enough.  I'll merge it in a couple of days if nobody else has objected.

(The one failing test case is nothing to do with this change, and is addressed in #720)

parsonsmatt:
That's a good point to raise. I have this enabled on the work repo and it can do a really nice job of speeding up the builds, though we only need to work on one resolver and on one test suite :smile:

paul-rouse:
Merging since there are no further comments.  Thanks for the change!

PR #721, opened 2017-10-22
Document precision of time in MySQL

paul-rouse:
See #705


paul-rouse:
This is just an extra note in the documentation, so I am merging it.

PR #730, opened 2017-11-02
pass Connection into step, so it can be provided to sqlError

dlthomas-leapyear:
This gives us a more informative error message than "SQLite3 returned
ErrorError while attempting to perform step.", which isn't a very good
starting point for debugging.

I don't know whether there was a good reason it wasn't done this way originally - I dug a bit (across a few repositories) but wasn't able to find where that code originated.


paul-rouse:
This looks fine.  Could you please add a minor version bump and a changelog entry, please?

MaxGabriel:
@dlthomas-leapyear Just checking back in if you're able to bump the version number in the cabal file and the changelog, otherwise I was going to take this over. Thanks for your fix!

MaxGabriel:
I ended up handling the changelog/cabal bump in #743. Thanks @dlthomas-leapyear!

dlthomas-leapyear:
@MaxGabriel Thanks for getting this!  I've been away from the computer for a bit, with the holiday.

MaxGabriel:
👍

Issue #731, opened 2017-11-22
Identify mismatches between `sqltype` and Haskell type at compile time


parsonsmatt:
[This question was posted on Reddit](https://www.reddit.com/r/haskell/comments/7erfio/persistmarshalerror_when_using_text_haskell_value/). The brief summary was:

> The problem is that persistent doesn't know how to turn a value of type PersistDbSpecific to a value of type Text. The persistent wiki shows what values persistent can marshal between out of the box, and the PostgreSQL value of citext is not on that list.

It should be possible to move this into a compile-time error instead of a run-time error. That would improve ergonomics of the library -- "Failed to generate schema: SqlType and HaskellType are incompatible. Consider writing a newtype and an instance of PersistField and PersistFieldSql."

MaxGabriel:
As long as the runtime error is still around, it'd be nice if the error message was improved from `PersistMarshalError "field email: Cannot convert PersistDbSpecific to Text"` to mention the newtype/PersistField/PersistFieldSql approach.

Also apropos of that Reddit thread, I asked the OP to make an issue on persistent to support case insensitive text, but we should do that if he or she doesn't get around to it. I'm looking into supporting it for `persistent-simple`, but haven't developed on persistent-simple before so we'll see how that goes.

parsonsmatt:
If we do, we should rope in the `Data.CaseInsensitive` type, as it already captures this use nicely.

parsonsmatt:
I'm not certain that we can actually do this meaningfully. The entire point of the `sqltype=` annotation is to override `persistent`'s default idea, so in most cases we'd get a failure when we're trying to specialize.

I am however uncertain why it's doing `PersistDbSpecific` for that.

PR #733, opened 2017-11-23
Recommend the PersistDbSpecific docs if someone gets an error about them

MaxGabriel:
This change is in hopes of improving the experience/offering a quicker resolution for users like this one: https://www.reddit.com/r/haskell/comments/7erfio/persistmarshalerror_when_using_text_haskell_value/

These are the docs for PersistDbSpecific, which have a great example of using a custom type: https://www.stackage.org/haddock/lts-9.14/persistent-2.7.1/Database-Persist-Types.html#t:PersistValue

A longer term solution might be #731.

cc @parsonsmatt

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,3 +1,7 @@
+## 2.7.2.1
+
+* Recommend the `PersistDbSpecific` docs if someone gets an error about converting from `PersistDbSpecific`
+
 ## 2.7.2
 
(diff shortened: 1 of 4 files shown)


parsonsmatt:
Nice! Errors with pointers to solutions are great.

MaxGabriel:
@paul-rouse Does this one need to not be released to hackage because 2.7.2 has an accidental breaking change in it, and was thus deprecated?

PR #723, opened 2017-10-24
More BackendCompatible generalizations

parsonsmatt:
This PR includes some additional generalizations. Unfortunately, my work project does not use the `SqlReadT` or `SqlWriteT` features, and it was only when the code was tested on those did the issues with Esqueleto come up.

This PR generalizes `SqlReadT` so that Esqueleto can be updated in a way that doesn't cause any breaking changes.

If any folks are using SqlReadT, I'd appreciate testing this PR/commit out with [this Esqueleto PR](https://github.com/bitemyapp/esqueleto/pull/53) to verify that it doesn't break any corner cases I've missed :smile:

Issue #578, opened 2016-06-14
Persistent Entity Declaration doesn't work with `DeriveAnyClass` enabled


pseudonom:
``` haskell
{-# LANGUAGE DeriveAnyClass #-}
import ClassyPrelude.Yesod
import Database.Persist.Quasi

share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
Foo
  bar Text
|]
```

fails with 

```
No instance for (Generic (Key Foo))
          arising from a use of ?aeson-0.11.2.0:Data.Aeson.Types.Class.$gdmtoJSON?
        In the expression: aeson-0.11.2.0:Data.Aeson.Types.Class.$gdmtoJSON
        In an equation for ?toJSON?:
            toJSON = aeson-0.11.2.0:Data.Aeson.Types.Class.$gdmtoJSON
        When typechecking the code for  ?toJSON?
          in a derived instance for ?ToJSON (Key Foo)?:
          To see the code I am typechecking, use -ddump-deriv
        In the instance declaration for ?ToJSON (Key Foo)?
```

Changing the extension to `GeneralizedNewtypeDeriving` works. If both are enabled, `DeriveAnyClass` takes precedence as per page 277 (261 of the internal page numbering) of https://downloads.haskell.org/~ghc/latest/docs/users_guide.pdf and we get the error again. This was particularly sneaky as I had `DeriveAnyClass` enabled only via my cabal file.

Adding `Generic` to https://github.com/yesodweb/persistent/blob/master/persistent-template/Database/Persist/TH.hs#L710 would make the error go away but could be a little confusing in that the JSON instance now depends on which extensions are enabled.

gregwebs:
It seems like you shouldn't put DeriveAnyClass in a cabal file then. My rule is anything in there should only be an enabler of a feature that without the program would fail. I do make an exception to this sometimes for OverloadedStrings because it is such an ingrained part of Haskell now.

Hmm, so this works now

```
:set -XDeriveGeneric
:set -XDeriveAnyClass
import GHC.Generics
import Data.Aeson
data Coord2 = Coord2 { x2 :: Double, y2 :: Double } deriving (Generic, FromJSON)
```

I am not opposed however to adding the `Generic` instance and letting the user sort out how they want to derive things with their extensions.

Profpatsch:
I’m so glad this issue exists, I would have never found the reason why it didn’t work.

parsonsmatt:
[...]

PR #734, opened 2017-11-23
Add Generic instance and remove constraint on Entity

### LINKED ISSUE #694: Add Generic instance to Entity type
Hi.

I want to use persistent with servant and servant-purescript - package that automatically generates datatypes and utils for fetching them for the client. 

I want to send Entity instances from backend, since I need to have access to ids of sent entities, but it is required by the servant-purescript library that every datatype which will be sent from server to client is an instance of Generic typeclass.

Is it possible to add such typeclass instance to Entity? It would automate much of the work - 
currently I need to define my own Generic types and then map Entity instances returned by sql queries before sending them to client.

Thanks in advance.

parsonsmatt:
This PR removes the `PersistEntity` constraint on the `Entity` data constructor, which allows us to derive `Generic` for the type.

Fixes #694

### DIFF
--- a/persistent/Database/Persist/Class/PersistEntity.hs
+++ b/persistent/Database/Persist/Class/PersistEntity.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE OverloadedStrings #-}
@@ -43,6 +44,7 @@ import Data.Monoid (mappend)
 import qualified Data.HashMap.Strict as HM
 import Data.Typeable (Typeable)
 import Data.Maybe (isJust)
+import GHC.Generics
 
 -- | Persistent serialized Haskell records to the database.
 -- A Database 'Entity' (A row in SQL, a document in MongoDB, etc)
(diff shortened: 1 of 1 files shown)


MaxGabriel:
Related to this, I found that generic instances are not generated for `Key`s. If you add `''Generic` to [this list](https://github.com/yesodweb/persistent/blob/0b8c716e8910c49da434a6e84bd7cccd505bee24/persistent-template/Database/Persist/TH.hs#L724) they are generated. I'm thinking they might have been there originally and removed, because of the issue described in #578 (If you had `DeriveAnyClass` enabled, it would cause errors when deriving Generic).

Possibly Generic should be added back? Two reasons this could be more doable now or in the near future:

1. GHC 8.2 introduces [deriving strategies](https://ghc.haskell.org/trac/ghc/wiki/Commentary/Compiler/DerivingStrategies), so maybe the Template Haskell code could specify the `NEWTYPE` deriving strategy to work around this? But GHC 8.2 is pretty new.
2. Yesod's scaffolding encourages per-module language extensions now

MaxGabriel:
This is the git blame for that line: https://github.com/yesodweb/persistent/commit/05bd93974ef8427120675d343d2cff205abd1c81

@gregwebs can you comment on @parsonsmatt's change? Seemed unlikely the `PersistEntity record` constraint would be added for no reason, so wanted to make sure that it wasn't doing something helpful like improving type inference or ensuring correctness in some way.

parsonsmatt:
I'll test it out on my applications and `persistent-typed-db`. I'll see if I can get another few apps tested against this; I don't know how common it is for apps to be using that dictionary.

MaxGabriel:
@parsonsmatt Esqueleto also compiles with this change (after I reverted #723 which is currently breaking it)

snoyberg:
Merged to the newer-conduit branch, part of PR #773

PR #736, opened 2017-11-24
Sqlite: do not add foreign key constraints to temporary tables

### LINKED ISSUE #735: persistent-sqlite regression between 2.6.0.1 and 2.6.2
Here's program that works with `persistent-2.5` and `persistent-sqlite-2.6` but doesn't work with `persistent-2.7.1` and `persistent-sqlite-2.6.3`:

```haskell
{-# LANGUAGE ConstraintKinds            #-}
{-# LANGUAGE DeriveGeneric              #-}
{-# LANGUAGE EmptyDataDecls             #-}
{-# LANGUAGE FlexibleContexts           #-}
{-# LANGUAGE GADTs                      #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE LambdaCase                 #-}
{-# LANGUAGE MultiParamTypeClasses      #-}
{-# LANGUAGE OverloadedStrings          #-}
{-# LANGUAGE QuasiQuotes                #-}
{-# LANGUAGE ScopedTypeVariables        #-}
{-# LANGUAGE TemplateHaskell            #-}
{-# LANGUAGE TupleSections              #-}
{-# LANGUAGE TypeFamilies               #-}

module Main where

import Control.Monad
import Control.Monad.Logger
import Control.Monad.Reader
import Control.Monad.Trans.Resource
import Data.Pool
import qualified Data.Text as T
import Database.Persist
import Database.Persist.Sqlite hiding (SqlBackend)
import qualified Database.Persist.Sqlite as Sqlite
import Database.Persist.TH
import Database.Sqlite (Error (..), SqliteException (..))

share [mkPersist sqlSettings, mkMigrate "migrateDb"] [persistLowerCase|

User
  username T.Text

Message
  user UserId

|]

-- On the second run move this line to `Message` table above:
-- new_field T.Text Maybe default="null"

main :: IO ()
main = runStdoutLoggingT $ do
    conn_pool <- createSqlitePool "db" 15
    withResource conn_pool $ runReaderT $ do
      void (runMigrationSilent migrateDb)
```

This is a migration program so we run it twice. On the second run we move the commented-out line to `Message` table as mentioned in the code.

On older persistent migration works as expected:

```
[Debug#SQL] CREATE TEMP TABLE "message_backup"("id" INTEGER PRIMARY KEY,"user" INTEGER NOT NULL REFERENCES "user","new_field" VARCHAR NULL DEFAULT null); []
[Debug#SQL] INSERT INTO "message_backup"("id","user") SELECT "id","user" FROM "message"; []
[Debug#SQL] DROP TABLE "message"; []
[Debug#SQL] CREATE TABLE "message"("id" INTEGER PRIMARY KEY,"user" INTEGER NOT NULL REFERENCES "user","new_field" VARCHAR NULL
DEFAULT null); []
[Debug#SQL] INSERT INTO "message" SELECT "id","user","new_field" FROM "message_backup"; []
[Debug#SQL] DROP TABLE "message_backup"; []
```

On newer persistent it doesn't:

```
[Debug#SQL] CREATE TEMP TABLE "message_backup"("id" INTEGER PRIMARY KEY,"user" INTEGER NOT NULL REFERENCES "user","new_field" VARCHAR NULL DEFAULT null); []
[Debug#SQL] INSERT INTO "message_backup"("id","user") SELECT "id","user" FROM "message"; []
bug: SQLite3 returned ErrorError while attempting to perform prepare "INSERT INTO \"message_backup\"(\"id\",\"user\") SELECT \"id\",\"user\" FROM \"message\"": no such table: temp.user
```

cabal and stack files:

```
name:                persistent-bug
version:             0.1.0.0
-- synopsis:
-- description:
homepage:            https://github.com/githubuser/persistent-bug#readme
license:             BSD3
license-file:        LICENSE
author:              Author name here
maintainer:          example@example.com
copyright:           2017 Author name here
category:            Web
build-type:          Simple
extra-source-files:  README.md
cabal-version:       >=1.10

executable bug
  hs-source-dirs:      app
  main-is:             Main.hs
  ghc-options:         -threaded -rtsopts -with-rtsopts=-N

  build-depends:
    base,
    monad-logger,
    mtl,
    persistent,
    persistent-sqlite,
    persistent-template,
    resourcet,
    resource-pool,
    text

  default-language:    Haskell2010
```

```
# buggy
# resolver: lts-9.14
# resolver: nightly-2017-11-24
# works
resolver: lts-7.14
packages:
- .
```

paul-rouse:
Fixes #735 

@MaxGabriel,  @psibi would one of you mind checking my logic, please?  The issue is that migration can involve creating a temporary table, and we were adding foreign key constraints that existed in the table being copied.  Sqlite doesn't handle FK constraints from temporary tables to non-temporary ones - it probably doesn't make sense to do so - and when checking is switched on, it reports them as violated.

The fix is simply to omit the FK constraint from the temporary table!

### DIFF
--- a/persistent-sqlite/Database/Persist/Sqlite.hs
+++ b/persistent-sqlite/Database/Persist/Sqlite.hs
@@ -429,7 +429,7 @@ mkCreateTable isTemp entity (cols, uniqs) =
         , " TABLE "
         , escape $ entityDB entity
         , "("
-        , T.drop 1 $ T.concat $ map sqlColumn cols
+        , T.drop 1 $ T.concat $ map (sqlColumn isTemp) cols
         , ", PRIMARY KEY "
         , "("
         , T.intercalate "," $ map (escape . fieldDB) $ compositeFields pdef
@@ -447,7 +447,7 @@ mkCreateTable isTemp entity (cols, uniqs) =
         , showSqlType $ fieldSqlType $ entityId entity
         ," PRIMARY KEY"
         , mayDefault $ defaultAttribute $ fieldAttrs $ entityId entity
-        , T.concat $ map sqlColumn cols
+        , T.concat $ map (sqlColumn isTemp) cols
         , T.concat $ map sqlUnique uniqs
         , ")"
         ]
@@ -457,8 +457,8 @@ mayDefault def = case def of
     Nothing -> ""
     Just d -> " DEFAULT " <> d
 
-sqlColumn :: Column -> Text
-sqlColumn (Column name isNull typ def _cn _maxLen ref) = T.concat
+sqlColumn :: Bool -> Column -> Text
+sqlColumn noRef (Column name isNull typ def _cn _maxLen ref) = T.concat
     [ ","
     , escape name
     , " "
@@ -467,7 +467,7 @@ sqlColumn (Column name isNull typ def _cn _maxLen ref) = T.concat
     , mayDefault def
     , case ref of
         Nothing -> ""
-        Just (table, _) -> " REFERENCES " <> escape table
+        Just (table, _) -> if noRef then "" else " REFERENCES " <> escape table
     ]
 
 sqlUnique :: UniqueDef -> Text
--- a/persistent-test/persistent-test.cabal
+++ b/persistent-test/persistent-test.cabal
@@ -187,6 +187,8 @@ executable persistent-test
                  , resourcet
                  , scientific
 
+  if !flag(postgresql) && !flag(mysql) && !flag(mongodb) && !flag(zookeeper)
+      cpp-options: -DWITH_SQLITE -DDEBUG
   if flag(zookeeper)
       cpp-options: -DWITH_NOSQL -DWITH_ZOOKEEPER -DDEBUG
   if flag(mongodb)
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,3 +1,7 @@
+## 2.6.3.1
+
+* Fix migration to avoid creating foreign-key constraints in temporary tables [#736](https://github.com/yesodweb/persistent/pull/736)
+
 ## 2.6.3
 
 * Add 'use-pkgconfig' flag to use pkg-config to find system SQLite library.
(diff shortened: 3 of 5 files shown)


persistent-test/persistent-test.cabal
```diff
@@ -187,6 +187,8 @@ executable persistent-test
                  , resourcet
                  , scientific
 
+  if !flag(postgresql) && !flag(mysql) && !flag(mongodb) && !flag(zookeeper)
+      cpp-options: -DWITH_SQLITE -DDEBUG
```
MaxGabriel: This seems a little weird—why not have a flag for SQLite?

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -467,7 +467,7 @@ sqlColumn (Column name isNull typ def _cn _maxLen ref) = T.concat
     , mayDefault def
     , case ref of
         Nothing -> ""
-        Just (table, _) -> " REFERENCES " <> escape table
+        Just (table, _) -> if noRef then "" else " REFERENCES " <> escape table
```
MaxGabriel: So the other option would be to set `ref` on each `Column` to `Nothing`. I kind of like that because it reduces the amount of branching. But I could also see an argument that it's confusing to modify the data of `Column`, which actually hasn't changed, and the `Bool` parameter is better because it shows that the data is being overridden. Thoughts?

persistent-test/persistent-test.cabal
paul-rouse: It's just doing for test/main.hs what was already being done for the main test modules in the library

persistent-sqlite/Database/Persist/Sqlite.hs
paul-rouse: We are in the function which is being mapped over the columns, so it seemed easiest to do the work here.  Does that sound reasonable?

persistent-test/persistent-test.cabal
MaxGabriel: Oh right, it's this way so that SQLite is the default. 👍

persistent-sqlite/Database/Persist/Sqlite.hs
MaxGabriel: Ok cool. PR LGTM!

paul-rouse: Thanks @MaxGabriel !

paul-rouse:
Thanks @psibi !

PR #742, opened 2017-11-26
[persistent-postgresql] Recognize column-widths

MaxGabriel:
This is #686 with a test added to confirm it works (I also tested against a real Yesod project). 

This closes #686

Btw: Thanks @JaSpa! Sorry your PR never got looked at before.


MaxGabriel:
Released to hackage

PR #748, opened 2017-11-27
Allow Postgres migrations to enable extensions

enolan:
This is kind of ugly, I'm happy to take feedback.

### DIFF
--- a/persistent-postgresql/ChangeLog.md
+++ b/persistent-postgresql/ChangeLog.md
@@ -1,3 +1,7 @@
+## 2.6.3
+
+* Added new function `migrateEnableExtension`, to enable Postgres extensions in migrations.
+
 ## 2.6.2.1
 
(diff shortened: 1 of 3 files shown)


MaxGabriel:
@enolan Why not just run `CREATE EXTENSION IF NOT EXISTS` instead of checking, then running? I think that would simplify the code greatly.

enolan:
That's what I had originally, but it means that the migration always prints when you start your program, which is a little annoying and a little confusing.

MaxGabriel:
@enolan I think this is a good PR then. Can you bump the version number of the cabal file and add a changelog entry?

enolan:
I bumped the version number and added a changelog entry. Something weird is happening with Travis, but I'm pretty sure it has nothing to do with my code.

MaxGabriel:
Merged! Thanks for your contribution :)

PR #751, opened 2017-12-03
Export SQLite connection info record functions

MaxGabriel:
* These were previously only exported as lenses

I personally think it's weird for persistent-sqlite to export lenses for this at all, but I also just don't use lens so I'm not appreciative of its benefits.

### DIFF
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,3 +1,8 @@
+## 2.6.5
+
+* Exports the record update functions for `SqliteConnectionInfo`
+	* These were previously only available as lenses.
+
 ## 2.6.4
 
 * Adds a new function `stepConn`, which uses an additional parameter to give more detailed error messages [#750](https://github.com/yesodweb/persistent/pull/750)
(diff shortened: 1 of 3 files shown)


paul-rouse:
There was a discussion of this in #646, and particularly [Michael's comment](https://github.com/yesodweb/persistent/pull/646#issuecomment-270314449) pointing to his blog article about [designing APIs for extensibility](http://www.snoyman.com/blog/2016/11/designing-apis-for-extensibility).

That discussion seems to give a good argument for avoiding exposing the actual field accessors, but I have no strong feelings about whether the getters and setters should be implemented using the microlens library, or just written out by hand.

MaxGabriel:
Hm, I wasn’t aware of this new approach, and would be fine writing getters and setters by hand.

My general complaint is that I don’t think end users should have to learn any lens library to configure SQLite.

paul-rouse:
Yes, I can see the point.  It sounds like another piece of magic to learn, even though actually you just say things like `set fkEnabled False theConnectionInfo`.

MaxGabriel:
I do think it’s much improved by providing example code of how to use the lenses. Re-implementing the lenses by hand would also improve things, because then they could have Haddock documentation generated. Maybe that’s sufficient.

parsonsmatt:
> Hm, I wasn’t aware of this new approach, and would be fine writing getters and setters by hand.
> 
> My general complaint is that I don’t think end users should have to learn any lens library to configure SQLite.

Agreed - I think I'd be in favor of exporting lenses *and* getters/setters. This gives you simplicity in the `getX/setX` functions, composition with the lenses, and ability to change the configuration type without a major version bump.

PR #753, opened 2017-12-04
fix the docker image FROM

### LINKED ISSUE #747: Dockerfile doesn't work, "haskell-stack:7.8" doesn't exist
The Dockerfile has the line "FROM haskell-stack:7.8", but there's no such image, at least not in the public repository under Docker's official images.

```
$ docker build -t persistent .
Sending build context to Docker daemon 4.096 kB
Step 1 : FROM haskell-stack:7.8
Pulling repository docker.io/library/haskell-stack
Error: image library/haskell-stack:7.8 not found
```

gregwebs:
fixes #747


enolan:
Hey Greg. I used the modified Dockerfile and got it working, although it's a bit annoying to use. The right version of GHC wasn't installed, so I ran `stack setup`. That failed, saying that `xz` and `make` were missing. I went and installed them - Debian packages `xz-utils` and `make`. Then, when testing the `test` database didn't exist, so I had to create it.

I don't know how much effort you want to go to to make it easy to test persistent, this change does at least improve the situation.

gregwebs:
okay, thanks for the feedback. I think its time to completely abandon this approach in favor of an external docker-compose approach.

gregwebs:
I added another commit with a test runner. Is this better for you? Do you want to contribute a backend? I think we would probably still want to move docker image configuration to a docker-compose file.

enolan:
This is a great idea, but the `docker logs -f persistent_postgres_run_1` command never exits, so the test never actually runs.

gregwebs:
Thanks, removed that. Its working now. Obviously the sleeps here are horrible, I would like to figure out a standard technique for waiting for container startup.

PR #754, opened 2017-12-05
Fix mysql sqltype migrations [#752]

### LINKED ISSUE #752: [mysql] Specifying sqltype=double(6,5) or sqltype=mediumtext causes repeated migrations
It looks like choosing a custom sqltype for a Text or a Double causes a repeated, unnecessary migration every time runMigration is called.

Example
```
Page
    ...
    content T.Text Maybe sqltype=mediumtext
    score Double sqltype=double(6,5)
    ...
```
This causes 2 alter table statements to be executed every time I run my app:
```
Migrating: ALTER TABLE `page` CHANGE `content` `content` mediumtext NULL
Migrating: ALTER TABLE `page` CHANGE `score` `score` double(6,5) NOT NULL
```

I'm using persistent 2.7.1 and persistent-mysql 2.6.2.

dgaw:
A fix and a test for MySQL migrations containing certain backend-specific sqltypes (i.e. `mediumtext`, `longtext`, `mediumblob`, `longblob` and custom precision `double`).

Closes #752

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -627,13 +627,9 @@ parseColumnType "decimal" ci                                   =
 -- Text
 parseColumnType "varchar" ci                                   = return (SqlString, ciMaxLength ci)
 parseColumnType "text" _                                       = return (SqlString, Nothing)
-parseColumnType "mediumtext" _                                 = return (SqlString, Nothing)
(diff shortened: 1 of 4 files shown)


paul-rouse:
The test failure looks random - I've just restarted that one case.

dgaw:
Thanks @paul-rouse. I think it was a network timeout or something.

PR #758, opened 2017-12-16
Re-release Persistent 2.7.2 as 2.8

### LINKED ISSUE #740: Thoughts on releases, please
We have a breaking change due to #723, so I have had to deprecate the release of `persistent` I did earlier - we accidentally numbered it as a minor version change.

Any thoughts on whether we want to do a major release now, or queue up anything else to go into it? 

cc @parsonsmatt, @MaxGabriel, @psibi

MaxGabriel:
Rolls 2.7.2.1 into 2.8, since it was never released (I don't want to release it separately then deprecate 2.7.2.1 on hackage)

Closes #740

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,9 +1,11 @@
-## 2.7.2.1
+## 2.8
 
+* Re-releases 2.7.2 as a major version bump
 * Recommend the `PersistDbSpecific` docs if someone gets an error about converting from `PersistDbSpecific`
 
-## 2.7.2
+## 2.7.2 [DEPRECATED]
 
+* This version accidentally introduced a breaking change and is deprecated.
 * Many of the functions have been generalized using the `BackendCompatible` class. [#723](https://github.com/yesodweb/persistent/pull/723)
 * Add raw sql quasi quoters [#717](https://github.com/yesodweb/persistent/pull/717)
 
--- a/persistent/persistent.cabal
+++ b/persistent/persistent.cabal
@@ -1,5 +1,5 @@
 name:            persistent
-version:         2.7.2.1
+version:         2.8
 license:         MIT
 license-file:    LICENSE
 author:          Michael Snoyman <michael@snoyman.com>


MaxGabriel:
@parsonsmatt and I were considering adding in #734 to this just in case it's somehow a breaking change

MaxGabriel:
Hm, I realized that #723 also broke persistent-mysql. persistent-mysql's cabal file currently has:

```
, persistent            >= 2.6.1    && < 3
```

So that implies that we'd need to release this as persistent-3, otherwise old versions of persistent-mysql will continue to try to build against persistent-mysql with a breaking change. Alternatively, we could change a bunch of the version constraints on Hackage, but I hear that's a bad practice.

MaxGabriel:
I talked with @snoyberg about this today and we concluded there are three options, all of which are acceptable:

1. Release as persistent 3
2. Release as 2.8 and use hackage revisions to fix older versions of persistent-mysql
3. Just release as 2.8

I would probably default to releasing as persistent 3, which will require bumping the rest of persistent postgres/sqlite/etc. to say that they work with it. But I'm flexible—chime in if you have a preference here.

He also recommended releasing a 2.7.3 that reverted the breaking change, so that 2.x/2.7.x users can upgrade to something workable.

cc @paul-rouse @parsonsmatt

parsonsmatt:
I'd be fine with hackage revisions, to be honest -- this seems like exactly the use case for them.

paul-rouse:
My own preference would be for 3.0.  I dislike hackage revisions except, perhaps, to *increase* an upper bound on a dependency.  Also going for 3.0 seems to me to make it clearer that there really is a breakage.

That's a good point about doing a 2.7.3 release without #723, too.

parsonsmatt:
Actually, there's a change I want to make to the `persistent-mysql` duplicate-key-update query update syntax that's breaking. I'll make a patch real quick

MaxGabriel:
Ok so let's release 2.7.3 first: https://github.com/yesodweb/persistent/pull/759

After 2.7.3 is released I would like to merge in two PRs that I've been sitting on (couldn't release them because `master` was in this broken state).

https://github.com/yesodweb/persistent/pull/757
https://github.com/yesodweb/persistent/pull/741

parsonsmatt:
I'll toss #760 as something I'd like merged in for the 3.0, but that's a `persistent-mysql` change only.

MaxGabriel:
Ok 2.7.3 is released on Hackage. My plan is to:

1. Let it percolate to a stackage release to confirm everything builds again
2. Release some stuff on top of 2.7.3 (PRs sitting around to be merged)
3. Merge this PR, either with the 2.8 w/ revisions or 3.0 route (I'm agnostic).

parsonsmatt:
This one seems stale; gonna close. feel free to reopen if it's still useful to have up

PR #760, opened 2017-12-17
[persistent-mysql] Rename SomeField type to HandleUpdateCollision

parsonsmatt:
I want to get this in before 3.0 :smile: 

`SomeField` worked as a name before the responsibility of the type grew. Renaming it makes it more clear what it does and what it is for. Hiding the constructors allows us to add additional combinators without them being breaking changes -- I want to add `copyUnlessIn`, `modifyUsing` and possibly others at some point.

I'm not 100% on the new type alias and pattern constructor for `SomeField`. They'll prevent it from breaking with a compile error and instead issue a deprecation notice for migration, which is helpful, but also extends the amount of backwards compatibility cruft we have to carry around.

### DIFF
--- a/persistent-mysql/ChangeLog.md
+++ b/persistent-mysql/ChangeLog.md
@@ -4,6 +4,9 @@
 
 -- This can be released as a minor change on the next update. Currently persistent-mysql can't be released because 2.6.2.2 depends on persistent-2.7.2 being released.
 
+* The `SomeField` type was renamed to `HandleUpdateCollision` and deprecated. Please migrate to using `HandleUpdateCollision`.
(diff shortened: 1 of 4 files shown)


parsonsmatt:
OK, so the pattern synonym stuff is breaking with earlier GHCs (no type signatures on 7.8, can't export related pattern synonym on 7.10). I can write CPP to fix that, but that'll require some multiline string literals to be rewritten.

MaxGabriel:
I don't think it's required that we have a breaking API change on persistent-mysql, but the name `HandleUpdateCollision`/`CopyField` naming is a hundred times more clear.

snoyberg:
Merged to the newer-conduit branch, part of PR #773

PR #762, opened 2017-12-19
Don't migrate between text and varchar in Postgres / domains

MaxGabriel:
* These types are synonyms, with the exact same memory representation in Postgres
* So there's no need to suggest migrating from one to the other https://stackoverflow.com/a/4849030/1176156

My use case is that I have a database not using Persistent for migrations, and I want to use persistent to interact with it. I would also like to check that Persistent's models match the database schema by running `getMigration` and ensuring there are no changes to be made, but if there are spurious migrations then I'll have false positives.

In this database, `text` is being used instead of `varchar`, so Persistent claims it needs migrations. Since `text` and `varchar` are synonyms, this should be unnecessary.

**Edit:**

I found another, similar issue where Persistent would try to migrate a domained type to itself. This needed all the same test infrastructure and has the same theme, so I added it here. This is documented extensively in the comments; see there for details.

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -619,7 +619,7 @@ mayDefault def = case def of
 
 type SafeToRemove = Bool
 
-data AlterColumn = Type SqlType Text
+data AlterColumn = ChangeType SqlType Text
                  | IsNull | NotNull | Add' Column | Drop SafeToRemove
                  | Default Text | NoDefault | Update' Text
                  | AddReference DBName [DBName] [Text] | DropReference DBName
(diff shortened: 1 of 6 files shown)


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -619,7 +619,7 @@ mayDefault def = case def of
 
 type SafeToRemove = Bool
 
-data AlterColumn = Type SqlType Text
+data AlterColumn = ChangeType SqlType Text
```
MaxGabriel: In the course of debugging this issue I found it extremely hard to search for `Type`, so I renamed this to `ChangeType`.

PR #765, opened 2017-12-24
Introduce new module for RDBMS backend

psibi:
Also adds a function named `deleteAllRows`.

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -2,7 +2,7 @@
 
 * Introduce new module named `Database.Persist.RDBMS.Utils` which will contain utility function which works only on the RDBMS backends.
 
-## 2.7.3 
+## 2.7.3
 
 * Reverts [#723](https://github.com/yesodweb/persistent/pull/723), which generalized functions using the `BackendCompatible` class. These changes were an accidental breaking change.
 * Recommend the `PersistDbSpecific` docs if someone gets an error about converting from `PersistDbSpecific`
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,3 +1,7 @@
+## Version HEAD
+
+* Introduce new module named `Database.Persist.RDBMS.Utils` which will contain utility function which works only on the RDBMS backends.
+
 ## 2.7.3 
 
 * Reverts [#723](https://github.com/yesodweb/persistent/pull/723), which generalized functions using the `BackendCompatible` class. These changes were an accidental breaking change.
--- a/persistent/Database/Persist/RDBMS/Utils.hs
+++ b/persistent/Database/Persist/RDBMS/Utils.hs
@@ -27,30 +27,26 @@
 -- > +-----+-----+-----+
 --
 module Database.Persist.RDBMS.Utils
-  ( dropTable
+  ( deleteTable
   ) where
 
 import Control.Monad.IO.Class (MonadIO)
 import Control.Monad.Trans.Reader
 import Database.Persist
 
--- | Drops an existing table in a database. Issues \"DELETE FROM table_name\" command to the database.
+-- | Removes all the rows in the existing table. Issues \"DELETE FROM table_name\" command to the database.
 --
 -- @
 -- deleteUserRows :: MonadIO m => ReaderT SqlBackend m ()
--- deleteUserRows = dropTable (undefined :: User)
+-- deleteUserRows = deleteTable (undefined :: User)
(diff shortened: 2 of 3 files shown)


MaxGabriel:
Could this be implemented in terms of deleteWhere?

MaxGabriel:
Just checking, is the purpose of this to be used to clear the database during testing or is this for general use?

psibi:
@MaxGabriel Thanks for the review and comments.

> Just checking, is the purpose of this to be used to clear the database during testing or is this for general use?

I actually needed that for some benchmarking related code. :) Also, I have used `persistent` for making cli tool for interfacing with dbs and I had previously implemented the related functionality there itself. Also, I will slowly introduce other functions like `truncate`, `drop` etc.

> Could this be implemented in terms of deleteWhere?

I don't think that's possible: What will be the filter value you will be passing on ? (Even if we have some filter value for a primary key >= 0, it won't work in conditions when we have changed the type of a primary key. I think this was the most simplest way to do. Will be happy to hear your thoughts on this)

MaxGabriel:
If you pass no filters you'll just get a delete on all rows:

```
deleteWhere ([] :: [Filter User])
```

```
28/Dec/2017:18:00:53 -0500 [Debug#SQL] DELETE FROM "users"; []
```

> Also, I will slowly introduce other functions like truncate, drop etc.

These sound useful though

MaxGabriel:
@psibi Is this ready to merge or is this an in progress PR?

psibi:
@MaxGabriel I renamed the function to `deleteTable` now as that will easily indicate to the user which query it is generating underneath. This is ready for merging now.

MaxGabriel:
1. So now that this is implemented in terms of deleteWhere, it’s not specific to RDBMS systems right? It could go with the rest of the code. The module is still a good idea for e.g. a dropTable function that does scheme change
2. Should we use a fancy Proxy or the new TypeApplications extension instead of passing a dummy parameter?
3. I think `deleteTable` is kind of a confusing name because it sounds like it could be actually dropping the table. What about deleteAllRecords or deleteAll or deleteAllRows or something?

psibi:
> So now that this is implemented in terms of deleteWhere, it’s not specific to RDBMS systems right? It could go with the rest of the code. The module is still a good idea for e.g. a dropTable function that does scheme change

I agree. I think it can be moved under `Database.Persist.Class`.

> Should we use a fancy Proxy or the new TypeApplications extension instead of passing a dummy parameter?
I think any of them would be better than the current `undefined` value being passed. :-) What do you think would be the better option ?

>I think deleteTable is kind of a confusing name because it sounds like it could be actually dropping the table. What about deleteAllRecords or deleteAll or deleteAllRows or something?

Initially, it was `deleteAllRows` and I thought that was confusing and so changed it!. The reason I made it `deleteTable` is because it issues `delete * from table` command. But yeah, I can think having the former name would make the functionality clear.

MaxGabriel:
I haven’t used TypeApplications before, but it seems like the better option over Proxy. If I understand correctly only the client needs to use it, so persistent itself wouldn’t require GHC 8

MaxGabriel:
(But I’m not positive)

persistent/ChangeLog.md
```diff
@@ -1,4 +1,8 @@
-## 2.7.3 
+## Version HEAD
+
+* Introduce new module named `Database.Persist.RDBMS.Utils` which will contain utility function which works only on the RDBMS backends.
```
parsonsmatt: Isn't that `Database.Persist.Sql`?

persistent/Database/Persist/RDBMS/Utils.hs
```diff
@@ -0,0 +1,56 @@
...
+-- deleteUserRows :: MonadIO m => ReaderT SqlBackend m ()
+-- deleteUserRows = deleteTable (undefined :: User)
+-- @
+deleteTable
+  :: (PersistEntityBackend record ~ BaseBackend backend
+     ,PersistEntity record
+     ,PersistQueryWrite backend
+     ,MonadIO m)
+  => record -> ReaderT backend m ()
+deleteTable entity = deleteWhere filts
```
parsonsmatt: This can be `deleteWhere ([] :: [Filter record])` if you enable `ScopedTypeVariables` and do `forall record backend m.` before the constraint :smile:

parsonsmatt:
I'm going to close this PR due to inactivity. Feel free to reopen! :smile:

PR #781, opened 2018-02-08
Postgres: Improve error message when getting invalid scale/precision …

MaxGabriel:
…for numeric values

I ran into this error recently when not specifying precision for a numeric column, and the error message was pretty bad (it didn't tell you the table or column name that was the problem)

I also cleaned up the variable names. They were named things like `x` or `y` which made it pretty hard to tell what was what.

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -752,21 +752,21 @@ getAlters defs def (c1, u1) (c2, u2) =
 getColumn :: (Text -> IO Statement)
           -> DBName -> [PersistValue]
           -> IO (Either Text Column)
-getColumn getter tname [PersistText x, PersistText y, PersistText z, d, npre, nscl, maxlen] =
+getColumn getter tableName' [PersistText columnName, PersistText isNullable, PersistText typeName, defaultValue, numericPrecision, numericScale, maxlen] =
     case d' of
         Left s -> return $ Left s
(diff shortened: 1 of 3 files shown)


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -818,13 +818,34 @@ getColumn getter tname [PersistText x, PersistText y, PersistText z, d, npre, ns
...
-    return $ Left $ T.pack $ "Invalid result from information_schema: " ++ show x
+    getNumeric PersistNull PersistNull = Left $ T.concat
+      [ "No precision and scale were specified for the column: "
+      , columnName
+      , " in table: "
+      , unDBName tableName'
+      , ". Postgres defaults to a maximum scale of 147,455 and precision of 16383,"
+      , " which is probably not what you intended."
+      , " Specify the values as numeric(total_digits, digits_after_decimal_place)."
+      ]
```
MaxGabriel: An alternative to this error message would be to try to use a numeric value with extremely high scale and precision. It isn't possible to match the Postgres maximums, because the unspecified value has a precision of 16383, while [the Postgres docs](https://www.postgresql.org/docs/10/static/datatype-numeric.html) say:

> The maximum allowed precision when explicitly specified in the type declaration is 1000

See table 8.2 for these limits: https://www.postgresql.org/docs/10/static/datatype-numeric.html

PR #789, opened 2018-02-24
Sqlite upgrade to 3.22.0

joncfoo:

After submitting your PR:

### DIFF
--- a/development.md
+++ b/development.md
@@ -52,14 +52,3 @@ Lets develop easy entry points for testing a database using a command runner.
 Or you can look in `./bin/` for the commands it will call and run them directly
 Or
 Lets develop a docker-compose file for running different databases.
-
-
-# Updating Sqlite C Source in the repo
-
-Usually some patches are applied on top of the official sqlite C
-source. This section tries to document the patch to make future
-updates easier:
-
-* [Workaround for stat64 bug](https://github.com/yesodweb/persistent/commit/0df11d70e936389ca6ed15afb227f6224ad16f22)
-
-* [Workaround for stat64 bug on OSX](https://github.com/yesodweb/persistent/commit/d5836ed5eb76ce6a340442bdfc231895a23c29dc)
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,3 +1,7 @@
+## 2.8.1.1
+
+* Update `sqlite` cbit sources to 3.22.0 from 3.19.3
+
 ## 2.8.1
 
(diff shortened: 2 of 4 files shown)


joncfoo:
I also have another branch based on this PR that I'd like to submit though I'm not sure what the best way to do that is - https://github.com/joncfoo/persistent/tree/sqlite-enable-common-extensions

Thoughts?

MaxGabriel:
You can submit the second PR. I’ll merge the first and then we can merge the second (you may have to rebase the second against master after we merge the first, not sure if there’ll be merge conflicts).

These are great PRs btw! Thanks for contributing

MaxGabriel:
I’m on a ski trip this weekend and will probably be able to merge this on Monday

joncfoo:
Happy to contribute as I need this functionality as well :smile:

MaxGabriel:
@joncfoo I noticed [your commit that applies the stat64 workaround](https://github.com/yesodweb/persistent/pull/789/commits/2027b575cf206451b44713ef3d7574c5b73f3361) doesn't actually change any SQLite code to call those new functions, like [Michael's commit did](https://github.com/yesodweb/persistent/commit/0df11d70e936389ca6ed15afb227f6224ad16f22). Is that intentional?

MaxGabriel:
Hm, I notice that the [GHC ticket](https://ghc.haskell.org/trac/ghc/ticket/7072) the workaround was meant for is now closed, so potentially we can just not have this workaround at all.

joncfoo:
Oh, I did not apply the call-site patch!

Since the GHC ticket is closed regarding the stat64 workaround should I revert the extraneous addition to cbits?  Is there a good way to test the workaround is not needed (IIRC this issue affected macOS as well but I don't have a macOS machine)?

MaxGabriel:
@joncfoo I'll test macOS

MaxGabriel:
@joncfoo Ok, I was able to load persistent-sqlite in a repl from your branch (`stack repl persistent-sqlite`) and also wasn't able to reproduce with the reproduction code in the GHC ticket (`cabal build`, `cabal repl lib:stat-test`). And it seems like Linux is working because the GHC ticket is closed.

So here's what I propose: You delete your stat64 workaround code, and also delete the part of `development.md` that says to make this change.

Also: big thanks for helping remove this legacy cruft!

joncfoo:
I've updated the code.  I'm happy to contribute!  Thanks for looking into the stat64 workaround.  I'm glad we no longer need to do this.

MaxGabriel:
@joncfoo Released 👍

PR #791, opened 2018-02-25
Fix #778 | persistent-template - Add `persistManyFileWith`

### LINKED ISSUE #778: [persistent-template] schema organization and compilation times
This isn't so much a bug or support request, but something I discovered that could enhance developer experience. See sample project/findings here [naushadh/persistent-examples](https://github.com/naushadh/persistent-examples/tree/master/model-organization). TLDR: breaking entity definitions into smaller components can yield immense compilation time gains.

Is there any downside to the approach? Should `persistManyFileWith` be added as an official `persistent-template` API? Should guides suggest this approach (for compile time gains)?

Possibly related things:
- [Long compile time/high memory usage for modules with many deriving clauses](https://ghc.haskell.org/trac/ghc/ticket/9669)
- [ghc using 6+ GB of memory for module with Persistent models and Yesod handlers](https://groups.google.com/forum/#!msg/yesodweb/MX1bnOFu8Hc/VLQXhnvpIkMJ)
- [Is it possible to construct a `persistent` schema in pieces?](https://stackoverflow.com/questions/39800423/is-it-possible-to-construct-a-persistent-schema-in-pieces)

naushadh:
Address [persistent#778](https://github.com/yesodweb/persistent/issues/778).

After submitting your PR:

    - Linked to Issue instead since it has reference to the why and usage sample for new API
- [X] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

### DIFF
--- a/persistent-template/Database/Persist/TH.hs
+++ b/persistent-template/Database/Persist/TH.hs
@@ -22,6 +22,7 @@ module Database.Persist.TH
     , persistUpperCase
     , persistLowerCase
     , persistFileWith
+    , persistManyFileWith
       -- * Turn @EntityDef@s into types
     , mkPersist
     , MkPersistSettings
@@ -111,14 +112,29 @@ persistLowerCase = persistWith lowerCaseSettings
 -- | Same as 'persistWith', but uses an external file instead of a
 -- quasiquotation.
 persistFileWith :: PersistSettings -> FilePath -> Q Exp
-persistFileWith ps fp = do
+persistFileWith ps fp = persistManyFileWith ps [fp]
+
+-- | Same as 'persistWith', but uses several external files instead of a
+-- quasiquotation.
+--
+-- See <https://github.com/yesodweb/persistent/issues/778 persistent#778> for
+-- more info on how this can be exploited for improving (re)compilation times.
+--
+-- @since 2.5.4
+persistManyFileWith :: PersistSettings -> [FilePath] -> Q Exp
+persistManyFileWith ps fps = do
 #ifdef GHC_7_4
-    qAddDependentFile fp
+    mapM_ qAddDependentFile fps
 #endif
-    h <- qRunIO $ SIO.openFile fp SIO.ReadMode
-    qRunIO $ SIO.hSetEncoding h SIO.utf8_bom
-    s <- qRunIO $ TIO.hGetContents h
+    ss <- mapM getS fps
+    let s = mconcat ss
     parseReferences ps s
+  where
+    getS fp = do
+      h <- qRunIO $ SIO.openFile fp SIO.ReadMode
+      qRunIO $ SIO.hSetEncoding h SIO.utf8_bom
+      s <- qRunIO $ TIO.hGetContents h
+      return s
 
 -- calls parse to Quasi.parse individual entities in isolation
 -- afterwards, sets references to other entities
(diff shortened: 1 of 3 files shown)


persistent-template/Database/Persist/TH.hs
```diff
@@ -111,14 +112,29 @@ persistLowerCase = persistWith lowerCaseSettings
...
+persistManyFileWith ps fps = do
 #ifdef GHC_7_4
-    qAddDependentFile fp
+    mapM_ qAddDependentFile fps
 #endif
-    h <- qRunIO $ SIO.openFile fp SIO.ReadMode
-    qRunIO $ SIO.hSetEncoding h SIO.utf8_bom
-    s <- qRunIO $ TIO.hGetContents h
+    ss <- mapM getS fps
+    let s = mconcat ss
```
naushadh: Noticed a minor hiccup when playing with this. Should any of the `Entity` files be missing a line-break at the EOF, then we end up gluing them wrong. Ex:

- Foo/Entity
```
Foo
  name Text
  age Natural
```

- Bar/Entity
```
Bar
  title Text

```

- currently will be read as:
```
Foo
  name Text
  age NaturalBar
  title Text

```

Should this be updated to `T.intercalate "\n" ss` to be fault tolerant? Or should we leave it as is and just document that all files must end with a line break (since it is a [common convention](https://stackoverflow.com/a/729795) anyway)?

MaxGabriel:
I’m strongly in favor of the fault tolerant approach—it’s great when everything just works!

naushadh:
@MaxGabriel I tried to reproduce it with my sample project [naushadh/persistent-examples/model-organization](https://github.com/naushadh/persistent-examples/tree/master/model-organization#timings) with 20 entities, but I see ~160MB of memory usage for all cases:
- monolith (no .stack-work)
- monolith with a `Model` change
- component style (no .stack-work)
- component style with a `Model` change

Do you have a sample repo/gist/snippet handy that can repro the multi GB compile issue? I can try breaking it apart component style to see if it makes it go away.

MaxGabriel:
@naushadh So I created ~30 models, each deriving `Show`, `Eq`, and `Generic` in this repo: https://github.com/MaxGabriel/persistence-of-memory

I tried using both a single model file and split across 3 model files, but the compile time and memory usage are pretty similar for me (~85 seconds, ~950 MB to compile). I'm compiling with `stack build --ghc-options='-j +RTS -s -RTS'`. Am I missing something there?

naushadh:
@MaxGabriel the repo didn't actually utilize the new API `persistManyFileWith` for migration. I've pushed a PR that shows how to utilize the API along with consuming the smaller model files from separate modules (which GHC can then build in parallel).

Stats

samples |memory (MB)|time (s)
--|--|--
baseline|877|49.385
new|824|25.079
new+tricks*|1009|21.373

Tricks (since we aren't actually saving much memory)
- last sample shows cut down GC time by throwing more memory at it
- hyper-threading can be exploited by invoking `-N8` (on a 4 core machine) for GHC 8.2+. However the current example does not enough independent models to demo its benefits.

MaxGabriel:
Ah ok great. I had used `persistManyFileWith` but I didn't realize I needed to add separate modules and run `mkPersist` in each of them.

I made a PR to your branch adding an example to the documentation: https://github.com/naushadh/persistent/pull/3

Otherwise I think this is ready to merge. One final thought: How do you feel about the name? I'm constantly having to look up `persistManyFileWith` because it's kind of an awkward, but I guess it matches existing conventions.

naushadh:
Docs look beautiful! Much easier to consume than following 3 levels of urls.

Yeah I thought the same, so I went with convention. I'm open to suggestions.

MaxGabriel:
Ok, released on hackage! Thanks for your work on this, it'll be great to point Persistent users to this solution

PR #793, opened 2018-03-01
JSON addition to Database.Persist.Postgresql

Vlix:
Added `PersistField` and `PersistFieldSql` instances for `Data.Aeson.Value` and new `Filter` operators `(@>.)` and `(<@.)` that work with `Value`.

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql/JSON.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql/JSON.hs
@@ -1,5 +1,7 @@
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 {-# LANGUAGE OverloadedStrings #-}
+
+-- | Filter operators for JSON values added to PostgreSQL 9.4
 module Database.Persist.Postgresql.JSON
   ( (@>.)
   , (<@.)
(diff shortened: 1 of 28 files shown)


persistent-postgresql/Database/Persist/Postgresql/JSON.hs
```diff
@@ -0,0 +1,146 @@
...
+-- true    @> true    --> True
+-- true    @> false   --> False
+-- false   @> true    --> False
+-- true    @> "true"  --> False
+--
+-- null    @> null    --> True
+-- null    @> 23      --> False
+-- null    @> "null"  --> False
+-- null    @> {}      --> False
+-- @
```
MaxGabriel: I don' think these haddocks are rendering how you want them to

image

persistent-postgresql/Database/Persist/Postgresql/JSON.hs
```diff
@@ -0,0 +1,146 @@
...
+infix 4 @>., <@.
+
+-- | This operator checks inclusion of the JSON value
+-- on the right hand side in the JSON value on the left
+-- hand side.
+--
+-- Any empty Object matches any object
+-- @
+-- {"a":1,"b":false} @> {} --> True
+-- @
```
MaxGabriel: This documentation has quite a few examples, and I think if formatted correctly would be pretty huge. My default would be to hide them in an `==== __Examples__` collapsible section, but it might be OK since this module is quite small

persistent-postgresql/Database/Persist/Postgresql/JSON.hs
```diff
@@ -0,0 +1,146 @@
...
+--
+-- null    @> null    --> True
+-- null    @> 23      --> False
+-- null    @> "null"  --> False
+-- null    @> {}      --> False
+-- @
+(@>.) :: EntityField record Value -> Value -> Filter record
+(@>.) field val = Filter field (Left val) $ BackendSpecificFilter " @> "
+
+-- | Same as `(<@.)` except the inclusion check is reversed.
```
MaxGabriel: I think you want single quotes, not backticks, for that

persistent-postgresql/Database/Persist/Postgresql/JSON.hs
```diff
@@ -0,0 +1,146 @@
...
+fromPersistValueJsonB :: FromJSON a => PersistValue -> Either Text a
+fromPersistValueJsonB (PersistText t) =
+    case eitherDecodeStrict $ TE.encodeUtf8 t of
+      Left str -> Left $ fromPersistValueParseError "FromJSON" t $ T.pack str
+      Right v -> Right v
+fromPersistValueJsonB (PersistByteString bs) =
+    case eitherDecodeStrict bs of
+      Left str -> Left $ fromPersistValueParseError "FromJSON" bs $ T.pack str
+      Right v -> Right v
+fromPersistValueJsonB x = Left $ fromPersistValueError "FromJSON" "string or bytestring" x
```
MaxGabriel: Instead of `bytestring`, could you use a database name like Postgres' `bytea`?

Vlix:
Updated a lot. Should also pass all tests.

MaxGabriel:
Thanks for your work on this! Great patch; really like the documentation with all the examples, and the copious tests. I've merged and released on Hackage 👍

Issue #609, opened 2016-09-16
Proper array fields


edgmnt:
Arrays/lists are currently stored as serialized JSON. The problem with this representation is we can't run DB queries against such arrays.

For example, suppose you have a database of a library. There's a table that stores each shelf along with the books on it (in order) and another table that stores book information. How do you query the database for shelves which contain books by E. A. Poe? With proper arrays, a database may support a query like:

SELECT ... FROM shelf, book WHERE book.id = ANY(shelf.books) AND book.author = "E. A. Poe"

This is primarly geared towards / inspired by Postgres arrays. Could we support this?

gregwebs:
Yes, that has always been the plan

parsonsmatt:
`persistent-postgresql-2.10.0` supports arrays with the new `PersistArray :: [PersistValue] -> PersistValue` constructor. #884

Issue #571, opened 2016-06-04
malformed array literal


Qinka:
I inserted a row of data, and got the followed error.

```
SqlError {sqlState = \"22P02\", sqlExecStatus = FatalError, sqlErrorMsg = \"malformed array literal: \\\"[1,1,1]\\\"\", sqlErrorDetail = \"Missing \\\"]\\\" after array dimensions.\", sqlErrorHint = \"\"}
```

And my Postgres has such output:

```
ERROR:  malformed array literal: "[1,1,1]" at character 200
DETAIL:  Missing "]" after array dimensions.
STATEMENT:  INSERT INTO "table_task_info"("key_tid","key_ew","key_ns","key_r","key_wei","key_size","key_note","key_cost","key_des") VALUES('T2016-06-048c330b133d149fd972ad0bd700e5664bb71e1cf1',1.0,1.0,100.0,1.0,'[1,1,1]',null,100,null)
```

Qinka:
I don't whether I used PostgreSQL array and persistent in a wrong way, but to seemed that some functions is not suitable.

I don't know why array is transformed to JSON's array but not to SQL's array. However, it do.

gregwebs:
Are you using persistent for migrations?

Qinka:
I just use `insert` function.

Qinka:
I think `listToJSON` (at [Database/Persist/PostgreSQL.hs:line:305](https://github.com/yesodweb/persistent/blob/master/persistent-postgresql/Database/Persist/Postgresql.hs#L305) ) should be changed to `PG.PGArray`.

And if you think it's necessary, I will fork and send a new pull request.

gregwebs:
PR would be great.

Qinka:
Sorry, I'm not good at English, and what does PR stand for?

psibi:
Pull Request. :)

Qinka:
😂

MaxGabriel:
For the longest time I thought "LGTM" meant "Let's get this merged"

parsonsmatt:
`persistent-postgresql-2.10.0` supports arrays with the new `PersistArray :: [PersistValue] -> PersistValue` constructor. #884

PR #799, opened 2018-03-09
Use native PostgreSQL arrays when serializing lists

nmk:
There are a number of issues requesting native PostgreSQL arrays (#609, #571, and a few older ones), instead of the current roundtrip through JSON. I looked at the code and it seems the only change needed to support this is the one-line change in this PR. I tested this locally and the list serialization to PGArray works as intended and tests (run as `stack test` in the root directory) all pass.

If this is accepted, it would be great to backport to the 2.6.* version which is the latest available on stackage as well.

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -422,7 +422,7 @@ instance PGTF.ToField P where
     toField (P (PersistTimeOfDay t))   = PGTF.toField t
     toField (P (PersistUTCTime t))     = PGTF.toField t
     toField (P PersistNull)            = PGTF.toField PG.Null
-    toField (P (PersistList l))        = PGTF.toField $ listToJSON l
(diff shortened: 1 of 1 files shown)


psibi:
I don't think it's that simple. IIRC, You also need to deal with proper migration.

MaxGabriel:
There are also some limitations like Postgres sub arrays needing to have the same length as the parent array, or something like that.

@nmk it’s kind of weird, you want this command to run tests for a specific backend:

`stack test persistent-test --flag persistent-test:<backend> --exec persistent-test`

MaxGabriel:
Or see here for test failures: https://travis-ci.org/yesodweb/persistent/jobs/351408374

nmk:
I will have a go at adjusting the migrations and adding a new SqlType to host the list-like types.

parsonsmatt:
@nmk A recent patch #884 has been done which adds `PersistArray` constructor to the `PersistValue` type. It's treated a a `List` for other backends, but uses arrays with postgresql.

PR #821, opened 2018-06-30
added: PersistentTest: upsert fail, it expected equivalent repsert

ncaq:
I see
[haskell - Persistent `upsert` isn't working -
Stack Overflow](https://stackoverflow.com/questions/50050431/persistent-upsert-isnt-working)

The question is that if the second update argument of upsert is empty, the data will not be updated.

Certainly it is written in the document comment as follows.

> leaving this empty is the equivalent of performing a repsert on a unique key
>
> [Database.Persist.Class](https://www.stackage.org/haddock/lts-11.15/persistent-2.8.2/Database-Persist-Class.html#v:upsert)

According to the document, upsert needs to behave like repsert if the second argument is empty.

However, looking at the source code, it seems that such behavior is not supposed to be.

So I wrote a test that failed.

To send test results, this bug report uses pull request instead of issue.

I do not know which is better for me, whether to modify the document or modify the implementation.

It is fail test, I do not check version number and CI.

After submitting your PR:

### DIFF
--- a/persistent-test/src/PersistentTest.hs
+++ b/persistent-test/src/PersistentTest.hs
@@ -573,6 +573,15 @@ specs = describe "persistent" $ do
         upsertAge (entityVal update') @== 5
         upsertExtra (entityVal update') @== "extra"
 #endif
+    it "leaving this empty is the equivalent of performing a repsert on a unique key" $ db $ do
+        Entity key _ <- insertEntity (Upsert "foo" "initial" "" 2)
+        repsert' <- do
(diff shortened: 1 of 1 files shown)


parsonsmatt:
This is great, thanks for the PR :smile: 

The [documentation](https://www.stackage.org/haddock/lts-13.15/persistent-2.9.1/Database-Persist-Class.html#v:upsert) has been changed to reflect that an empty list of updates is *not* equivalent to a `repsert`.

PR #829, opened 2018-08-26
Create separate persistent-qq (fixes #827)

### LINKED ISSUE #827: haskell-src-meta is a heavyweight dependency
Sorry I'm only noticing this a year later, I realize I was pinged on #717. That PR introduced a dependency on haskell-src-meta in order to allow for parsing Haskell expressions. Unfortunately, that library introduces (via itself and dependencies) a relatively heavy compile time cost. I would like to propose that we move the `Database.Persist.Sql.Raw.QQ` module to a separate package so that users of `persistent` itself do not need to pay that cost. Any objection? Note that this would be a breaking change.

CC @felixSchl @cdepillabout @paul-rouse (commenters on the PR).

snoyberg:


snoyberg:
I've just uploaded, and I'll look into Yesod next.

PR #833, opened 2018-10-01
Fix #832: `repsertMany == mapM_ (uncurry repsert)` and is atomic.

### LINKED ISSUE #832: `repsertMany` does not match `mapM_ (uncurry repsert)`
## Issue

`repsert` works like so:
- Find record by ID
- If found, replace record for ID
- else, insert record with ID

`repsertMany` works like so:
- Find records by Ids
- new ids = given ids `diff` found ids
- insert records matching new ids

The following step is missing in `repsertMany` to make it symmetric with `mapM_ (uncurry repsert)`:
- replace records matching found ids

Issue was introduced at the very conception of the feature, and there are no tests to verify this API.

## Suggested fix

Since there is no batch means to add the missing step mentioned above, we should perhaps let each driver implement its own `repsertMany` natively akin to `putMany`; i.e., create a new field in `SqlBackend` called `connRepsertManySql` before letting each backend define it.

Bonus, this will actually address race condition errors/deadlocks arising from non-atomic client side operations (SELECT + INSERT) on the same table. I've run into parallel insert errors with `repsertMany` but never with the natively/atomically implemented `putMany` using Postgres 10.

naushadh:
Fix [832](https://github.com/yesodweb/persistent/issues/832): `repsertMany` now matches `mapM_ (uncurry repsert)` and is atomic for supported sql back-ends.

After submitting your PR:

### DIFF
--- a/persistent-mysql/ChangeLog.md
+++ b/persistent-mysql/ChangeLog.md
@@ -1,3 +1,9 @@
+# Changelog for persistent-mysql
+
+## 2.9.2
+
+* Fix [832](https://github.com/yesodweb/persistent/issues/832): `repsertMany` now matches `mapM_ (uncurry repsert)` and is atomic.
+
 ## 2.9.1
 
 * Added support for SQL isolation levels to via SqlBackend. [#812]
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -69,7 +69,7 @@ import qualified Data.Text.Encoding as T
 
 import Database.Persist.Sql
 import Database.Persist.Sql.Types.Internal (mkPersistBackend, makeIsolationLevelStatement)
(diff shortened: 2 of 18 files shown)


psibi:
@naushadh Can sqlite upgrade changes be a single individual commit ?

naushadh:
@psibi you mean moving the changes to cbits into a single commit? I can certainly do that. Just note tho, the batched `putMany` and `repsertMany` implementations only work with sqlite>`3.25.0`.

naushadh:
[Build error](https://travis-ci.org/yesodweb/persistent/jobs/436379942) appears to be random in nature since it [passes in my fork](https://travis-ci.org/naushadh/persistent/builds/436379917).

naushadh:
@psibi done https://github.com/yesodweb/persistent/pull/833/commits/070024f0bb86bda87565669476ff5960b230f5a3

psibi:
Thanks!

Issue #783, opened 2018-02-17
Eliminate `IsSqlBackend` constraint from various `withSqlConn` and `withSqlPool` functions


merijn:
In #772 I created a new datatype to expose underlying SQLite internals more easily. However, due to the `IsSqlBackend` restriction on various functions in `Database.Persist.Sql` I find myself effectively reimplementing these functions to work with `RawSqlite` which is an instance of `PersistCore` and `HasPersistBackend` but *not* `IsSqlBackend` as that requires a `IsPersistBackend` instance which turns a `BaseBackend` (e.g. `SqlBackend`) into the relevant type. This can't be done, since the SQLite pointer can't be materialised from thin air.

At the same time, none of these functions seem to actually require an `IsSqlBackend` constraint as the typeclass isn't actually used. It's just there for the polymorphic return type, but then that can easily be propagated from the callers of `withSqlPool`. Removing these constraints would let these functions be reused for more types, avoiding the need to reimplement them.

MaxGabriel:
Hm, that seems like it could be a big breaking change, wouldn't code like this (taken from the Yesod SQLite scaffolding) be broken, because the `pool` that `createSqlitePool` creates no longer has the `IsSqlBackend` class?

```haskell
    -- Create the database connection pool
    pool <- flip runLoggingT logFunc $ createSqlitePool
        (sqlDatabase $ appDatabaseConf appSettings)
        (sqlPoolSize $ appDatabaseConf appSettings)

    -- Perform database migration using our application's logging settings.
    runLoggingT (runSqlPool (runMigration migrateAll) pool) logFunc
```

Correct me if I'm wrong on this, just want to understand the full implications of this change

merijn:
I'm only proposing the removal of the constraint on the generic functions inside `Database.Persist.Sql`, not the backend specific ones like `Database.Persist.Sqlite`. The constraint is *obviously* needed on, e.g. `withSqlitePool` and `createSqlitePool`. My observation is just that `Database.Persist.Sql.withSqlPool` (and friends) don't actually use the IsSqlBackend constraint, but due to having it can be reused to implement, for example, the `withRawSqlitePool` I proposed in #772 (because that doesn't fulfil the constraint). Backend specific variants like `withSqlitePool` should definitely keep the constraint, but if those have it, it becomes completely superfluous on `withSqlPool`.

MaxGabriel:
[...]

PR #772, opened 2018-01-12
Expose underlying raw SQLite database pointer.

merijn:
Allows people to use the FFI to call SQLite's C API functions. Fixes #771.

Before submitting your PR, check that you've:

- [x] Bumped the version number
- [x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

- [x] Update the Changelog.md file with a link to your PR
- [x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

PR #839, opened 2018-11-05
Change SQL functions to use `BackendCompatible` over `IsSqlBackend`.

merijn:
This addresses #783 so that #772 can be updated to use this and finally be merged.

### DIFF
--- a/persistent-sqlite/Database/Persist/Sqlite.hs
+++ b/persistent-sqlite/Database/Persist/Sqlite.hs
@@ -67,7 +67,7 @@ import Lens.Micro.TH (makeLenses)
 -- Note that this should not be used with the @:memory:@ connection string, as
 -- the pool will regularly remove connections, destroying your database.
 -- Instead, use 'withSqliteConn'.
-createSqlitePool :: (MonadLogger m, MonadUnliftIO m, IsSqlBackend backend)
+createSqlitePool :: (MonadLogger m, MonadUnliftIO m, IsSqlBackend backend, BackendCompatible SqlBackend backend)
                  => Text -> Int -> m (Pool backend)
 createSqlitePool = createSqlitePoolFromInfo . conStringToInfo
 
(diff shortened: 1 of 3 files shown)


merijn:
I'm not sure if/how this breaks the non-sqlite persistents, as I can't get them to build reliably on my system anyway. We'll see what CI says.

parsonsmatt:
@merijn This PR looks like a good start, but we'll need to continue propagating the changes through so that CI passes.

merijn:
@parsonsmatt If someone else is willing to take over the PR that'd probably be the best way to make progrss, as I'm basically swamped in real life at the moment. I can keep an eye on any related PRs/changes and chime in if needed, but can't really dedicate a bunch of time to fixing up the other persistent packages.

Vlix:
I could see if I can go through all the non-SQLite backends and just propagate these changes next time I feel like I need some distraction.
Is it basically replacing all `IsSqlBackend` with `BackendCompatible SqlBackend`?

merijn:
@Vlix Basically, yes. `IsSqlBackend` is unnecessarily restrictive. See also the discussions linked in my first comment.

Vlix:
I've started working on this, and I think I'm pretty much done, but I've removed most `backend` parameters in the MySQL/PostgreSQL/SQLite libraries where `IsSqlBackend backend` was used, because it was just a `SqlBackend` being `mkPersistBackend`-ed to `backend` to be used as a `SqlBackend`.

Only part that's giving me trouble is the `rawQueryRes` function... changing the `IsSqlBackend` to `BackendCompatible SqlBackend backend` turns into a wild goose chase until I get to `selectKeysRes` and getting rid of `BaseBackend (BaseBackend backend)` seems impossible.

parsonsmatt:
The `BaseBackend (BaseBackend backend)` thing always felt very hacky to me

Vlix:
@merijn
Is there a reason you wanted the `BackendCompatible SqlBackend backend` in the `persistent-sqlite` library constraining the `backend`, instead of it just being a `SqlBackend`?
Going through the types (signatures), just setting it to `SqlBackend` makes the most sense, since it's a SQL DB, and the root function causing it, is `warpConnectionInfo` which literally outputs a `SqlBackend`.

merijn:
@Vlix 
The problem was basically that all the bracket-like functions are very restricted by `IsSqlBackend`, because it requires that any instance can be created/project from *just* an `SqlBackend`. Whereas, for example, the `persistent-sqlite` PR for exposing the internal SQLite pointer carries additional data (i.e. the actual pointer for the SQLite connection) that cannot be recreated from thin air when you have just an `SqlBackend`.

The idea behind the original introduction of `BackendCompatible` was to make the constraints in persistent more flexible, and discussion in the other PR led to going down this road.

Vlix:
I've managed to pretty much remove the need for `IsSqlBackend` from the code base. I hope it didn't break anything... I don't think it should.
My PR is over [here](https://github.com/yesodweb/persistent/pull/886)

parsonsmatt:
Yup, #886 covers this one - I am going to close this PR out since #886 is nearly ready to go.

Issue #841, opened 2018-11-11
[Suggestion] Add `Like` in `PersistFilter`


pythonissam:
Currently, we need to use [`BackendSpecificFilter`](https://www.stackage.org/haddock/lts-12.17/persistent-2.8.2/src/Database.Persist.Types.Base.html#PersistFilter) to represent `LIKE` operator. Since there's no sanity check for wildcard characters, there's chance that they are injected via user's input.

For example, let's say we want to `SELECT` books `LIKE` authors which start with the user's input, we can write like this:

```haskell
books <- runDB $ selectList [Filter BookAuthor (Left $ concat [author, "%"]) (BackendSpecificFilter "LIKE")] [Asc BookId]
```

If users pass inputs something like `%Haskell` as `author`, `WHERE` clause would be `%Haskell%` and the result would become books which `contain` `Haskell`, which we didn't expect.

To solve this problem, I think we need to add value `Like` in [PersistFilter](https://www.stackage.org/haddock/lts-12.17/persistent-2.8.2/src/Database.Persist.Types.Base.html#PersistFilter) and provide helper functions like [these](https://www.stackage.org/haddock/lts-12.17/persistent-2.8.2/Database-Persist.html#g:3).

Any thoughts?

psibi:
Sounds, like a good feature. Would you like to send a PR for this ? A good start would be to implement the feature along with tests for it and we can probably iterate from there.

PR #845, opened 2018-12-23
Escape wildcards in like

pythonissam:

After submitting your PR:

### DIFF
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -878,6 +878,8 @@ filterToBSON fname v filt = case filt of
     showFilter In = "$in"
     showFilter NotIn = "$nin"
     showFilter Eq = error "EQ filter not expected"
+    showFilter Like = error "LIKE filter not expected"
+    showFilter NotLike = error "NOT LIKE filter not expected"
     showFilter (BackendSpecificFilter bsf) = throw $ PersistMongoDBError $ T.pack $ "did not expect BackendSpecificFilter " ++ T.unpack bsf
 
 
--- a/persistent-test/persistent-test.cabal
+++ b/persistent-test/persistent-test.cabal
@@ -138,6 +138,9 @@ library
                    , tagged
                    , old-locale
 
+    if impl (ghc < 8.0)
+      build-depends:   semigroups == 0.18.*
+
    if flag(nooverlap)
      cpp-options: -DNO_OVERLAP
 
--- a/persistent-test/src/PersistentTest.hs
+++ b/persistent-test/src/PersistentTest.hs
@@ -59,6 +59,10 @@ import Init
 import PersistTestPetType
 import PersistTestPetCollarType
 
+#ifndef WITH_MONGODB
+import Data.Semigroup ((<>))
+#endif
+
 #ifdef WITH_NOSQL
 mkPersist persistSettings [persistUpperCase|
 #else
@@ -223,6 +227,25 @@ specs = describe "persistent" $ do
       _ <- insert p
       ps <- selectList [FilterAnd []] [Desc PersonAge]
       assertNotEmpty ps
+
+#ifndef WITH_MONGODB
+  it "LIKE" $ db $ do
+      let p = Person "Shakespeare" 52 Nothing
+      _ <- insert p
+
+      let maltext = "re%"
+      ps <- selectList [likeWithEscape '@' PersonName ("%espea" <> toEscapedLikeText maltext)] []
+      assertEmpty ps
+
+  it "NOT LIKE" $ db $ do
+      let p = Person "Shakespeare" 52 Nothing
+      _ <- insert p
+
+      let maltext = "re%"
+      ps <- selectList [notLikeWithEscape '@' PersonName ("%espea" <> toEscapedLikeText maltext)] []
+      assertNotEmpty ps
+#endif
+
 
   it "order of opts is irrelevant" $ db $ do
       let eq (a, b, _) (c, d) = (a, b) @== (c, d)
--- a/persistent/Database/Persist.hs
+++ b/persistent/Database/Persist.hs
@@ -34,7 +34,9 @@ module Database.Persist
     , (=.), (+=.), (-=.), (*=.), (/=.)
 
       -- * Query filter combinators
-    , (==.), (!=.), (<.), (>.), (<=.), (>=.), (<-.), (/<-.), (||.)
+    , (==.), (!=.), (<.), (>.), (<=.), (>=.), (<-.), (/<-.), (||.), likeWithEscape, notLikeWithEscape
+
+    , EscapedLikeText(..)
 
       -- * JSON Utilities
     , listToJSON
@@ -50,6 +52,7 @@ import Database.Persist.Types
 import Database.Persist.Class
 import Database.Persist.Class.PersistField (getPersistMap)
 import qualified Data.Text as T
+import Data.Text (Text)
 import Data.Text.Lazy (toStrict)
 import Data.Text.Lazy.Builder (toLazyText)
 import Data.Aeson (toJSON, ToJSON)
@@ -294,6 +297,70 @@ f >=. a  = Filter f (Left a) Ge
 
 infix 4 <-., /<-.
 (<-.), (/<-.) :: forall v typ.  PersistField typ => EntityField v typ -> [typ] -> Filter v
+
+
+-- |
+-- Check if value is like given pattern. This function takes
+-- field, a character to escape wildcards and a like pattern
+--
+-- === __Example usage__
+--
+-- @
+-- selectLikeSimon :: MonadIO m => ReaderT SqlBackend m ()
+-- selectLikeSimon = selectList [likeWithEscape '@' UserName "%Si%"]
+-- @
+--
+-- The above query when applied on dataset-1, will produce this:
+--
+-- > +-----+-----+-----+
+-- > |id   |name |age  |
+-- > +-----+-----+-----+
+-- > |2    |Simon|41   |
+-- > +-----+-----+-----+
+--
+-- and the SQL like the following:
+--
+-- > SELECT id, name, age, color FROM Person WHERE name LIKE "%si%" '@'
+--
+-- Undesirable wildcards, such as the one mixed in user inputs
+-- can be escaped explicitly:
+--
+-- > maltext = "mon%"
+-- > selectList [likeWithEscape '#' UserName $ "%Si" <> toEscapedLikeText maltext]
+
+likeWithEscape :: Char -> EntityField v Text -> EscapedLikeText-> Filter v
+likeWithEscape c f t = Filter f (Right [runWildcardEscape t c, T.singleton c]) Like
+
+-- |
+-- Check if value is not like given pattern
+--
+-- === __Example usage__
+--
+-- @
+-- selectNotLikeSimon :: MonadIO m => ReaderT SqlBackend m ()
+-- selectNotLikeSimon = selectList [notLikeWithEscape '@' UserName "Simon"]
+-- @
+--
+-- The above query when applied on dataset-1, will produce this:
+--
+-- > +-----+-----+-----+
+-- > |id   |name |age  |
+-- > +-----+-----+-----+
+-- > |1    |SPJ  |40   |
+-- > +-----+-----+-----+
+--
+-- and the SQL like the following:
+--
+-- > SELECT id, name, age, color FROM Person WHERE name NOT LIKE "Simon" '@'
+--
+-- Undesirable wildcards, such as the one mixed in user inputs
+-- can be escaped explicitly:
+--
+-- > maltext = "mon%"
+-- > selectList [notLikeWithEscape '@' UserName $ "%Si" <> toEscapedLikeText maltext]
+
+notLikeWithEscape :: Char -> EntityField v Text -> EscapedLikeText -> Filter v
+notLikeWithEscape c f t = Filter f (Right [runWildcardEscape t c, T.singleton c]) NotLike
 
 -- | Check if value is in given list.
 --
--- a/persistent/Database/Persist.hs
+++ b/persistent/Database/Persist.hs
@@ -303,6 +303,8 @@ infix 4 <-., /<-.
 -- Check if value is like given pattern. This function takes
 -- field, a character to escape wildcards and a like pattern
 --
+-- @since 2.9.2
+--
 -- === __Example usage__
 --
 -- @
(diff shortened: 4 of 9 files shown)


pythonissam:
By default, `like` uses `'#'` as an escape character.

I have no confidence if my English is right as usual. So glad if you check it.

pythonissam:
issue ref: #841

psibi:
@pythonissam Sorry it took some time for me to review the PR. The PR looks good and it is a good start. What kind of sql query is generated by this code:

```
ps <- selectList [like PersonName ("Sha%es_eare")] []
```

pythonissam:
@psibi No problem at all. It should generate something like this:

```sql
SELECT Name FROM Person LIKE 'Sha%es_eare'
```

psibi:
@pythonissam Thanks!

 Is there any reason why the sql query uses `#` as the escape character ?

I think for this haskell code:

```
ps <- selectList [like PersonName ("Sha%es_eare")] []
```
the corresponding sql query should not have any escape character at all:

```
"SELECT \"id\", \"name\", \"age\", \"color\" FROM \"Person\" WHERE (\"name\" LIKE ?); [PersistText \"Sha%es_eare\"]"
```

Also in your above generated query which you have written in your comment, it contains "NOT LIKE". Don't you think it should only contain "LIKE" ?

When we use the function `likeWithEscape`, for that I think we can have the `ESCAPE` clause in the generated query.

psibi:
> It's because people around me seem to use # for default escape character (but this can be a Japanese convention. If this is wrong, please tell me).

Yeah, from what I see there should not be no default escape character.

> Is that so? Then, what do you think is the suitable name for this function?

I'm fine with the current name.

Thanks for the haddock documentation. Can you also add the generated sql query for both of the cases there ?

Sorry for my late reply and belated New year wishes!

pythonissam:
> Yeah, from what I see there should not be no default escape character.

So, just providing `likeWithEscape` and `notLikeWithEscape` is better? I feel it's a bit hassle to always set an escape character by myself...
I don't stick using `#`, but for example, aren't there any alternatives?

> Thanks for the haddock documentation. Can you also add the generated sql query for both of the cases there ?

Yeah, sure:)

psibi:
> So, just providing likeWithEscape and notLikeWithEscape is better? I feel it's a bit hassle to always set an escape character by myself...

I think four functions can be exported:
* like
* notLike
* likeWithEscape
* notLikeWithEscape

The last two one will contain the `ESCAPE` clause in the query.

> I don't stick using #, but for example, aren't there any alternatives?

I was not able to find any. Is there any official resource which recommends any specific default ESCAPE character ?

pythonissam:
> I was not able to find any. Is there any official resource which recommends any specific default ESCAPE character ?

Me, neither... Then, alright. Your opinion seems better. I'm going to change as you said.

pythonissam:
@psibi Excuse me. There was one more thing to consider.

If `like` and `notLike` don't escape wildcards by default, then it means that the vulnerability I pointed remains (wildcard injection). My first motivation was to create safe `LIKE`-related utilities to prevent the vulnerability. I guess ordinary people except security professionals just use `like` or `notLike`.  So, if we go this way, it becomes more likely to mix the vulnerability. If this is permissive, then I'll follow your opinion.

What do you think?

psibi:
@pythonissam Fair points. The same thing is pointed here also: https://www.owasp.org/index.php/SQL_Injection_Prevention_Cheat_Sheet#Escaping_Wildcard_characters_in_Like_Clauses

In that case, if we want to have the `ESCAPE` clause too, how about just exporting these two functions:
```
likeWithEscape
notLikeWithEscape
```

and avoid the exporting of `like` and `notLike` altogether ?

psibi:
@pythonissam Any updates in this ?

pythonissam:
@psibi Sorry for late... I was too busy in this week.

> In that case, if we want to have the ESCAPE clause too, how about just exporting these two functions

No objection. I'll send you a PR ASAP.

pythonissam:
@psibi Added the sql queries and rebased to resolve merge conflicts. Is this alright?

psibi:
@pythonissam Thanks! I'm slightly occupied this week. Let me get back to you on coming Friday.

Vlix:
I see this hasn't been implemented yet, but wanted to note that, from what I could tell, it only escapes the `%` and `_` characters, but it should also escape the escape character.
If you wouldn't also escape the escape character, it might match (or not) while you'd expect it shouldn't (or should), depending on which escape character is used.

---

The following (trying to match on `Shakespeare`):
```
wut = "\re"
selectList [likeWithEscape '\' PersonName ("%espea" <> toEscapedLikeText wut)]
```
should result in:
```
SELECT * FROM person
  WHERE name LIKE '%espea\\re' ESCAPE '\'
```

but I'm pretty sure the part after `LIKE` in the current implementation would result in:
`... LIKE '%espea\re' ESCAPE '\'` (note the single `\` in `%espea\re`)
since, like I said, only `%` and `_` are escaped, and that last output WOULD match (at least in PostgreSQL), though you'd expect, given it's escaped input, it shouldn't. (Because a lonely escape character is ignored)

persistent-mongoDB/Database/Persist/MongoDB.hs
```diff
@@ -878,6 +878,8 @@ filterToBSON fname v filt = case filt of
     showFilter In = "$in"
     showFilter NotIn = "$nin"
     showFilter Eq = error "EQ filter not expected"
+    showFilter Like = error "LIKE filter not expected"
+    showFilter NotLike = error "NOT LIKE filter not expected"
```
parsonsmatt: We should be able to support the `Like` query in MongoDB: https://stackoverflow.com/questions/3305561/how-to-query-mongodb-with-like

persistent-test/persistent-test.cabal
```diff
@@ -138,6 +138,9 @@ library
                    , tagged
                    , old-locale
 
+    if impl (ghc < 8.0)
+      build-depends:   semigroups == 0.18.*
```
parsonsmatt: Fortunately we have recently dropped support for GHC < 8, so this won't be necessary :smile:

persistent-test/src/PersistentTest.hs
```diff
@@ -224,6 +228,25 @@ specs = describe "persistent" $ do
...
+      assertEmpty ps
+
+  it "NOT LIKE" $ db $ do
+      let p = Person "Shakespeare" 52 Nothing
+      _ <- insert p
+
+      let maltext = "re%"
+      ps <- selectList [notLikeWithEscape '@' PersonName ("%espea" <> toEscapedLikeText maltext)] []
+      assertNotEmpty ps
+#endif
```
parsonsmatt: The test suite has been revamped to not use CPP to split out backends. I'd prefer to see a new module `LikeTest` that can be used to demonstrate support for `LIKE` on many backends, including MongoDB when that is done.

persistent/Database/Persist.hs
```diff
@@ -295,6 +298,74 @@ f >=. a  = Filter f (Left a) Ge
...
+--
+-- > +-----+-----+-----+
+-- > |id   |name |age  |
+-- > +-----+-----+-----+
+-- > |2    |Simon|41   |
+-- > +-----+-----+-----+
+--
+-- and the SQL like the following:
+--
+-- > SELECT id, name, age, color FROM Person WHERE name LIKE "%si%" '@'
```
parsonsmatt: Nitpick: The function has `"%Si%"` but this example has `%si%`. Is the function case sensitive?

persistent/Database/Persist.hs
```diff
@@ -295,6 +298,74 @@ f >=. a  = Filter f (Left a) Ge
...
+--
+-- and the SQL like the following:
+--
+-- > SELECT id, name, age, color FROM Person WHERE name NOT LIKE "Simon" '@'
+--
+-- Undesirable wildcards, such as the one mixed in user inputs
+-- can be escaped explicitly:
+--
+-- > maltext = "mon%"
+-- > selectList [notLikeWithEscape '@' UserName $ "%Si" <> toEscapedLikeText maltext]
```
parsonsmatt: Excellent docs :a: :heavy_plus_sign:

persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
```diff
@@ -249,6 +249,12 @@ filterClauseHelper includeTable includeWhere conn orNull filters =
                            let xxs = transpose (map fromPersistList allVals)
                                sqls=map (\(a,xs) -> connEscapeName conn (fieldDB a) <> showSqlFilter pfilter <> "(" <> T.intercalate "," (replicate (length xs) " ?") <> ") ") (zip (compositeFields pdef) xxs)
                            in (wrapSql (T.intercalate " or " (map wrapSql sqls)), concat xxs)
+                        ([PersistList xs], Like, _) ->
+                           let sqlcl=T.intercalate " and " (map (\a -> connEscapeName conn (fieldDB a) <> showSqlFilter pfilter <> "?" <> " ESCAPE ? ") (compositeFields pdef))
```
parsonsmatt: Formatting nitpick: Can we put spaces around the `=` and try to keep the line length down
```suggestion
                           let sqlcl = T.intercalate " and " 
                                   . map (\a -> mconcat
                                       [ connEscapeName conn (fieldDB a)
                                       , showSqlFilter pfilter
                                       , "?", " ESCAPE ? ") 
                                       ]
                                   $ compositeFields pdef
```

persistent/Database/Persist/Types/Base.hs
```diff
@@ -1,6 +1,9 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE OverloadedStrings #-}
+#if __GLASGOW_HASKELL__ >= 861
+{-# LANGUAGE DerivingVia #-}
+#endif
```
parsonsmatt: As cool as this feature is, I'd rather not have CPP if we're going to need the boilerplate-y instance anyway.

persistent/Database/Persist/Types/Base.hs
```diff
@@ -461,7 +466,7 @@ data SqlType = SqlString
              | SqlOther T.Text -- ^ a backend-specific name
     deriving (Show, Read, Eq, Typeable, Ord)
 
-data PersistFilter = Eq | Ne | Gt | Lt | Ge | Le | In | NotIn
+data PersistFilter = Eq | Ne | Gt | Lt | Ge | Le | In | NotIn | Like | NotLike
```
parsonsmatt: I feel like `Like` and `NotLIke` are really more appropriate as a `BackendSpecificFilter` since they're only really supported in SQL databases (even if some NoSQL databases can support them; eg Mongo can use regex).

persistent/Database/Persist/Types/Base.hs
```diff
@@ -483,3 +488,36 @@ instance Exception OnlyUniqueException
...
+#endif
+
+#if __GLASGOW_HASKELL__ < 861
+instance Semigroup EscapedLikeText where
+  EscapedLikeText f <> EscapedLikeText g = EscapedLikeText (\c -> f c <> g c)
+
+instance Monoid EscapedLikeText where
+  mempty = EscapedLikeText (\_ -> T.pack "")
+  mappend = (<>)
+#endif
```
parsonsmatt: As above - rather than have the `deriving via` instance *and* the manual handwritten instances, let's just use the hand written instances for now. That'll remove 8 lines of code.

pythonissam:
@Vlix Sorry for replying so late.

>  but I'm pretty sure the part after LIKE in the current implementation would result in:
>
>> ... LIKE '%espea\re' ESCAPE '\'

Is this true? I created a test for this behavior and checked it out. Fortunately I couldn't confirm it.

```plain
"SELECT \"id\", \"name\", \"age\", \"color\" FROM \"Person\" WHERE (\"name\" LIKE ? ESCAPE ?); [PersistText \"%espea\\re\",PersistText \"\\\\\"]"
```

[Here's the link for the test branch](https://github.com/pythonissam/persistent/tree/test-for-vlix). Please check this with `stack build && stack exec persistent-test`.

Anyway, thank you for your report.

pythonissam:
@parsonsmatt 

Thanks for the review. I don't have time now, so I'm going to check them later today.

Vlix:
> @Vlix Sorry for replying so late.
> 
> > but I'm pretty sure the part after LIKE in the current implementation would result in:
> > > ... LIKE '%espea\re' ESCAPE ''
> 
> Is this true? I created a test for this behavior and checked it out. Fortunately I couldn't confirm it.
> 
> ```
> "SELECT \"id\", \"name\", \"age\", \"color\" FROM \"Person\" WHERE (\"name\" LIKE ? ESCAPE ?); [PersistText \"%espea\\re\",PersistText \"\\\\\"]"
> ```
> 
> [Here's the link for the test branch](https://github.com/pythonissam/persistent/tree/test-for-vlix). Please check this with `stack build && stack exec persistent-test`.
> 
> Anyway, thank you for your report.

I've just tested it in PostgreSQL (`psql`) directly. Will test it on your branch.

Vlix:
Ah, I see what went wrong. My example haskell code (which I've never run) was trying to imitate what I did in `psql` directly. I forgot that the `\` should be escaped in Haskell strings. If you change it into `let wut = "\\re"` it will fail.
(`\r` is a CARRIAGE RETURN, I believe..., I meant a backslash and then `"re"`)

Your test should look like this:
```haskell
let wut = "\\re"
ps' <- selectList [likeWithEscape '\\' PersonName ("%espea" <> toEscapedLikeText wut)] []
assertEmpty ps'
```

persistent/Database/Persist/Types/Base.hs
```diff
@@ -483,3 +488,36 @@ instance Exception OnlyUniqueException
...
+toEscapedLikeText = EscapedLikeText . escapeWildcards
+
+escapeWildcards :: Text -> Char -> T.Text
+escapeWildcards t c = T.foldr go "" t
+  where
+    isWildcard '%' = True
+    isWildcard '_' = True
+    isWildcard  _  = False
+    go x xs
+      | isWildcard x = c `T.cons` x `T.cons` xs
```
Vlix: This should fix it, I think.
```suggestion
      | isWildcard x || x == c = c `T.cons` x `T.cons` xs
```

Vlix:
@pythonissam 
Any chance you might continue working on this?

re-xyr:
Hi,

This PR looks good to me, but it seems that no one is currently working on it. I hope it can be merged soon. The lack of `LIKE` operator is a [long-existed problem](https://stackoverflow.com/questions/11048143/example-of-persistent-with-backend-specific-operator) to `persistent` users (including me). 

*PS One solution I found is to [use `esqueleto`](https://hackage.haskell.org/package/esqueleto-2.5.3/docs/Database-Esqueleto.html#v:ilike), but I personally think its eDSL syntax is too cumbersome, and I'm sure there are many people who dislike an SQL-like syntax.*

hugohp:
Hi , why was this PR closed and not merged? Do you intend to add any "like" functionality to persistent? Thanks

PR #846, opened 2019-01-14
Improve error message when marshalling fails

danbroooks:
Alters the error message that appears when marshalling DB values to include the name of the table. I have found it useful to know the table when this error rears its head with generic field names (`lastAccessed` for example), allowing me to go straight to the offending table.

---

I'm not sure which of these I need to complete for this change (bumping version number seems drastic, and I've not changed any public facing API):

After submitting your PR:


psibi:
@danbroooks Thanks for the PR! I like this. :-) I think some part of the code change in the PR are just formatting. Can we avoid that or keep that in a separate commit ( I would like to avoid it if the code isn't changed for easier review and easier git-blame for future references) ?

Also, can you give a sample example of the previous and the current error message with this PR ?

psibi:
@danbroooks Thanks!

psibi:
Also, a new version has been released to Hackage with the changes. Thanks!

PR #863, opened 2019-03-22
Addition of the "question mark" operators to persistent-postgresql

Vlix:
Since `postgresql-simple-0.6.1`, question marks can now be escaped in the SQL query (by doubling them), so now `persistent` can also use them.
Had to add a few new data constructors to have the resulting query be correct, and am looking into a way to make the other JSON operators work in `persistent` queries, but those will take some more thinking/experimenting time.

This does mean the `postgresql-simple` version's lower bound is upped to 0.6.1, so I've increased the `persistent-postgresql` version to `2.10.0`

Oh, and I also upped the stackage LTS to 13.13, I figured it wouldn't hurt. If I'm wrong, either the CI will break or I hope someone will inform me.


Vlix:
Ok, so nothing seems to make the CI happy. `persistent-postgresql` just needs the newest version of `postgresql-simple`, and that means having one of the newer LTS-es.
Why is the library still being tested against all these old LTS-es anyway? Is there a good reason to do so (and it not being tested against the newer ones?)

If there is, do tell me, because this seems rather limiting.

parsonsmatt:
@Vlix I'm currently updating CI to include newer versions of LTS. I believe that it was just overlooked in the past.

I'm going to cancel your current CI build to let the new CI finish faster. If you'd like to merge those changes in when they're done, that'd be great :smile:

Vlix:
I noticed I did everything on the master branch. Closing this PR and making a new one.

PR #867, opened 2019-04-02
Fix URL in README

momohatt:
The URL at the top of README doesn't seem to be working.


parsonsmatt:
Great catch, thanks! :smile:

PR #879, opened 2019-04-03
[WIP] Splitting off the tests

parsonsmatt:
This PR tracks splitting out the tests and removing CPP.


parsonsmatt:
Bah I've decided I don't like this approach.

Would be better if `persistent-test` were parameterized over the various things that are CPPed in, rather than having tons of duplicated code, and then `persistent-mongoDB`'s test suite just passes in the required stuff. Optional stuff is `Maybe` ofc.

PR #880, opened 2019-04-05
Remove CPP from test suite, take 2

parsonsmatt:
This PR removes all the CPP from the test suite.

I've pinged a few of you on a request; it's a big PR though and I won't be mad if you don't look at all of it. Self-review icoming.


.travis.yml
```diff
@@ -14,23 +14,13 @@ services:
...
-  - env: ARGS="--resolver lts-13" BACKEND=mongodb
-  - env: ARGS="--resolver lts-13" BACKEND=postgresql
-  - env: ARGS="--resolver lts-13" BACKEND=mysql
+  - env: ARGS="--resolver lts-6  --stack-yaml stack_lts-10.yaml"
+  - env: ARGS="--resolver lts-7  --stack-yaml stack_lts-10.yaml"
+  - env: ARGS="--resolver lts-9  --stack-yaml stack_lts-10.yaml"
+  - env: ARGS="--resolver lts-10 --stack-yaml stack_lts-10.yaml" 
+  - env: ARGS="--resolver lts-11"
+  - env: ARGS="--resolver lts-12"
+  - env: ARGS="--resolver lts-13"
```
parsonsmatt: Hooray for simpler CI!

persistent-mongoDB/persistent-mongoDB.cabal
```diff
@@ -46,6 +46,38 @@ library
...
+                   , hspec >= 1.3
+                   , text
+                   , persistent
+                   , bytestring
+                   , QuickCheck
+                   , transformers
+                   , mtl
+                   , containers
+                   , blaze-html
+                   , time
```
parsonsmatt: The deps are somewhat sloppy. They could be `weeder`ed now that we don't have so much CPP.

persistent-mongoDB/test/EmbedTestMongo.hs
```diff
@@ -0,0 +1,434 @@
...
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+module EmbedTestMongo (specs) where
```
parsonsmatt: Some of these tests were just really hard to unentangle, and this was one of them. Instead of detangling it I just copied it.

persistent-mysql/test/main.hs
```diff
@@ -0,0 +1,212 @@
...
+      , InsertDuplicateUpdate.duplicateMigrate
+      , MigrationIdempotencyTest.migration
+      , CustomPrimaryKeyReferenceTest.migration
+      , MigrationColumnLengthTest.migration
+      , TransactionLevelTest.migration
+      ]
+    PersistentTest.cleanDB
+
+  hspec $ do
+    RenameTest.specsWith db
```
parsonsmatt: All of the test suites are now parameterized by a database runner function: `type DbRunner backend m = ReaderT bakend m () -> IO ()`. This,combined with `mpsGeneric = True`, allows us to have most of these tests be generic over the backend.

persistent-postgresql/test/ArrayAggTest.hs
```diff
@@ -0,0 +1,67 @@
...
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NoMonomorphismRestriction #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module ArrayAggTest where
```
parsonsmatt: The tests that were enabled for a single backend were factored into the specific test suite.

persistent-qq/test/Spec.hs
```diff
@@ -0,0 +1,153 @@
+{-# LANGUAGE QuasiQuotes, GADTs, OverloadedStrings #-}
```
parsonsmatt: This test suite is annoying. We can't have a circular dependency, so in order to depend on `persistent-sqlite` to have a driver to run these tests against, we have to *not* depend on `persistent-test`

persistent-sqlite/test/Spec.hs
```diff
@@ -1,53 +0,0 @@
-{-# LANGUAGE EmptyDataDecls             #-}
```
parsonsmatt: Was incorproated into the other test suite.

parsonsmatt:
I'm gonna merge this on Wednesday if no one has any objections

Vlix:
Will try to take a look at it tonight.

Vlix:
Couldn't look in too much detail, and you've moved mountains, so from a somewhat intensive skimming, I'd say: "If it compiles, and tests, should probably be ok?"

parsonsmatt:
Yeah, the only thing that's really changed is that the MongoDB tests have totally different behavior on upsert. I'm tempted to call that a bug in the MongoDB library, but it may just be that they need to have totally different specifications factored out.

.travis.yml
parsonsmatt: They're all resolvers for distinct GHC versions. 7.10.3, 8.0.2, 8.2, 8.4, and 8.6 are represented here. More resolvers seemed redundant to me.

PR #884, opened 2019-04-09
JSON "question mark" operators for persistent-postgresql

Vlix:
Since postgresql-simple-0.6.1, question marks can now be escaped in the SQL query (by doubling them), so now persistent can also use them.
Had to add a few new data constructors to have the resulting query be correct, and am looking into a way to make the other JSON operators work in persistent queries, but those will take some more thinking/experimenting time.

This does mean the postgresql-simple version's lower bound is upped to 0.6.1, so I've increased the persistent-postgresql version to 2.10.0


persistent/Database/Persist/Types/Base.hs
```diff
@@ -293,6 +293,7 @@ data PersistValue = PersistText Text
                   | PersistList [PersistValue]
                   | PersistMap [(Text, PersistValue)]
                   | PersistObjectId ByteString -- ^ Intended especially for MongoDB backend
+                  | PersistArray [PersistValue] -- ^ Intended especially for PostgreSQL backend for text arrays
```
parsonsmatt: This addition makes me a little uneasy. I don't think the `PersistValue` type should have too many values that are unsupported by the majority of backends.

parsonsmatt: I think I'd rather have `PersistList` use the underlying Array type instead of adding a new type that throws a bunch of errors. Or we could have the `PersistArray` constructor fail over to the `PersistList` constructor.

persistent-postgresql/Database/Persist/Postgresql/JSON.hs
```diff
@@ -126,16 +131,192 @@ infix 4 @>., <@.
...
+-- "ab"  ?& ["a","b"] == False
+-- 1     ?& ["1"]     == False
+-- null  ?& ["null"]  == False
+-- true  ?& ["true"]  == False
+-- 31337 ?& []        == True
+-- true  ?& []        == True
+-- null  ?& []        == True
+-- @
+--
+-- @since 2.10.0
```
parsonsmatt: these docs are really really awesome

persistent/Database/Persist/Class/PersistEntity.hs
```diff
@@ -134,14 +136,20 @@ type family BackendSpecificFilter backend record
...
-    , filterValue  :: Either typ [typ] -- FIXME
+    , filterValue  :: FilterValue typ
     , filterFilter :: PersistFilter -- FIXME
     }
     | FilterAnd [Filter record] -- ^ convenient for internal use, not needed for the API
     | FilterOr  [Filter record]
     | BackendFilter
           (BackendSpecificFilter (PersistEntityBackend record) record)
 
+-- | Value to filter with. Highly dependant on the type of filter used.
```
parsonsmatt: Can you add the `@since` annotation for this?

persistent/Database/Persist/Types/Base.hs
Vlix: This might break people's programs, though. If suddenly `persistent` wants to read arrays, instead of the JSON-y implementation. Maybe a different way of handling this would be consolidating `PersistObjectId` and `PersistArray` into an `PersistBackendSpecific ByteString`, which can be used to input any kind of raw SQL?

parsonsmatt: Yeah, altering the underlying datatype is definitely going to be a *big* breaking change for that. For now, I'm happy with treating this as `PersistList` in the non-Postgres backends and adding array functionality for this to Postgres.

Vlix:
Ah, seems it's been a long day for me, haha

persistent/Database/Persist/Types/Base.hs
Vlix: done

parsonsmatt:
I'm going to squash+merge the commit in :) awesome PR, thanks!

PR #886, opened 2019-04-12
Backend compatible

Vlix:
I've factored out what I've found to be redundant constraints in some SQL functions.
Most notably all instances of `IsSqlBackend` have been replaced or removed and `BackendCompatible SqlBackend` has been integrated where possible.

I've noticed a ton of redundant (at least seemingly) classes/types/etc.
After this refactor at least the following aren't noticeably used anywhere:

* SqlReadBackend
* SqlWriteBackend
* IsPersistBackend
* IsSqlBackend

---

If anyone can look this over, to see if I didn't interfere with the API, or some guarantee that's lost by fiddling with constraints, that'd be great.


parsonsmatt:
`SqlReadBackend` and `SqlWriteBackend` are used by clients of the library. I know they're used extensively by Freckle, so as long as those types are still working good after these changes then we should be good to go.

Vlix:
> `SqlReadBackend` and `SqlWriteBackend` are used by clients of the library. I know they're used extensively by Freckle, so as long as those types are still working good well after these changes then we should be good to go.

If someone can check with whatever uses it, that'd be great, because I have no idea how they'd be used looking at this library alone.

bitemyapp:
@akurilin I suggest building this against your org repo. I can't do it for you.

parsonsmatt:
I'm gonna write some tests for the feature.

Vlix:
Ok, well if the Read/Write types have a function, that still leaves the `IsPersistBackend` (and consequently `IsSqlBackend`) to not be used anywhere. If it is decided they are actually redundant, I can remove the definitions and instances of them.

Seeing as this change does shift almost all functionality to `BackendCompatible`, instead of `HasPersistBackend`, would that break people's programs?
It'd probably just give type errors when they update their `persistent` dependency, right?

parsonsmatt:
The `BackendCompatible` stuff is used so that `SqlReadBackend` can be used for `persistent` entities that are defined for `SqlBackend` but *only* in read queries. Note that `SqlReadBackend` doesn't have any instances for the `Write` classes, so it can't do any updates to the DB. This is used to give certain parts of the program read-only access to the database. I wrote that class because the `IsSqlBackend` stuff was too restrictive for [`persistent-typed-db`](https://hackage.haskell.org/package/persistent-typed-db), where I had `PersistEntityBackend record ~ SqlFor SomeDb`.

I'd *love* it if the new changes were mostly "drop in." I think changing the functions to return a concrete `Sqlbackend` and then allowing clients to apply those functions manually will probably be a good thing - should make type inference a lot better and error messages easier to follow.

parsonsmatt:
I think it may be a good idea to deprecate the `IsSqlBackend`-related stuff so we can give good deprecation warnings for an upgrade/migration path. ideally we only have one mechanism for saying "This backend is compatible with this other backend."

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -265,9 +263,9 @@ wrapConnectionInfo connInfo conn logFunc = do
 -- that all log messages are discarded.
 --
 -- @since 1.1.4
-runSqlite :: (MonadUnliftIO m, IsSqlBackend backend)
+runSqlite :: (MonadUnliftIO m)
           => Text -- ^ connection string
-          -> ReaderT backend (NoLoggingT (ResourceT m)) a -- ^ database action
+          -> ReaderT SqlBackend (NoLoggingT (ResourceT m)) a -- ^ database action
```
parsonsmatt: This is another thing that should be easy to generalize on our side.

persistent/Database/Persist/Class/PersistQuery.hs
```diff
@@ -65,7 +65,7 @@ class (PersistQueryRead backend, PersistStoreWrite backend) => PersistQueryWrite
 -- | Get all records matching the given criterion in the specified order.
 -- Returns also the identifiers.
 selectSource
-       :: (PersistQueryRead (BaseBackend backend), MonadResource m, PersistEntity record, PersistEntityBackend record ~ BaseBackend (BaseBackend backend), MonadReader backend m, HasPersistBackend backend)
+       :: (PersistQueryRead backend, MonadResource m, PersistEntity record, PersistRecordBackend record backend, MonadReader backend m)
```
parsonsmatt: This signature is so much nicer :heart:

persistent/Database/Persist/Sql/Run.hs
```diff
@@ -113,7 +113,7 @@ withSqlPool mkConn connCount f = withUnliftIO $ \u -> bracket
     (unliftIO u . f)
 
 createSqlPool
-    :: (MonadLogger m, MonadUnliftIO m, IsSqlBackend backend)
+    :: (MonadLogger m, MonadUnliftIO m, BackendCompatible SqlBackend backend)
```
parsonsmatt: Hm, if these functions can return a `BackendCompatible SqlBackend backend => backend` then shouldn't the other `{with,create}{Postgresql,MySQL,etc}Pool` functions also be able to create that?

AKurilin:
> @AKurilin I suggest building this against your org repo. I can't do it for you. I remember y'all using those types too but I have no idea where things are at now.

@jdreaver is taking a look 👍

parsonsmatt:
I added some tests in #887 though they're not super complete.

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -195,18 +194,17 @@ retryOnBusy action =
 -- | Wait until some noop action on the database does not return an 'Sqlite.ErrorBusy'. See 'retryOnBusy'.
 --
 -- @since 2.9.3
-waitForDatabase :: (MonadUnliftIO m, MonadLogger m, IsSqlBackend backend, BackendCompatible SqlBackend backend) => ReaderT backend m ()
+waitForDatabase :: (MonadUnliftIO m, MonadLogger m) => ReaderT SqlBackend m ()
```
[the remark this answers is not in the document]
Vlix: Ah, you mean change it to this?
```haskell
waitForDatabase :: (MonadUnliftIO m, MonadLogger m, BackendCompatible SqlBackend backend)
                => ReaderT backend m () 
```

And do `projectBackend` in the function body?

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -195,18 +194,17 @@ retryOnBusy action =
...
 -- @since 2.6.2
-wrapConnectionInfo :: (IsSqlBackend backend)
-                  => SqliteConnectionInfo
-                  -> Sqlite.Connection
-                  -> LogFunc
-                  -> IO backend
+wrapConnectionInfo :: SqliteConnectionInfo
+                   -> Sqlite.Connection
+                   -> LogFunc
+                   -> IO SqlBackend
```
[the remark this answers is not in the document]
Vlix: Seems fair.

persistent-sqlite/Database/Persist/Sqlite.hs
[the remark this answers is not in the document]
parsonsmatt: Yeah

persistent-sqlite/Database/Persist/Sqlite.hs
Vlix: Well, no, actually :stuck_out_tongue: 

The entire point it that the `backend/SqlBackend` here, is from the `open'` function, which creates the `SqlBackend` in the `withSqliteConn`. Making a `SqlBackend` into a `BackendCompatible SqlBackend backend => backend` would imply creating a `backend` from a `SqlBackend`. That being annoying and undesirable was the entire point.
Remember, this is Sqlite specific, so it's gonna have to work with a `SqlBackend`.

The `RawSqlite` PR should work now afaict, though. Then the `SqlBackend` here would turn into a `RawSqlite SqlBackend` which is what `merijn` was trying to accomplish, I think.

persistent/Database/Persist/Sql/Run.hs
Vlix: Well, no, since those functions are actual implementations, they will create `SqlBackend`s.

The only thing I can think of that wouldn't just imply using `IsSqlBackend`, would be an extra argument somewhere in which the user of the function can give the following function which will be called before creating the `Pool`:
`BackendCompatible SqlBackend backend => SqlBackend -> backend`
Maybe with a `MonadIO m` so they can use IO to create their own `backend`.

persistent-sqlite/Database/Persist/Sqlite.hs
parsonsmatt: That's not quite right - `BackendCompatible SqlBackend backend` implies that you can create a `SqlBackend` from a `backend`. `runSqlite` would be able to run an action that uses `SqlBackend`, or `SqlReadBackend`, or `SqlFor SomeDb`, or `RawSqlite SqlBackend` (provided that was given an `instance BackendCompatible SqlBackend (RawSqlite SqlBackend)`).

The problem with `IsSqlBackend` is that it requires an *isomorphism* - that you can go from a `SqlBackend` to your type. `RawSqlite SqlBackend` has more than a mere `SqlBackend`, so there's no `SqlBackend -> RawSqlite SqlBackend` you can use.

parsonsmatt: We can try the polymorphic `BackendCompatbile SqlBackend backend => backend` stuff and if it presents problems with the raw SQLite stuff then we can revert those functions to concrete types.

It's not a deal breaker either way - requiring `ReaderT SqlBackend ...` just means that library users call `local SqlReadBackend` to map the types how they want instead of the library doing it automatically.

Vlix: I'll try it one more time, maybe I'm just derping with the type system.

Vlix:
Ok, so here's the dealio:

This is the chain:
```haskell
wrapConnectionInfo :: SqliteConnectionInfo -> Sqlite.Connection -> LogFunc -> IO SqlBackend
wrapConnectionInfo = {the entire SqlBackend is constructed here}

open' :: SqliteConnectionInfo -> LogFunc -> IO SqlBackend
open' connInfo logFunc = do
    conn <- Sqlite.open $ _sqlConnectionStr connInfo
    wrapConnectionInfo connInfo conn logFunc `E.onException` Sqlite.close conn

withSqliteConnInfo :: (MonadUnliftIO m, MonadLogger m)
                   => SqliteConnectionInfo -> (SqlBackend -> m a) -> m a
withSqliteConnInfo = withSqlConn . open'

withSqlConn
    :: (MonadUnliftIO m, MonadLogger m, BackendCompatible SqlBackend backend)
    => (LogFunc -> IO backend) -> (backend -> m a) -> m a
withSqlConn open f = do
    logFunc <- askLogFunc
    withRunInIO $ \run -> bracket
      (open logFunc)
      close'
      (run . f)

runSqlite :: (MonadUnliftIO m)
          => Text -- ^ connection string
          -> ReaderT SqlBackend (NoLoggingT (ResourceT m)) a -- ^ database action
          -> m a
runSqlite connstr action = runResourceT . runNoLoggingT $
    withSqliteConn connstr $ runSqlConn action
```

`runSqlConn` takes the action and will run it with whatever the `open` argument to `withSqlConn` is. In this case, that's the `SqlBackend` from `open'`. (`withSqliteConn` is just `withSqliteConnInfo`, but the Text is first turned into a `SqliteConnectionInfo`)
I honestly don't see a way to make that `SqlBackend` into anything else? I don't see a way to get around the fact that the database action `ReaderT` will receive a `SqlBackend`. What one does with it is up to them, I guess.

parsonsmatt:
Okay, let's land this PR, then get the SQLite one working, and I'll give these `open`-etc functions a try to see if I can get them generalized without borking any expected functionality.

parsonsmatt:
I tried to test on my work codebase but I need a new version of `persistent-typed-db` for that to work out :) May get that done next week if @jdreaver hasn't reported back.

parsonsmatt:
@Vlix There are some merge conflicts from your other PR. Let's get those resolved and merge this in. Freckle and folks can test on `master` and if it's not working then we can figure out how to work around it there.

merijn:
> I think it may be a good idea to deprecate the `IsSqlBackend`-related stuff so we can give good deprecation warnings for an upgrade/migration path. ideally we only have one mechanism for saying "This backend is compatible with this other backend."

Agreed and I'm very happy the (more flexible) `BackendCompatible` ends up the final winner with this PR :)

@Vlix once this PR is merged I'll try and make some time to revisit the SQLite bits you mentioned, since those needed changing in my other PR and since it's grossly out of date by now I need to revisit it anyway.

parsonsmatt:
I'm going to get this PR finished up

PR #887, opened 2019-04-12
This PR adds read/write tests to check functionality on the read/write backens.

parsonsmatt:

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Sql/Types/Internal.hs
+++ b/persistent/Database/Persist/Sql/Types/Internal.hs
@@ -157,6 +157,8 @@ instance IsPersistBackend SqlBackend where
     mkPersistBackend = id
 
 -- | An SQL backend which can only handle read queries
+--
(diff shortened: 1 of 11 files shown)


persistent/Database/Persist/Sql/Types/Internal.hs
```diff
@@ -9,8 +9,8 @@
 module Database.Persist.Sql.Types.Internal
     ( HasPersistBackend (..)
     , IsPersistBackend (..)
-    , SqlReadBackend (unSqlReadBackend)
-    , SqlWriteBackend (unSqlWriteBackend)
+    , SqlReadBackend (..)
+    , SqlWriteBackend (..)
```
parsonsmatt: I have no idea why the constructors were hidden. Presumably you were only supposed to use the `IsSqlBackend` instance, but that lets you construct them, so :man_shrugging:

PR #888, opened 2019-04-12
Cleanup of pragmas, imports and dependencies

Vlix:
Just cleaned up tons of redundant imports, pragmas and cabal dependencies.


persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -21,68 +19,62 @@ module Database.Persist.MySQL
      -- * @ON DUPLICATE KEY UPDATE@ Functionality
     , insertOnDuplicateKeyUpdate
     , insertManyOnDuplicateKeyUpdate
-#if MIN_VERSION_base(4,7,0)
     , HandleUpdateCollision
     , pattern SomeField
-#elif MIN_VERSION_base(4,9,0)
-    , HandleUpdateCollision(SomeField)
-#endif
     , SomeField
```
Vlix: Not sure what the CPP above should do? Just kept the most stuff in there to be sure.

parsonsmatt: My guess is that view patterns are only supported on GHC 8 or above? In any case we have dropped support for pre-GHC 8 so we should be fine there.

persistent-sqlite/test/main.hs
```diff
@@ -220,3 +207,12 @@ main = do
...
         insertMany_ $ replicate 1000 (Test $ read "2014-11-30 05:15:25.123")
+
+    it "afterException" $ asIO $ runSqliteInfo (mkSqliteConnectionInfo ":memory:") $ do
+        runMigration testMigrate
+        let catcher :: forall m. Monad m => SomeException -> m ()
+            catcher _ = return ()
+        _ <- insert $ Person "A" 0 Nothing
+        _ <- insert_ (Person "A" 1 Nothing) `catch` catcher
+        _ <- insert $ Person "B" 0 Nothing
+        return ()
```
parsonsmatt: Thanks for catching that! :smile:

persistent-template/Database/Persist/TH.hs
```diff
@@ -530,19 +511,15 @@ uniqueTypeDec :: MkPersistSettings -> EntityDef -> Dec
...
             Nothing
-#endif
             (map (mkUnique mps t) $ entityUniques t)
             (derivClause $ entityUniques t)
   where
     derivClause [] = []
 #if MIN_VERSION_template_haskell(2,12,0)
     derivClause _  = [DerivClause Nothing [ConT ''Show]]
-#elif MIN_VERSION_template_haskell(2,11,0)
-    derivClause _  = [ConT ''Show]
```
parsonsmatt: Are we only supporting `template-haskell >= 2.12` now?

persistent-template/persistent-template.cabal
```diff
@@ -8,33 +8,33 @@ synopsis:        Type-safe, non-relational, multi-backend persistence.
...
+                   , http-api-data            >= 0.3.7
+                   , monad-control            >= 1.0       && < 1.1
                    , monad-logger
-                   , unordered-containers
-                   , tagged
                    , path-pieces
-                   , http-api-data            >= 0.2
-                   , ghc-prim
+                   , tagged
+                   , template-haskell
```
parsonsmatt: If we are going to drop CPP for features in older libraries then I think we should put lower bounds on them to exclude building with these lower versions

persistent-template/persistent-template.cabal
```diff
@@ -8,33 +8,33 @@ synopsis:        Type-safe, non-relational, multi-backend persistence.
...
+                   , text                     >= 1.2
+                   , transformers             >= 0.5       && < 0.6
+                   , unordered-containers
     exposed-modules: Database.Persist.TH
     ghc-options:     -Wall
-    if impl(ghc >= 7.4)
-       cpp-options: -DGHC_7_4
+    cpp-options:     -DGHC_7_4
```
parsonsmatt: mmm, since we only support GHC >= 8 at this point, I think we can just omit this whole thing.

persistent-template/Database/Persist/TH.hs
```diff
@@ -167,9 +153,6 @@ persistFileWith ps fp = persistManyFileWith ps [fp]
 -- @since 2.5.4
 persistManyFileWith :: PersistSettings -> [FilePath] -> Q Exp
 persistManyFileWith ps fps = do
-#ifdef GHC_7_4
-    mapM_ qAddDependentFile fps
-#endif
```
parsonsmatt: I think we want to keep this line. If I read the original code right, this line of code only gets included if the GHC version is `>= 7.4`. We're definitely doing that, so we want to keep this line.

persistent-template/Database/Persist/TH.hs
parsonsmatt: Added a comment on the cabal file - we should add lower bounds if we're dropping CPP support for versions of the libraries

persistent-template/Database/Persist/TH.hs
```diff
@@ -1116,6 +1082,7 @@ mkEntity entMap mps t = do
 
 mkUniqueKeyInstances :: MkPersistSettings -> EntityDef -> Q [Dec]
 mkUniqueKeyInstances mps t = do
+    -- FIXME: isExtEnabled breaks the benchmark
```
parsonsmatt: It should be possible to just add the flag to the benchmark.

Benchmarks should be built in CI - oops!

persistent-test/src/CompositeTest.hs
```diff
@@ -1,27 +1,15 @@
...
-{-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE NoMonomorphismRestriction #-}
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE QuasiQuotes #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-} -- FIXME
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
```
parsonsmatt: this is a great warning though!

persistent-test/src/CompositeTest.hs
```diff
@@ -109,7 +97,7 @@ specsWith runDb = describe "composite" $
       kp2 <- insert p2
       xs <- selectList [TestParentId <-. [kp1,kp2]] []
       length xs @== 2
-      let [e1@(Entity newkp1 newp1),e2@(Entity newkp2 newp2)] = xs
+      let [(Entity newkp1 newp1),(Entity newkp2 newp2)] = xs
```
parsonsmatt:
```suggestion
      [(Entity newkp1 newp1),(Entity newkp2 newp2)] <- pure xs
```

This uses hte `MonadFail` machinery instead of a partial pattern match.

persistent-test/src/EmptyEntityTest.hs
```diff
@@ -1,16 +1,6 @@
...
-{-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE GADTs #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE QuasiQuotes #-}
-{-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-unused-top-binds #-}
```
parsonsmatt: This is probably a good flag to put in the cabal file for this package.

persistent-test/src/EquivalentTypeTest.hs
```diff
@@ -1,25 +1,14 @@
-{-# OPTIONS_GHC -fno-warn-unused-binds #-}
-{-# LANGUAGE UndecidableInstances #-}
-{-# LANGUAGE CPP #-}
-{-# LANGUAGE EmptyDataDecls #-}
-{-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE GADTs #-}
+-- {-# LANGUAGE CPP #-}
```
parsonsmatt:
```suggestion
```

persistent/Database/Persist/Class/PersistUnique.hs
```diff
@@ -424,10 +424,11 @@ onlyUnique
     => record -> ReaderT backend m (Unique record)
 onlyUnique = pure . onlyUniqueP
 
-onlyUniqueEither
+-- Was this supposed to be exported for use by users?
```
parsonsmatt: Might as well delete it.

persistent/Database/Persist/Class/PersistUnique.hs
```diff
@@ -492,10 +493,10 @@ getByValueUniques uniqs =
             Just z -> return $ Just z
 
 -- TODO: expose this to users
-recordName
+_recordName
     :: (PersistEntity record)
     => record -> Text
-recordName = unHaskellName . entityHaskell . entityDef . Just
+_recordName = unHaskellName . entityHaskell . entityDef . Just
```
parsonsmatt: If this were exposed to users it would be from the module that defines `EntityDef` I'd think

persistent-mysql/Database/Persist/MySQL.hs
Vlix: I've never seen the `pattern` export before. Not sure what that does. :/

persistent-sqlite/test/main.hs
Vlix: np :yum:

persistent-template/Database/Persist/TH.hs
Vlix: `LTS 9.21` has `template-haskell-2.11.1.0`, so I've made the lower bound `>= 2.11`

persistent-mysql/Database/Persist/MySQL.hs
parsonsmatt: It exports a pattern synonym. The `pattern` keyword is necessary to distinguish it from the type and data constructor namespace.

persistent-template/persistent-template.cabal
Vlix: will do (incidentally, missed this cabal with purging dependencies. Got rid of `tagged` and `ghc-prim`, thnx.

persistent-template/Database/Persist/TH.hs
Vlix: Ah, good catch, probably misread it as `if ghc == 7.4`

parsonsmatt: CPP should be bant

persistent-test/src/CompositeTest.hs
Vlix: I added it, because it's used a ton with the idea that, if it doesn't match, the test would fail (because of the partial match), so no reason to have all those warnings flood the screen, right?
I can remove it if you prefer.

persistent-template/Database/Persist/TH.hs
Vlix: I take it this doesn't require any action on my part?

persistent-test/src/CompositeTest.hs
Vlix: Sure, fine by me. I just removed the `e1` and `e2` because they weren't used anywhere P:

persistent-test/src/EmptyEntityTest.hs
Vlix: Can't seem to disable these with `-Wall`... will just leave them in.

persistent-test/src/EquivalentTypeTest.hs
Vlix: is removed in coming commit.

persistent/Database/Persist/Class/PersistUnique.hs
Vlix: done

persistent/Database/Persist/Class/PersistUnique.hs
Vlix: That would mean it'd go into `Database.Persist.Types.Base`, but `entityDef` is defined in `Database.Persist.Class.PersistEntity`, which depends on `Database.Persist.Types.Base`... Should I just throw it into `Database.Persist.Class.PersistEntity`?

persistent-template/Database/Persist/TH.hs
Vlix: bant?

persistent-template/persistent-template.cabal
Vlix: done

Vlix:
I've done pretty much all requested changes. If the `recordName` thing is ok, this should be done now, no?

persistent/Database/Persist/Class/PersistUnique.hs
parsonsmatt: :thinking: Yeah that makes the most sense. Thanks :)

persistent-template/Database/Persist/TH.hs
parsonsmatt: It's a dumb way of saying "banned." I really dislike how difficult it makes ot understand and use the codebase.

persistent-template/Database/Persist/TH.hs
parsonsmatt: Correct :)

persistent-test/src/CompositeTest.hs
parsonsmatt: It's fine to have this warning flag still in. A later refactoring could remove the need for it.

PR #892, opened 2019-04-15
Tidying up a few things

### LINKED ISSUE #890: Builds benchmarks in CI
Currently we just do `stack test`. We should do `stack test --bench --no-run-benchmarks` to ensure that changes don't break benchmarks.

### LINKED ISSUE #891: How to Test Postgres is Unclear
Hello,

I have been unsuccessful in trying to run the tests for `persistent-postgresql`, even after reading through `development.md`.

The following command does work:

```
stack test persistent-sqlite
```

However, this fails `stack test persistent-postgresql` fails with the following error message:

```
persistent-postgresql-2.10.0: test (suite: test)

)

persistent-postgresql-2.10.0: Test suite test failed
Test suite failure for package persistent-postgresql-2.10.0
    test:  exited with: ExitFailure 1
Logs printed to console
```

I imagine that the reason that this is failing is because I have failed to setup the database correctly. There are two files here that look like they could be relevant: `persistent-postgresql/settings.yaml` and `persistent-postgresql/test-settings.sh`. I tried setting a database called `test` owned by a postgres role called `test` that has the `SUPERUSER` attribute, can login via md5 password authentication, and and changed `persistent-postgresql/settings.yaml` to match `test-settings.sh`, but the same error occured.

I would love to add specific instructions for running postgres tests, but can't figure out what to do here.

A separate issue, the command `stack test persistent-test --flag persistent-test:postgresql --exec persistent-test` fails with:

```
Invalid flag specification:
- Package 'persistent-test' does not define the following flags (specified on command line):
  postgresql
- Flags defined by package 'persistent-test':

```

parsonsmatt:
* Add support for building with cabal new-*
* Build benchmarks in CI (fixes #890)
* Add docs for testing persistent-postgresql (fixes #891)
* Add a few badges
* Remove confusing files
* Remove references to running persistent-test

@StevenXL can you let me know if these instructions match up with what you did to get started on Postgres?


persistent-mysql/README.md
```diff
@@ -0,0 +1,10 @@
+# `persistet-mysql`
```
psibi: Type: persistent

StevenXL:
@parsonsmatt This looks great! I had to supply a password regardless of how I set up `pg_hba.conf`, but frankly I think that's out of scope. Thank you!

parsonsmatt:
Thanks :)

persistent-mysql/README.md
parsonsmatt: :+1: good catch thanks!

PR #893, opened 2019-04-15
Vlix/backend compatible

parsonsmatt:
A continuation of #886

### DIFF
--- a/persistent/Database/Persist/Class/PersistQuery.hs
+++ b/persistent/Database/Persist/Class/PersistQuery.hs
@@ -22,13 +22,13 @@ class (PersistCore backend, PersistStoreRead backend) => PersistQueryRead backen
     -- | Get all records matching the given criterion in the specified order.
     -- Returns also the identifiers.
     selectSourceRes
-           :: (PersistRecordBackend record backend, MonadIO m1, MonadIO m2)
+           :: (PersistEntity record, PersistRecordBackend record backend, MonadIO m1, MonadIO m2)
            => [Filter record]
            -> [SelectOpt record]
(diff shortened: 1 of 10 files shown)


persistent/Database/Persist/Class/PersistQuery.hs
```diff
@@ -22,13 +22,13 @@ class (PersistCore backend, PersistStoreRead backend) => PersistQueryRead backen
     -- | Get all records matching the given criterion in the specified order.
     -- Returns also the identifiers.
     selectSourceRes
-           :: (PersistRecordBackend record backend, MonadIO m1, MonadIO m2)
+           :: (PersistEntity record, PersistRecordBackend record backend, MonadIO m1, MonadIO m2)
```
Vlix: Isn't this redundant? `PersistRecordBackend record backend` is a synonym for `PersistEntity Record` and something else iirc

persistent/Database/Persist/Class/PersistQuery.hs
```diff
@@ -22,13 +22,13 @@ class (PersistCore backend, PersistStoreRead backend) => PersistQueryRead backen
...
            => [Filter record]
            -> [SelectOpt record]
            -> ReaderT backend m1 (Acquire (ConduitM () (Entity record) m2 ()))
 
     -- | Get just the first record for the criterion.
-    selectFirst :: (MonadIO m, PersistRecordBackend record backend)
+    selectFirst :: (MonadIO m, PersistEntity record, PersistRecordBackend record backend)
```
Vlix: Same here

parsonsmatt: Good catch! I've pruned those constraints.

PR #895, opened 2019-04-16
Fix upsert SQL to take a list of unique defs

parsonsmatt:

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -59,6 +59,8 @@ import Data.Int (Int64)
 import qualified Data.IntMap as I
 import Data.IORef
 import Data.List (find, sort, groupBy)
+import Data.List.NonEmpty (NonEmpty)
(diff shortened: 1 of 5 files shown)


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -315,7 +317,7 @@ upsertSql' ent updateVal = T.concat
                            , " RETURNING ??"
                            ]
     where
-      wher = T.intercalate " AND " $ map singleCondition $ entityUniques ent
+      wher = T.intercalate " AND " $ map singleCondition $ NEL.toList uniqs
```
parsonsmatt: The issue that prompted this change is documented in #856 . Basically, the type `UpsertBy` in the tests has two unique defs. The old implementation summoned up all the uniquedefs, and this was OK since the `upsert` function would blow up at runtime before generating this unless the entity had exactly *one* unique key.

This change allows the user to provide the unique keys to upsert by. This enables backend-specific upsert-by which should be faster and safer.

Issue #896, opened 2019-04-16
Bug Report: rawSql and Update clause expand columns


StevenXL:
Hi everyone. I have a [failing test](https://github.com/StevenXL/persistent/blob/efbfcfe02986ef9a22041dfdddf0cd3902552c4e/persistent-test/src/RawSqlTest.hs#L13) that should be working.

This bug was originally filed as an [issue on the Yesod Repo](https://github.com/yesodweb/yesod/issues/582). I'm going to close it down there and move it to this repo. 

To summarize, using `rawSql "UPDATE person SET age = 55" []` leads to an error. In the SQLite test suite (`stack test persistent-sqlite`), the error manifests itself as:

```
  src/RawSqlTest.hs:13:3:
  1) rawSql Can UPDATE
       uncaught exception: SqliteException
       SQLite3 returned ErrorError while attempting to perform prepare "UPDATE person SET age = 55\"Person\".\"id\", \"Person\".\"name\", \"Person\".\"age\", \"Person\".\"color\"": near ""Person"": syntax error
```

parsonsmatt:
For SQL commands that don't have return values, then you should use `rawExecute`. This behavior is very strange though.

StevenXL:
@parsonsmatt that sounds good enough to me; I'd be fine with just a small update to the docs. (See #897). I'm closing this down but feel free to reopen if you think it is worth tracking down.

parsonsmatt:
I'd like to know *why* this is happening more clearly and give a better error message :smile:

PR #897, opened 2019-04-17
Recommend `rawExecute` when return value is not expected.

StevenXL:
Easy way to avoid most of the problems associated with #896.


parsonsmatt:
Thanks!

Issue #906, opened 2019-05-13
`Primary` doesn't trigger migrations as it should


parsonsmatt:
- Removing a `Primary` declaration does not cause the uniqueness constraint or primary key to be dropped
    ```
    <<<
    User
      name Text
      Primary name
    ====
    User
      name Text
    >>>
   ```
  This should result in a migration that drops the primary key. It does not.

- Adding a `Primary` declaration to an existing non-unique column isn't considered an unsafe migration by any of the current backends. It fails on sqlite at runtime.

hdgarrood:
Another similar issue I just came across: adding a `Primary` declaration to an existing table does not cause primary keys on the table to be updated, at least not on Postgres. Example:

```haskell
-- app/Support.hs
module Support where

import Database.Persist
import Database.Persist.TH

share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
User
    name String
    age Int
    deriving Show

Color
    hexcode String
    deriving Show
|]

-- app/Schema1.hs
module Schema1 where

import Database.Persist
import Database.Persist.TH

import Support

share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
ColorPreference
    userId UserId
    colorId ColorId
    deriving Show
|]

-- app/Schema2.hs
module Schema2 where

import Database.Persist
import Database.Persist.TH

import Support

share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
ColorPreference
    userId UserId
    colorId ColorId
    Primary userId colorId
    deriving Show
|]

-- app/Main.hs
module Main where

import Control.Monad.Logger (runStdoutLoggingT)
import Control.Monad.Trans.Reader (runReaderT)
import Data.String (fromString)
import Database.Persist.Postgresql
import Database.Persist

import qualified Support
import qualified Schema1
import qualified Schema2

connStr =
  fromString "dbname=persist-postgres-pkey-testing"

main =
  runStdoutLoggingT
  $ withPostgresqlConn connStr
  $ runReaderT
  $ do
      logInfoN "Support"
      runMigration Support.migrateAll
      logInfoN "Schema1"
      runMigration Schema1.migrateAll
      logInfoN "Schema2"
      runMigration Schema2.migrateAll
```

Output:
```
[Info] Support
[...]

PR #907, opened 2019-05-13
Primary Key Migration Woes

parsonsmatt:
This PR tracks #906 

Test suite demonstrates problems

---

After submitting your PR:


persistent-test/src/MigrationTest.hs
```diff
@@ -30,21 +31,43 @@ Source1 sql=source
...
-      liftIO $ again @?= []
+        -- Failing test case for #735.  Foreign-key checking, switched on in
+        -- version 2.6.1, caused persistent-sqlite to generate a `references`
+        -- constraint in a *temporary* table during migration, which fails.
+        _ <- runMigration migrationMigrate
+        _ <- runMigration migrationAddCol
+        again <- getMigration migrationAddCol
+        liftIO $ again @?= []
+    describe "Add Primary key constraint on raw table" $ do
+        it "should not be considered safe" $ runDb $ do
```
parsonsmatt: Adding a primary key to a column is definitely *not* safe, because it's like adding a unique key to a column.

persistent-test/src/MigrationTest.hs
```diff
@@ -30,21 +31,43 @@ Source1 sql=source
...
+            liftIO $ migration
+                `shouldSatisfy`
+                    (\cm -> True `elem` map fst cm)
+
+        it "works" $ runDb $ do
+            rawExecute "CREATE TABLE from_raw_migration (name VARCHAR NOT NULL, age INT NOT NULL)" []
+            Right migration <- parseMigration addPrimKey
+            () <- runMigration addPrimKey
+            again <- getMigration addPrimKey
+            liftIO $ again @?= []
```
parsonsmatt: Currently fails on `sqlite` with:

```
      src/MigrationTest.hs:68:9: 
      2) Migration, Add Primary key constraint on raw table, works
           uncaught exception: SqliteException
           SQLite3 returned ErrorError while attempting to perform prepare "INSERT INTO \"from_raw_migration_backup\"(\"id\",\"name\",\"age\") SELECT \"id\",\"name\",\"age\" FROM \"from_raw_migration\"": table from_raw_migration_backup has no column named id
    
      To rerun use: --match "/Migration/Add Primary key constraint on raw table/works/"
```

Issue #909, opened 2019-05-22
SQLite deleteCascade does not handle all foreign keys


rvl:
Hi, I'm trying to use persistent-sqlite and `deleteCascade` with foreign keys referencing non-primary-key columns.

Here is my [example program](https://gist.github.com/rvl/9199d8c15afbfbb326a5900249f315be) based on the reproduction template.

When I run the example, `deleteCascade` correctly deletes rows of tables with foreign keys that reference the `username` PK column. But it does not delete the rows of tables with foreign keys that reference the non-PK `name` column.

Is this something that should work with persistent-sqlite? My current workaround is to replace `deleteCascade` with multiple `deleteWhere` statements to take care of the related entities.

parsonsmatt:
Thanks for the reproduction! This strikes me as something that *should* work, certainly.

parsonsmatt:
OK, so the issue is in `mkDeleteCascade`:

```haskell
-- | Generate a 'DeleteCascade' instance for the given @EntityDef@s.
mkDeleteCascade :: MkPersistSettings -> [EntityDef] -> Q [Dec]
mkDeleteCascade mps defs = do
    let deps = concatMap getDeps defs
    mapM (go deps) defs
  where
    getDeps :: EntityDef -> [Dep]
    getDeps def =
        concatMap getDeps' $ entityFields $ fixEntityDef def
      where
        getDeps' :: FieldDef -> [Dep]
        getDeps' field@FieldDef {..} =
            case foreignReference field of
                Just name ->
                     return Dep
                        { depTarget = name
                        , depSourceTable = entityHaskell def
                        , depSourceField = fieldHaskell
                        , depSourceNull  = nullable fieldAttrs
                        }
                Nothing -> []
    go :: [Dep] -> EntityDef -> Q Dec
    go allDeps EntityDef{entityHaskell = name} = do
        let deps = filter (\x -> depTarget x == name) allDeps
        key <- newName "key"
        let del = VarE 'delete
        let dcw = VarE 'deleteCascadeWhere
        just <- [|Just|]
        filt <- [|Filter|]
        eq <- [|Eq|]
        value <- [|FilterValue|]
        let mkStmt :: Dep -> Stmt
            mkStmt dep = NoBindS
                $ dcw `AppE`
                  ListE
                    [ filt `AppE` ConE filtName
                           `AppE` (value `AppE` val (depSourceNull dep))
                           `AppE` eq
                    ]
              where
[...]

PR #912, opened 2019-05-22
Delete Cascade

parsonsmatt:

After submitting your PR:

#909

### DIFF
--- a/persistent-test/src/CompositeTest.hs
+++ b/persistent-test/src/CompositeTest.hs
@@ -15,7 +15,7 @@ import Init
 share
     [ mkPersist persistSettings { mpsGeneric = False }
     , mkMigrate "compositeMigrate"
-    -- , mkDeleteCascade persistSettings { mpsGeneric = False }
+    , mkDeleteCascade persistSettings { mpsGeneric = False }
     ] [persistLowerCase|
   TestParent
       name  String maxlen=20
--- a/persistent-test/src/CompositeTest.hs
+++ b/persistent-test/src/CompositeTest.hs
@@ -12,7 +12,11 @@ import Init
 
 
 -- mpsGeneric = False is due to a bug or at least lack of a feature in mkKeyTypeDec TH.hs
-share [mkPersist persistSettings { mpsGeneric = False }, mkMigrate "compositeMigrate", mkDeleteCascade persistSettings { mpsGeneric = False }] [persistLowerCase|
+share
+    [ mkPersist persistSettings { mpsGeneric = False }
+    , mkMigrate "compositeMigrate"
+    -- , mkDeleteCascade persistSettings { mpsGeneric = False }
+    ] [persistLowerCase|
   TestParent
       name  String maxlen=20
       name2 String maxlen=20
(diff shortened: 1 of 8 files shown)


persistent-test/src/CompositeTest.hs
```diff
@@ -12,7 +12,11 @@ import Init
 
 
 -- mpsGeneric = False is due to a bug or at least lack of a feature in mkKeyTypeDec TH.hs
-share [mkPersist persistSettings { mpsGeneric = False }, mkMigrate "compositeMigrate", mkDeleteCascade persistSettings { mpsGeneric = False }] [persistLowerCase|
+share
+    [ mkPersist persistSettings { mpsGeneric = False }
+    , mkMigrate "compositeMigrate"
+    , mkDeleteCascade persistSettings { mpsGeneric = False }
```
parsonsmatt: Currently this fails with:

```

    /home/matt/Projects/persistent/persistent-test/src/CompositeTest.hs:15:1: error:
        • Couldn't match type ‘Key TestParent’ with ‘[Char]’
            arising from a use of ‘TestChildName’
        • In the first argument of ‘Filter’, namely ‘TestChildName’
          In the expression:
            ((Filter TestChildName) (FilterValue key_aGp3)) Eq
          In the first argument of ‘deleteCascadeWhere’, namely
            ‘[((Filter TestChildName) (FilterValue key_aGp3)) Eq,
              ((Filter TestChildName2) (FilterValue key_aGp3)) Eq,
              ((Filter TestChildAge) (FilterValue key_aGp3)) Eq]’
       |
    15 | share
       | ^^^^^...
```

Previously, we weren't generating *anything* here. But now we're generating something like:

```
deleteWhere [TestChildName ==. k, TestChildName2 ==. k, TestChildAge ==. k]
```

which is obviously not right. So the "fix" in thsi PR is borked.

persistent-test/src/DeleteCascadeTest.hs
```diff
@@ -0,0 +1,49 @@
...
+        liftIO $ void xs `shouldBe` []
+
+    it "should delete by non-primary key refernece" $ do
+        pendingWith "Depends on specifying non-PK references: see issue #909"
+        runDb $ do
+            pk <-insert (PersonX "username0" "name0" Nothing)
+            ak <- insert (AliasX "name0" "othername")
+            deleteCascade pk
+            ma <- get ak
+            liftIO $ void ma `shouldBe` Nothing
```
parsonsmatt: @rvl This should take care of your test case, right?

rvl: @parsonsmatt Thanks for looking at this and trying the fix. Yes it does match my use case. I can work around it by adding more `deleteWhere` statements though. If `persistent-sqlite` could add foreign key constraints to the table, that would be ideal, although I sense there might be some limitation of persistent and SQLite together that prevents this.

parsonsmatt: I think that `persistent-sqlite` does add foreign key indexes. It doesn't specify the cascade mode though. Typically you'd write something like:

```sql
CREATE TABLE foo (
  blah_id INT REFERENCES blah(id) ON DELETE CASCADE
);
```

and when the relevant `blah` gets deleted, the `foo` entries get deleted as well.

I do want to extend the QQ syntax with a cascade-mode for deletes and updates. Something like:

```
Foo
   blahId  BlahId OnDelete Cascade OnUpdate Restrict
```

would be reflected in the generated SQL migrations and then you wouldn't need the "manual" cascading entries.

PR #913, opened 2019-05-22
Atomic  insertUnique in SqlBackend and PostgreSQL 9.5+ support

parsonsmatt:
The `insertUnique` default implementation first checks for conflicts,
one query per unique constraint, and no conflict is found, it inserts a
new record. That insertion may fail, if concurrent insertions happen
while those conflict checks are being executed.

In PostgreSQL, the default isolation level allows such concurrent
insertions to be visible and make the insertion query throw an error.
But since version 9.5, PostgreSQL has upsert support, which includes the
"ON CONFLICT DO NOTHING" clause. It allows to atomically
insert-only-if-no-conflicts, even in committed-read isolation level,
without needing to switch to repeatable-read.

This commit adds an SqlBackend field allowing SQL backends to provide
SQL text for insert-if-no-conflicts, and SqlBackend's implementation of
insertUnique uses it if provided, otherwise resorting to the old
default.

This commit also provides this new field for the PostgreSQL backend.

Note: The insertion isn't 100% atomic as in actual repeatable-read. For
example, if the inserted record contains a foreign key referring to a
record being concurrently deleted, it will cause an error. But that's
true for regular insert too. The guarantee is that if an error is
raised, unique constraint conflicts aren't the cause, even in a
concurrent situation (while by default, in committed-read it's only
guaranteed for non-concurrent cases).

After submitting your PR:


persistent/Database/Persist/Sql/Types/Internal.hs
```diff
@@ -145,6 +145,14 @@ data SqlBackend = SqlBackend
...
     --
     -- @since 2.9.0
+    , connInsertUniqueSql :: Maybe (EntityDef -> [PersistValue] -> InsertSqlResult)
+    -- ^ Some databases support atomically checking for conflicts before
+    -- insertion, and skipping insertion (doing nothing) if there is a conflict.
+    --
+    -- When this field is left as 'Nothing', we default to checking for each
+    -- unique constraint separately, and finally insert if they all pass.
+    --
+    -- @since 2.10.0
```
parsonsmatt:
```suggestion
    -- @since 2.11.0
```

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1311,6 +1313,27 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+                else T.concat
+                    [ "("
+                    , T.intercalate "," $ map (escape . fieldDB) $ entityFields ent
+                    , ") VALUES("
+                    , T.intercalate "," $ map (const "?") $ entityFields ent
+                    , ") ON CONFLICT ("
+                    , T.intercalate "," $ concat $ map (\x -> map escape (map snd $ uniqueFields x)) (entityUniques ent)
+                    , ") DO NOTHING"
+                    ]
+            ]
```
parsonsmatt: This code is failing the following test:

```haskell
  it "insertUniqueEntity" $ runDb $ do
    let fo = Fo 3 5
    Just (Entity _ insertedFoValue) <- insertUniqueEntity fo
    Nothing <- insertUniqueEntity fo
    fo @== insertedFoValue
```

Because we want to return `Nothing` if the record was already present, while `ON CONFLICT DO NOTHING` will have the code proceed without error.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1311,6 +1313,27 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+            , escape $ entityDB ent
+            , if null (entityFields ent)
+                then " DEFAULT VALUES"
+                else T.concat
+                    [ "("
+                    , T.intercalate "," $ map (escape . fieldDB) $ entityFields ent
+                    , ") VALUES("
+                    , T.intercalate "," $ map (const "?") $ entityFields ent
+                    , ") ON CONFLICT ("
+                    , T.intercalate "," $ concat $ map (\x -> map escape (map snd $ uniqueFields x)) (entityUniques ent)
```
parsonsmatt: This is also not taking into account the `entityId`, so in the test suite example, we're not catching a conflict on `foo`, the `Primary` key.

parsonsmatt:
I'm going to close this PR. It was submitted by someone that refuses to use GitHub, and I don't have time to manage communication on that front. If someone wants to revive it, feel free!

PR #915, opened 2019-06-02
Fix qmarks

kderme:
This pr fixes a small bug, which appears when using the `Filter` data type. The test added fails with
```
  src/PersistentTest.hs:87:3: 
  1) persistent Filter In
       uncaught exception: SqliteException
       SQLite3 returned ErrorError while attempting to perform prepare "SELECT \"id\", \"name\", \"age\", \"color\" FROM \"Person\" WHERE (\"name\" IN ?)": near "?": syntax error
```
The issue is that `?` should be wrapped in `( )` or sqlite can't parse it.

There seems to be more issues related to the `Filter` data type. For example someone could write `Filter PersonName (FilterValues ["a", "b"]) Eq` which will also fail. An additional fix could be:
- Hide Filter from users so that it is a Opaque type. Users will have to use the combinators from `Database.Persist`
- make constuctors safer, so that ie `Eq` can only be combined with `FilterValue` and not `FilterValues`.

After submitting your PR:

### DIFF
--- a/persistent-test/src/PersistentTest.hs
+++ b/persistent-test/src/PersistentTest.hs
@@ -84,6 +84,10 @@ specsWith runDb = describe "persistent" $ do
       ps <- selectList [FilterAnd []] [Desc PersonAge]
       assertNotEmpty ps
 
+  it "Filter In" $ runDb $ do
+    _ <- selectList [Filter PersonName (FilterValue "Kostas") In] []
+    return ()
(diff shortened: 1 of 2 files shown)


parsonsmatt:
Thanks for the PR! I'm going to merge to a new branch to get CI to run, and if that's green, then I'll get this released tonight.

PR #925, opened 2019-07-01
Can read SQLite text columns with nulls (fixes #921)

### LINKED ISSUE #921: `\NUL` values not allowed in the middle of Sqlite texts
I got this test failure in one of my own projects that uses `persistent-sqlite`:

```
    -   expected: "h\910325\NULL\1073793\562934"
        but got: "h\910325"
```

Given that something goes wrong with a null byte, I figured that there was probably something involving C code going wrong.

The tests in `persistent:test:test` don't test for text values with nulls:
https://github.com/yesodweb/persistent/blob/a3eed2f7c5fed085abfeac209fff2e2d204ffbd0/persistent-test/src/Init.hs#L187-L195

This was supposedly fixed in #310 but when
but when I comment out that line (line 192 in `persistent-test/src/Init.hs`), I get this test failure in `persistent-sqlite:test:test`:

```
       expected: ("text","@\NUL7")
        but got: ("text","@")
```

While fixing this problem, would it be possible to also remove the other filters for `arbText`:

before:
```                                                                       
arbText :: Gen Text                                                                                                   
arbText =                                                                                                             
     T.pack                                                                                                           
  .  filter ((`notElem` forbidden) . generalCategory)                                                                 
  .  filter (<= '\xFFFF') -- only BMP                                                                                 
  .  filter (/= '\0')     -- no nulls                                                                              
  .  T.unpack                                                                                                         
  <$> arbitrary                                                                                                       
  where forbidden = [NotAssigned, PrivateUse]    
```
after:
```
arbText :: Gen Text
arbText = arbitrary
```

snoyberg:


NorfairKing:
Wohoo :tada:

NorfairKing:
Nit: It doesn't look like the line that filters out null characters from texts in the generated values in the tests suite was deleted in this PR. Is that intentional?

snoyberg:
Intentional in this case. I didn't know if other backends still require it to be disabled.

NorfairKing:
Ok!

PR #926, opened 2019-07-15
Support for composite primary key migration in SQLite. Fixes #669.

### LINKED ISSUE #669: Migrating a table with a custom primary key fails in SQLite
# Description

Let's say we have this model definition

    A
      fieldA Int
      Primary fieldA

and change it to

    A
      fieldA Int
      fieldB Int default=0
      Primary fieldA

The migration fails with `table a_backup has no column named id`, because the generated SQL still includes the non-existing `id`-column although we have a custom primary key:

```sql
CREATE TEMP TABLE "a_backup"("fieldA" INTEGER NOT NULL,"field_b" INTEGER NOT NULL DEFAULT 0, PRIMARY KEY ("fieldA"));
INSERT INTO "a_backup"("id","fieldA") SELECT "id","fieldA" FROM "a";
```

# Identified problem

The problem lies in [these](https://github.com/yesodweb/persistent/blob/master/persistent-sqlite/Database/Persist/Sqlite.hs#L380) [two](https://github.com/yesodweb/persistent/blob/master/persistent-sqlite/Database/Persist/Sqlite.hs#L383) lines, as they prepend the `id` column regardless of its actual existence.

# Patch

This following patch works for me, but I don't really understand the [`ReferenceDef`](https://github.com/yesodweb/persistent/blob/master/persistent/Database/Persist/Types/Base.hs#L178,L185) type, but if I have got it correctly and there are no other cases to consider, I'll prepare a pull request + test case.

```diff
diff --git a/persistent-sqlite/Database/Persist/Sqlite.hs b/persistent-sqlite/Database/Persist/Sqlite.hs
index ab72ec7..d811578 100644
--- a/persistent-sqlite/Database/Persist/Sqlite.hs
+++ b/persistent-sqlite/Database/Persist/Sqlite.hs
@@ -375,12 +375,11 @@ getCopyTable allDefs getter def = do
     let oldCols = map DBName $ filter (/= "id") oldCols' -- need to update for table id attribute ?
     let newCols = filter (not . safeToRemove def) $ map cName cols
     let common = filter (`elem` oldCols) newCols
-    let id_ = fieldDB (entityId def)
     return [ (False, tmpSql)
-           , (False, copyToTemp $ id_ : common)
+           , (False, copyToTemp $ addIdCol common)
            , (common /= filter (not . safeToRemove def) oldCols, dropOld)
            , (False, newSql)
-           , (False, copyToFinal $ id_ : newCols)
+           , (False, copyToFinal $ addIdCol newCols)
            , (False, dropTmp)
            ]
   where
@@ -418,6 +417,10 @@ getCopyTable allDefs getter def = do
         , " FROM "
         , escape tableTmp
         ]
+    addIdCol = let id_ = entityId def
+                in case fieldReference id_ of
+                     CompositeRef _ -> id
+                     _ -> (fieldDB id_ :)
 
 mkCreateTable :: Bool -> EntityDef -> ([Column], [UniqueDef]) -> Text
 mkCreateTable isTemp entity (cols, uniqs) =

```

merijn:
PR for issue #669, using the patch from the issue which seems to be working fine on my systems.

After submitting your PR:

### DIFF
--- a/persistent-sqlite/test/main.hs
+++ b/persistent-sqlite/test/main.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
@@ -76,6 +77,23 @@ DataTypeTable no-json
     pico Pico
     time TimeOfDay
     utc UTCTime
+|]
+
+share [mkPersist sqlSettings, mkMigrate "compositeSetup"] [persistLowerCase|
+Simple
+    int Int
+    text Text
+    Primary text int
+    deriving Show Eq
+|]
+
+share [mkPersist sqlSettings, mkMigrate "compositeMigrateTest"] [persistLowerCase|
+Simple2 sql=simple
+    int Int
+    text Text
+    bool Bool
+    Primary text int
+    deriving Show Eq
 |]
 
 instance Arbitrary DataTypeTable where
@@ -207,6 +225,10 @@ main = do
     it "issue #527" $ asIO $ runSqliteInfo (mkSqliteConnectionInfo ":memory:") $ do
         runMigration migrateAll
         insertMany_ $ replicate 1000 (Test $ read "2014-11-30 05:15:25.123")
+
+    it "issue #669" $ asIO $ runSqliteInfo (mkSqliteConnectionInfo ":memory:") $ do
+        runMigration compositeSetup
+        runMigration compositeMigrateTest
 
     it "afterException" $ asIO $ runSqliteInfo (mkSqliteConnectionInfo ":memory:") $ do
         runMigration testMigrate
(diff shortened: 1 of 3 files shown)


parsonsmatt:
This looks good to me and I approve of the logic. I'd like to see a test case before merging it in.

Why isn't CI running or visible on this? @snoyberg  ?

snoyberg:
Looks like the Travis webhook grants somehow were revoked. Hopefully it should be fixed now. @merijn could you try force-pushing to this branch?

merijn:
I'll just finish my attempt to add tests for this and then push again. Two birds with one stone and all that.

parsonsmatt:
Feel free to publish a WIP commit up, I'd be happy to give feedback on in-progress stuff.

merijn:
Added the simplest test I could think off. For me the test fails without the above fix (SQLite error about missing column id) and works with this fix applied.

persistent-sqlite/test/main.hs
```diff
@@ -208,6 +226,10 @@ main = do
         runMigration migrateAll
         insertMany_ $ replicate 1000 (Test $ read "2014-11-30 05:15:25.123")
 
+    it "issue #669" $ asIO $ runSqliteInfo (mkSqliteConnectionInfo ":memory:") $ do
```
parsonsmatt:
```suggestion
    it "properly migrates to a composite primary key (issue #669)" $ asIO $ runSqliteInfo (mkSqliteConnectionInfo ":memory:") $ do
```

parsonsmatt:
Oh, and we do need a Changelog entry for this. You're racking them up this release :smile:

PR #927, opened 2019-07-15
Enable the SQLITE_ENABLE_STAT4 flag by default.

merijn:
STAT4 expands the ANALYZE command's logic, logging more data when using ANALYZE allowing for better query plan optimisation. This PR enables this code by default and adds flags to toggle which profiling/analysis code to build.

### DIFF
--- a/persistent-sqlite/persistent-sqlite.cabal
+++ b/persistent-sqlite/persistent-sqlite.cabal
@@ -36,10 +36,10 @@ flag json1
   description: Enable json1 in the vendored SQLite library; has no effect if a system SQLite library is used.
   default: True
 flag use-stat3
-  description: Enable STAT3 in the vendored SQLite library; has no effect if a syste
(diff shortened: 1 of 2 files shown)


persistent-sqlite/persistent-sqlite.cabal
```diff
@@ -35,6 +35,12 @@ flag have-usleep
 flag json1
   description: Enable json1 in the vendored SQLite library; has no effect if a system SQLite library is used.
   default: True
+flag use-stat3
+  description: Enable STAT3 in the vendored SQLite library; has no effect if a syste
```
parsonsmatt: looks like this was cut off?

merijn: Yeah, I was manually cherrypicking and made a copy/paste fail, fixed in the new commit.

PR #931, opened 2019-07-17
fix qmarks

parsonsmatt:
tagging @kderme who originally wrote this up, a continuation of #915


parsonsmatt:
@kderme I added a note to the datatype saying that it's possible to use it unsafely, and to prefer using the other functions that the API provides. Do you think this satisfies your case?

kderme:
> @kderme I added a note to the datatype saying that it's possible to use it unsafely, and to prefer using the other functions that the API provides. Do you think this satisfies your case?

Yes, thank you!

PR #934, opened 2019-07-17
Replace liftPersist, fixes #933

### LINKED ISSUE #933: persistent-sqlite 2.10.2/3 should be deprecated
As mentioned in https://github.com/yesodweb/persistent/pull/932#issuecomment-512273624 the instances for RawSqlite were (unfortunately) broken, because the functionality of `liftPersist` changed between the original patch and persistent 2.10

I fixed the instances and am adding tests, but I hit a snag with the tests. The tests defined in `persistent-test` all seem to have a `backend ~ BaseBackend backend` constraint, so I can't simply rerun those tests with the RawSqlite backend without modifying `persistent-test`. This isn't a hard problem to fix, but I'm not sure if removing that constraint from the various `specsWith` definitions will break anything...

merijn:
This fixes the broken instances in 2.10.2 and 2.10.3, but we should probably add tests to avoid this happening again.

### DIFF
--- a/persistent-sqlite/Database/Persist/Sqlite.hs
+++ b/persistent-sqlite/Database/Persist/Sqlite.hs
@@ -46,7 +46,7 @@ import qualified Control.Exception as E
 import Control.Monad (forM_)
 import Control.Monad.IO.Unlift (MonadIO (..), MonadUnliftIO, withRunInIO, withUnliftIO, unliftIO, withRunInIO)
 import Control.Monad.Logger (NoLoggingT, runNoLoggingT, MonadLogger, logWarn, runLoggingT)
-import Control.Monad.Trans.Reader (ReaderT, runReaderT)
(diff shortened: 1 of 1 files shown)


parsonsmatt:
Ok, just getting back to a computer now - this should be good to release a 2.10.4 that is fixed up?

merijn:
Yeah, I haven't exhaustively tested them all, but everything I use works and given the simplicity of what broke and how this fixes it, it seems unlikely that there's any reason they wouldn't all work fine.

I'm working on ripping the "BaseBackend backed ~ backend" stuff out of the tests so they can be more easily applied to derived backends (and so I can rerun the tests with this backend), but it's a substantial amount of work and it's a bit low priority, so it might be awhile until I wrap it up.

parsonsmatt:
Uploaded as `persistent-sqlite-2.10.4` thanks :smile:

PR #937, opened 2019-07-31
Add support for constraint= attribute to override foreign reference constraint names

charukiewicz:
This change adds support for a `constraint=` attribute that allows users to override the automatically generated foreign reference constraint names with their own. The motivation for this change is that we had two instances of very long table/column names that led to automatically generated constraint names exceeding 64 characters, which [is not allowed in MySQL](https://dev.mysql.com/doc/refman/5.7/en/identifiers.html). Without a change like this we could not use the `printMigrations` output.

Please let myself (@charukiewicz) or @belevy know if you have suggestions for a better solution or alternative approach.

After submitting your PR:

### DIFF
--- a/docs/Persistent-entity-syntax.md
+++ b/docs/Persistent-entity-syntax.md
@@ -136,6 +136,18 @@ This makes a unique index requiring `phone` to be unique across `Person` rows. O
 
 The [tests for this feature](https://github.com/yesodweb/persistent/blob/master/persistent-test/src/CompositeTest.hs#L53) demonstrates their usage
 
+### constraint=
+
+You can use the `constraint=` attribute to override the constraint name used in migrations. This is useful particularly when the automatically generated constraint names exceed database limits (e.g. MySQL does not allow constraint names longer than 64 characters).
+
+```
+VeryLongTableName
+  name Text
+
+AnotherVeryLongTableName
+  veryLongTableNameId VeryLongTableNameId constraint=short_foreign_key
+```
+
 ## Laziness
 
 By default the records created by persistent have strict fields. You can prefix a field name with `~` to make it lazy (or `!` to make it strict).
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -341,8 +341,8 @@ migrate' connectInfo allDefs getter val = do
                         AddUniqueConstraint uname $
                         map (findTypeAndMaxLen name) ucols ]
         let foreigns = do
-              Column { cName=cname, cReference=Just (refTblName, _a) } <- newcols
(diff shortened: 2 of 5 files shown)


docs/Persistent-entity-syntax.md
```diff
@@ -136,6 +136,18 @@ This makes a unique index requiring `phone` to be unique across `Person` rows. O
...
+You can use the `constraint=` attribute to override the constraint name used in migrations. This is useful particularly when the automatically generated constraint names exceed database limits (e.g. MySQL does not allow constraint names longer than 64 characters).
+
+```
+VeryLongTableName
+  name Text
+
+AnotherVeryLongTableName
+  veryLongTableNameId VeryLongTableNameId constraint=short_foreign_key
+```
+
```
parsonsmatt: Thanks for extending the docs!

persistent/Database/Persist/Sql/Internal.hs
```diff
@@ -61,9 +61,13 @@ mkColumns allDefs t =
...
-    ref c _ (a:_)
-        | Just x <- T.stripPrefix "reference=" a =
-            Just (DBName x, refName tn c)
+    ref c fe (a:as)
+        | Just x <- T.stripPrefix "reference=" a = do
+            constraintName <- snd <$> (ref c fe as)
+            pure (DBName x, constraintName)
+        | Just x <- T.stripPrefix "constraint=" a = do
+            tableName <- fst <$> (ref c fe as)
+            pure (tableName, DBName x)
```
parsonsmatt: Oh, I see how this is implemented now! It doesn't require a breaking change addition to a datatype, which is great. It adds a new feature to the QuasiQuoter syntax, but doesn't change the `persistent-template` library at all.

So if you have a QuasiQuoter block with `constraint=fk_name` on an old version of `persistent`, then it'll get ignored. That's okay.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -704,15 +704,15 @@ findAlters tblName allDefs col@(Column name isNull type_ def _defConstraintName
...
+               Just (tname, cname) -> let cnstr = [addReference allDefs cname tname name]
                                   in (map ((,) tname) (Add' col : cnstr), cols)
         Column _ isNull' type_' def' _defConstraintName' maxLen' ref':_ ->
             let -- Foreign key
                 refDrop = case (ref == ref', ref') of
                             (False, Just (_, cname)) -> [(name, DropReference cname)]
                             _ -> []
                 refAdd  = case (ref == ref', ref) of
-                            (False, Just (tname, _cname)) -> [(tname, addReference allDefs (refName tblName name) tname name)]
+                            (False, Just (tname, cname)) -> [(tname, addReference allDefs cname tname name)]
```
parsonsmatt: This looks great to me. I think I would like to see it included in the other SQL backends, but that can come in a later PR.

PR #938, opened 2019-08-05
Tidying up code in Persist/TH.hs

danbroooks:
Another refactor/tidy up of the `TH.hs` file, makes the formatting/indentation consistent across all the functions, extracts/renames a function used to read a file from the filesystem, and makes some code fixes suggested by `hlint`.

### DIFF
--- a/persistent-template/Database/Persist/TH.hs
+++ b/persistent-template/Database/Persist/TH.hs
@@ -54,6 +54,7 @@ import Data.Aeson
     , Value (Object), (.:), (.:?)
     , eitherDecodeStrict'
     )
+import qualified Data.ByteString as BS
 import Data.Char (toLower, toUpper)
 import qualified Data.HashMap.Strict as HM
 import Data.Int (Int64)
@@ -163,10 +164,7 @@ persistManyFileWith ps fps = do
     parseReferences ps s
 
 getFileContents :: FilePath -> IO Text
-getFileContents fp = do
-    h <- SIO.openFile fp SIO.ReadMode
-    SIO.hSetEncoding h SIO.utf8_bom
-    TIO.hGetContents h
+getFileContents = fmap decodeUtf8 . BS.readFile
 
 -- | Takes a list of (potentially) independently defined entities and properly
 -- links all foreign keys to reference the right 'EntityDef', tying the knot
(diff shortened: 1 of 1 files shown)


persistent-template/Database/Persist/TH.hs
```diff
@@ -154,16 +155,16 @@ persistFileWith ps fp = persistManyFileWith ps [fp]
...
+  mapM_ qAddDependentFile fps
+  ss <- mapM (qRunIO . getFileContents) fps
+  let s = T.intercalate "\n" ss -- be tolerant of the user forgetting to put a line-break at EOF.
+  parseReferences ps s
+
+getFileContents :: FilePath -> IO Text
+getFileContents fp = do
+  h <- SIO.openFile fp SIO.ReadMode
+  SIO.hSetEncoding h SIO.utf8_bom
+  TIO.hGetContents h
```
parsonsmatt: While we're fixing this up, it may be good to port away from `Data.Text.IO`: https://www.snoyman.com/blog/2016/12/beware-of-readfile

persistent-template/Database/Persist/TH.hs
```diff
@@ -195,22 +196,23 @@ embedEntityDefsMap rawEnts = (embedEntityMap, noCycleEnts)
...
 
     breakCycleEmField ancestors emf = case embeddedHaskell <$> membed of
-        Nothing -> emf
-        Just embName -> if embName `elem` ancestors
-          then emf { emFieldEmbed = Nothing, emFieldCycle = Just embName }
-          else emf { emFieldEmbed = breakCycleEmbed ancestors <$> membed }
+      Nothing -> emf
+      Just embName -> if embName `elem` ancestors
+        then emf { emFieldEmbed = Nothing, emFieldCycle = Just embName }
+        else emf { emFieldEmbed = breakCycleEmbed ancestors <$> membed }
```
parsonsmatt: The convention in this package is to use 4 space indent, not 2 space. Please revert all 2 space indentation changes and  use 4 space indent on the project.

parsonsmatt:
be careful with redundant imports, some of them are necessary for backwards compatibility on old versions of `base`

danbroooks:
@parsonsmatt this should be good for another look now. I wasn't sure on the correct formatting of `where` blocks, so you may need to check that, I went with what I could see was common in the rest of the file/project. the redundant imports that were dropped were the ones being used in the `getFileContents` function, I'm thinking its fine to drop them now?

PR #945, opened 2019-08-19
Support Table Foreign Key Constraints in SQLite

kderme:
This pr adds support for foreign keys, table constraints in SQLite. They can only reference Primary keys of other tables and not Unique keys, since this would require changes in `Quasi` and I'm not sure how well this is supported in other dbs. I've tested migrations from existing schemas, which adds new foreign keys to existing Tables and it also works fine.

After submitting your PR:

### DIFF
--- a/persistent-sqlite/ChangeLog.md
+++ b/persistent-sqlite/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for persistent-sqlite
 
+## 2.10.5
+
+* Foreign keys table constraints are correctly generated [#945](https://github.com/yesodweb/persistent/pull/945) @kderme
+
 ## 2.10.4
 
 * Fix bug with 2.10.3 and 2.10.2 that caused the `RawSqlite` loop. [#934](https://github.com/yesodweb/persistent/pull/934) @merijn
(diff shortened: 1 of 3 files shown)


parsonsmatt:
Code looks good to me.

> They can only reference Primary keys of other tables and not Unique keys, since this would require changes in Quasi and I'm not sure how well this is supported in other dbs.

Can you elaborate?

kderme:
Sure! What I mean is that this migration works fine
```
share [mkPersist sqlSettings, mkSave "entityDefs"] [persistLowerCase|
Person
    name String
    lastNameId Int
    Primary name lastNameId
    deriving Eq Show

Car
   owner String
   ownerId Int
   color String
   Foreign Person belongs owner ownerId
|] 
```
but this
```

share [mkPersist sqlSettings, mkSave "entityDefs"] [persistLowerCase|
Person
    name String
    lastNameId Int
    UniquePerson name lastNameId
    deriving Eq Show

Car
   owner String
   ownerId Int
   color String
   Foreign Person belongs owner ownerId
|]
```
still fails with 
```
     • Exception when trying to run compile-time code:
        no explicit primary key fdef=ForeignDef ...
```
Such a migration is accepted by SQLite (actually now that I checked again, SQLites accepts it even without the unique constraint).

parsonsmatt:
Right, thanks. I'm good to accept and merge this as-is then if you want to bump the relevant version numbers and add a changelog entry :smile:

parsonsmatt:
CI is passing, merging and releasing. Thanks!

Issue #949, opened 2019-08-21
Transactions with Postgres and Persistent


erikd:
I am using Persistent version 2.10.1 with PostgreSQL.

I have an insert where I:

* Do an insert which returns an `Id`
* Do more inserts which use the `Id` returned from the first insert.

In order to provide these an an atomic operation, I would like to do the these two inserts within a transaction. The Persistent library provides two basic functions `transactionSave` and `transactionUndo`. 

Just sprinking my code with some `transactionSave` statements resulted in the warning:
```
WARNING:  there is no transaction in progress
```
and I am not even sure where that comes from. Its definitely not from Persistent itself.

I then tried enclosing the two inserts within a `transactionUndo` and `transactionSave` pair and get the same warning message.

It seems like the API is missing a `transactionStart` function or better yet, a `withTransaction` function.

parsonsmatt:
You should be using the `runSqlConn` or `runSqlPool` functions. These create a transaction and provide rollback-on-exception behavior. `transactionSave` and `transactionUndo` both work correctly in this context.

I consider this a documentation problem and want to keep it open until this is more obvious.

erikd:
Is my PR #951 sufficient? If so, please close this.

PR #950, opened 2019-08-21
Database.Persist.Sql: Add function transactionBegin

erikd:
This is a fix for the issue #949. I have built my code against this fix and it seems to do the right thing. The `withTransaction` function is less that great though. Happy to get suggestions for improvements.

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Sql.hs
+++ b/persistent/Database/Persist/Sql.hs
@@ -12,6 +12,9 @@ module Database.Persist.Sql
     , rawSql
     , deleteWhereCount
     , updateWhereCount
+    , transactionBegin
+    , transactionCommit
+    , withTransaction
     , transactionSave
     , transactionSaveWithIsolation
     , transactionUndo
@@ -38,6 +41,24 @@ import Database.Persist.Sql.Internal
 import Database.Persist.Sql.Orphan.PersistQuery
 import Database.Persist.Sql.Orphan.PersistStore
 import Database.Persist.Sql.Orphan.PersistUnique ()
+
+-- | Begin an SQL transaction.
+--
+-- @since 2.10.2
+transactionBegin :: MonadIO m => ReaderT SqlBackend m ()
+transactionBegin = do
+    conn <- ask
+    let getter = getStmtConn conn
+    liftIO $ connBegin conn getter Nothing
+
+-- | Begin an SQL transaction.
+--
+-- @since 2.10.2
+transactionCommit :: MonadIO m => ReaderT SqlBackend m ()
+transactionCommit = do
+    conn <- ask
+    let getter = getStmtConn conn
+    liftIO $ connCommit conn getter
 
 -- | Commit the current transaction and begin a new one.
 --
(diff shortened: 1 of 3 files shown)


persistent/Database/Persist/Sql.hs
```diff
@@ -39,6 +42,24 @@ import Database.Persist.Sql.Orphan.PersistQuery
 import Database.Persist.Sql.Orphan.PersistStore
 import Database.Persist.Sql.Orphan.PersistUnique ()
 
+-- | Begin an SQL transaction.
+--
+-- @since 2.10.2
+transactionBegin :: MonadIO m => ReaderT SqlBackend m ()
```
parsonsmatt: This needs to be generalized to `(MonadIO m, BackendCompatible SqlBackend backend) => ReaderT backend m ()`.

persistent/Database/Persist/Sql.hs
```diff
@@ -39,6 +42,24 @@ import Database.Persist.Sql.Orphan.PersistQuery
...
+transactionBegin = do
+    conn <- ask
+    let getter = getStmtConn conn
+    liftIO $ connBegin conn getter Nothing
+
+-- | Begin an SQL transaction.
+--
+-- @since 2.10.2
+transactionCommit :: MonadIO m => ReaderT SqlBackend m ()
```
parsonsmatt: Ditto

persistent/Database/Persist/Sql.hs
```diff
@@ -74,3 +95,17 @@ transactionUndoWithIsolation isolation = do
...
+--
+-- @since 2.10.2
+withTransaction :: MonadIO m => ReaderT SqlBackend m a -> ReaderT SqlBackend m a
+withTransaction action = do
+    conn <- ask
+    let getter = getStmtConn conn
+    liftIO $ connBegin conn getter Nothing
+    a <- action
+    liftIO $ connCommit conn getter
+    pure a
```
parsonsmatt: So this isn't exception-safe. We'll get a dangling transaction if `action` throws. The function `runSqlConn` does this right:

```haskell
runSqlConn :: (MonadUnliftIO m, BackendCompatible SqlBackend backend) => ReaderT backend m a -> backend -> m a
runSqlConn r conn = withRunInIO $ \runInIO -> mask $ \restore -> do
    let conn' = projectBackend conn
        getter = getStmtConn conn'
    restore $ connBegin conn' getter Nothing
    x <- onException
            (restore $ runInIO $ runReaderT r conn)
            (restore $ connRollback conn' getter)
    restore $ connCommit conn' getter
    return x
```

But, if you're using this function (which you should be), then you're always going to be in a transaction. So doing `BEGIN` here would make a nested transaction, which is not really well supported in many databases, and likely not what is desired.

erikd:
Thanks for the comments @parsonsmatt .

I am currently using `withPostgresqlConn` from the `persistent-postgresql`. Maybe that needs a comment to point people to `runSqlConn`. Comments on the existing `transactionSave` and `transactionUndo` functions.

erikd:
Ah, so using `withPostgesqlConn` I should use `runSqlConn` instead of `runReaderT`!!

erikd:
Will close this and submit a new one that just add the much needed comments.

erikd:
> However, it's becoming clear to me that a quick type-search is going to get you from `SqlPersistT m a ~ ReaderT SqlBackend m a`, and you'll reach for `flip runReaderT sqlBackend` to do this. That must be what's happening, because I've seen this pattern at my current employer and now you're asking for it.

Yes, that is how I go there. I am now using `runSqlConn` instead of `runReaderT`. Comments submitted as #951 .

PR #954, opened 2019-09-12
Select first record or create a new one if there is no match

sestrella:

After submitting your PR:

### DIFF
--- a/persistent-test/src/PersistentTest.hs
+++ b/persistent-test/src/PersistentTest.hs
@@ -438,6 +438,18 @@ specsWith runDb = describe "persistent" $ do
       x <- selectFirst [] [Desc PersonAge]
       x @== Just (Entity kOld pOld)
 
+  describe "selectFirstOrInsert" $ do
+      it "returns a record if there is one matching the search criteria" $ runDb $ do
+          insert_ $ Person "John" 26 Nothing
+          p <- insertEntity $ Person "Michael" 26 Nothing
+          x <- selectFirstOrInsert [PersonName ==. "Michael"] [] (Person "Michael" 26 Nothing)
+          x @== p
+
+      it "creates a new record if there is no record matching the search criteria" $ runDb $ do
+          insert_ $ Person "John" 26 Nothing
+          let p = Person "Michael" 26 Nothing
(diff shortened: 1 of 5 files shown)


Vlix:
This will throw an error when the to be inserted record can't be inserted because of a uniqueness constraint, for example. Is that the intended functionality?

sestrella:
@Vlix sorry for the late reply, perhaps bringing up the use case I had in mind might help here. So far I was thinking about using this function in scenarios like seed data, where the model looks like follows:

```hs
Role
  name        Text -- assuming there is a unique constraint on this column
  description Text
|]

findOrCreateRole Role{..}@role =  selectFirstOrInsert [RoleName ==. roleName] role
```

However, after a second thought, I completely agree with you this operation could fail, in which case perhaps it would make more sense to replace `[Filter record]` with `Unique record` in order to make more explicit there is a unique constraint:

```hs
getUniqueOrInsert
  :: (OnlyOneUniqueKey record, ...)
  => record
getUniqueOrInsert rec = getUniqueByOrInsert (onlyUniqueP rec) rec

getUniqueByOrInsert
  :: (..)
  => Unique record 
  -> record
getUniqueByOrInsert unique record = do
  mrecord <- getBy unique
  case mrecord of
    Nothing   -> insertEntity record
    (Just rec) -> return rec
```

What do you think?

Vlix:
Honestly, it sounds like you just want to use `insertBy`, which takes a record and returns `Either (Entity record) (Key record)`.
Using the record you inserted after the `Right` case is returned would be identical to what you're suggesting, and I don't really see why that would need to be a new function in the library...

```haskell
myFunc rec = insertBy rec >>= return . either id (const rec)
```

parsonsmatt:
A problem here is that the unique records may not be the same in the non-unique fields. And a problem with the suggested PR is that there's no guarantee that the `entity` you insert satisfies the conditions you provided.

I'm generally inclined to include helpers and convenience functions into the library if the obvious implementation isn't right or fast for some reason. `maybe (insertEntity entity) pure =<< getBy unique` is very short, and I can't see why it wouldn't be right.

sestrella:
@Vlix, @parsonsmatt fair enough, I appreciate all the feedback, for now, I would go with the short hard function @Vlix suggested.

PR #958, opened 2019-09-18
persistent: Support GHC 8.8

ocharles:


parsonsmatt:
Do you have any plans to continue working on this?

ocharles:
Sure, but what more needs to be done? CI failures?

parsonsmatt:
It should be passing CI, yes. Right now, the code doesn't compile for previous GHC versions, it looks like.

PR #959, opened 2019-09-18
persistent-template: Support template-haskell-2.15

ocharles:


ocharles:
This PR should be squash merged, the history is not interesting.

parsonsmatt:
Thanks so much! Would you be willing to add a Changelog entry for this?

PR #977, opened 2019-10-25
Support Stackage Nightly

parsonsmatt:

After submitting your PR:

Well, `persistent-mongoDB`'s `mongoDB` dependency is causing issues, so we're not building it - otherwise it'd fail the whole build.

### DIFF
--- a/persistent-redis/Database/Persist/Redis/Internal.hs
+++ b/persistent-redis/Database/Persist/Redis/Internal.hs
@@ -12,6 +12,7 @@ import qualified Data.ByteString as B
 import qualified Data.ByteString.UTF8 as U
 import Data.Text (Text, unpack)
 import qualified Data.Text as T
+import Control.Monad.Fail (MonadFail)
 
 import Database.Persist.Class
 import Database.Persist.Types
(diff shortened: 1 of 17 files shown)


persistent-redis/Database/Persist/Redis/Internal.hs
```diff
@@ -26,7 +27,7 @@ toEntityString = unDBName . entityDB . entityDef . Just
 toEntityName :: EntityDef -> B.ByteString
 toEntityName = U.fromString . unpack . unDBName . entityDB
 
-mkEntity :: (Monad m, PersistEntity val) => Key val -> [(B.ByteString, B.ByteString)] -> m (Entity val)
+mkEntity :: (MonadFail m, PersistEntity val) => Key val -> [(B.ByteString, B.ByteString)] -> m (Entity val)
```
parsonsmatt: technically a breaking change, but it's an `.Internal` module, so yolo

persistent-redis/persistent-redis.cabal
```diff
@@ -1,5 +1,5 @@
 name:            persistent-redis
-version:         2.5.2.2
+version:         2.5.2.3
```
parsonsmatt: should possibly be a major bump, but only type-changing thing is in the internal module

persistent-sqlite/test/main.hs
```diff
@@ -224,7 +224,7 @@ main = do
         return ()
     it "issue #527" $ asIO $ runSqliteInfo (mkSqliteConnectionInfo ":memory:") $ do
         runMigration migrateAll
-        insertMany_ $ replicate 1000 (Test $ read "2014-11-30 05:15:25.123")
+        insertMany_ $ replicate 1000 (Test $ read "2014-11-30 05:15:25.123Z")
```
parsonsmatt: I guess the `Read` instance for UTCTime changed :|

Issue #939, opened 2019-08-05
Support `constraint=` syntax on postgresql and sqlite backends


parsonsmatt:
#937 introduced `constraint=` syntax to the `persistent` quasiquoter and added support for `persistent-mysql`. This should also be added to `persistent-postgresql` and `persistent-sqlite`.

Phinneas:
I would like to take a crack at this issue.

PR #979, opened 2019-10-30
Postgresql `constraint=` attribute support

robbassi:
This PR adds support for the `constraint=` attribute in the Postgresql backend. It addresses part of issue #939. Unfortunately, I wasn't able to figure out how to accomplish this with sqlite, and I will make another PR for that once I get a better understanding of sqlite internals.

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -568,7 +568,7 @@ migrate' allDefs getter entity = fmap (fmap $ map showAlterDb) $ do
     migrationText exists old'' =
         if not exists
             then createText newcols fdefs udspair
-            else let (acs, ats) = getAlters allDefs entity (newcols, udspair) old'
+            else let (acs, ats) = getAlters allDefs entity (newcols, udspair) $ excludeForeignKeys $ old'
                      acs' = map (AlterColumn name) acs
                      ats' = map (AlterTable name) ats
                  in  acs' ++ ats'
@@ -577,6 +577,14 @@ migrate' allDefs getter entity = fmap (fmap $ map showAlterDb) $ do
          (newcols', udefs, fdefs) = mkColumns allDefs entity
          newcols = filter (not . safeToRemove entity . cName) newcols'
          udspair = map udToPair udefs
+         excludeForeignKeys (xs,ys) =
+           (map (\c -> case cReference c of
+                         Just (_,fk) ->
+                           case find (\f -> fk == foreignConstraintNameDBName f) fdefs of
+                             Just _ -> c { cReference = Nothing }
+                             Nothing -> c
+                         Nothing -> c) xs,ys)
+  
             -- Check for table existence if there are no columns, workaround
             -- for https://github.com/yesodweb/persistent/issues/152
 
@@ -792,18 +800,16 @@ getColumn getter tableName' [PersistText columnName, PersistText isNullable, Per
                 Nothing -> loop' ps
                 Just t' -> t'
     getRef cname = do
-        let sql = T.concat ["SELECT "
+        let sql = T.concat ["SELECT DISTINCT "
                            ,"ccu.table_name, "
                            ,"tc.constraint_name "
-                          -- ,"kcu.column_name, "
                            ,"FROM information_schema.constraint_column_usage ccu, "
                            ,"information_schema.key_column_usage kcu, "
                            ,"information_schema.table_constraints tc "
                            ,"WHERE tc.constraint_type='FOREIGN KEY' "
-                           ,"AND ccu.constraint_name=tc.constraint_name "
                            ,"AND kcu.constraint_name=tc.constraint_name "
-                           ,"AND tc.table_catalog=current_database() "
-                           ,"AND tc.table_schema=current_schema() "
+                           ,"AND ccu.constraint_name=kcu.constraint_name "
+                           ,"AND kcu.ordinal_position=1 "
                            ,"AND kcu.table_name=? "
                            ,"AND kcu.column_name=?"]
         stmt <- getter sql
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -577,14 +577,13 @@ migrate' allDefs getter entity = fmap (fmap $ map showAlterDb) $ do
          (newcols', udefs, fdefs) = mkColumns allDefs entity
          newcols = filter (not . safeToRemove entity . cName) newcols'
          udspair = map udToPair udefs
-         excludeForeignKeys (xs,ys) =
-           (map (\c -> case cReference c of
-                         Just (_,fk) ->
-                           case find (\f -> fk == foreignConstraintNameDBName f) fdefs of
-                             Just _ -> c { cReference = Nothing }
-                             Nothing -> c
-                         Nothing -> c) xs,ys)
-  
+         excludeForeignKeys (xs,ys) = (map excludeForeignKey xs,ys)
+         excludeForeignKey c = case cReference c of
+           Just (_,fk) ->
+             case find (\f -> fk == foreignConstraintNameDBName f) fdefs of
+               Just _ -> c { cReference = Nothing }
+               Nothing -> c
+           Nothing -> c
             -- Check for table existence if there are no columns, workaround
             -- for https://github.com/yesodweb/persistent/issues/152
 
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -792,23 +792,29 @@ getColumn getter tableName' [PersistText columnName, PersistText isNullable, Per
                 Nothing -> loop' ps
                 Just t' -> t'
     getRef cname = do
-        let sql = T.concat
-                [ "SELECT COUNT(*) FROM "
-                , "information_schema.table_constraints "
-                , "WHERE table_catalog=current_database() "
-                , "AND table_schema=current_schema() "
-                , "AND table_name=? "
-                , "AND constraint_type='FOREIGN KEY' "
-                , "AND constraint_name=?"
-                ]
-        let ref = refName tableName' cname
+        let sql = T.concat ["SELECT "
+                           ,"ccu.table_name, "
+                           ,"tc.constraint_name "
+                          -- ,"kcu.column_name, "
+                           ,"FROM information_schema.constraint_column_usage ccu, "
+                           ,"information_schema.key_column_usage kcu, "
+                           ,"information_schema.table_constraints tc "
+                           ,"WHERE tc.constraint_type='FOREIGN KEY' "
+                           ,"AND ccu.constraint_name=tc.constraint_name "
+                           ,"AND kcu.constraint_name=tc.constraint_name "
+                           ,"AND tc.table_catalog=current_database() "
+                           ,"AND tc.table_schema=current_schema() "
+                           ,"AND kcu.table_name=? "
+                           ,"AND kcu.column_name=?"]
         stmt <- getter sql
-        with (stmtQuery stmt
-                     [ PersistText $ unDBName tableName'
-                     , PersistText $ unDBName ref
-                     ]) (\src -> runConduit $ src .| do
-            Just [PersistInt64 i] <- CL.head
-            return $ if i == 0 then Nothing else Just (DBName "", ref))
+        cntrs <- with (stmtQuery stmt [PersistText $ unDBName tableName'
+                                      ,PersistText $ unDBName cname])
+                      (\src -> runConduit $ src .| CL.consume)
+        case cntrs of
+          [] -> return Nothing
+          [[PersistText table, PersistText constraint]] ->
+            return $ Just (DBName table, DBName constraint)
+          _ -> error "Postgresql.getColumn: error fetching constraints"
     d' = case defaultValue of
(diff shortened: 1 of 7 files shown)


robbassi:
> Looks great! Can you add a test to verify that it works?

For sure!

persistent-postgresql/test/CustomConstraintTest.hs
```diff
@@ -0,0 +1,46 @@
...
+
+alterQuery :: Text
+alterQuery = "ALTER TABLE \"custom_constraint2\" ADD CONSTRAINT \"custom_constraint\" FOREIGN KEY(\"cc_id\") REFERENCES \"custom_constraint1\"(\"id\")"
+
+specs :: (MonadIO m, MonadFail m) => RunDb SqlBackend m -> Spec
+specs runDb = do
+  describe "custom constraint used in migration" $ do
+    it "alter table uses custom constraint name" $ runDb $ do
+      ms <- getMigration customConstraintMigrate
+      let alter = find (== alterQuery) ms
```
robbassi: This seems a bit hacky, but I couldn't think of a better way to ensure that the migration uses the correct constraint name.

@parsonsmatt Thoughts?

robbassi: I ended up removing this test case as it makes the tests non-idempotent. Now we just ensure the constraint exists after running the migration.

persistent/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for persistent
 
+## 2.10.5
+
+* Added support for the `constraint=` attribute to the Postgresql backend. [#979](https://github.com/yesodweb/persistent/pull/979)
+
```
parsonsmatt: This should go in `persistent-postgresql/changelog.md`

persistent/persistent.cabal
```diff
@@ -1,5 +1,5 @@
 name:            persistent
-version:         2.10.4
+version:         2.10.5
```
parsonsmatt: Should go in `persistent-postgresql/persistent-postgresql.cabal`

parsonsmatt:
Once CI is passed I'll merge and release. Thanks so much!

robbassi:
Awesome. Thank you for the prompt replies!

MaxGabriel:
Oh, heads up that this change seems to break certain foreign key setups. Still investigating, but for example for this table:

```
mercury-web-backend-test=# \d+ synapse_recipient_ach_nodes
                                          Table "public.synapse_recipient_ach_nodes"
          Column           |           Type           | Collation | Nullable | Default | Storage  | Stats target | Description
---------------------------+--------------------------+-----------+----------+---------+----------+--------------+-------------
 recipient_payment_data_id | uuid                     |           | not null |         | plain    |              |
 bank_account_id           | uuid                     |           | not null |         | plain    |              |
 ach_node_id               | text                     |           | not null |         | extended |              |
 created_at                | timestamp with time zone |           | not null |         | plain    |              |
 updated_at                | timestamp with time zone |           | not null |         | plain    |              |
Indexes:
    "synapse_recipient_ach_nodes_pkey" PRIMARY KEY, btree (recipient_payment_data_id, bank_account_id, ach_node_id)
Foreign-key constraints:
    "synapse_recipient_ach_nodes_ach_node_id_fkey" FOREIGN KEY (ach_node_id) REFERENCES synapse_ach_nodes(node_id)
    "synapse_recipient_ach_nodes_bank_account_id_fkey" FOREIGN KEY (bank_account_id) REFERENCES bank_accounts(id)
    "synapse_recipient_ach_nodes_recipient_payment_data_id_fkey" FOREIGN KEY (recipient_payment_data_id) REFERENCES recipient_payment_data(id)
    "synapse_recipient_ach_nodes_recipient_payment_data_id_fkey1" FOREIGN KEY (recipient_payment_data_id, bank_account_id) REFERENCES recipient_accounts(recipient_payment_data_id, bank_account_id)
Triggers:
    synapse_recipient_ach_nodes_insert BEFORE INSERT ON synapse_recipient_ach_nodes FOR EACH ROW EXECUTE PROCEDURE create_timestamps()
    synapse_recipient_ach_nodes_update BEFORE UPDATE ON synapse_recipient_ach_nodes FOR EACH ROW EXECUTE PROCEDURE update_timestamps()
```

the query returns:

```
[[PersistText "recipient_accounts",PersistText "synapse_recipient_ach_nodes_recipient_payment_data_id_fkey1"],[PersistText "recipient_payment_data",PersistText "synapse_recipient_ach_nodes_recipient_payment_data_id_fkey"]]
```

robbassi:
@MaxGabriel Good catch, I should be able to look into this tonight.

jkeuhlen:
@robbassi Were you able to look into it? If not, it's on my list today and I'm happy to dig in/help take a look.

robbassi:
@jkeuhlen Ok great. I'm planning to work on it today, but any help would be appreciated!

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -792,23 +799,27 @@ getColumn getter tableName' [PersistText columnName, PersistText isNullable, Per
...
-            Just [PersistInt64 i] <- CL.head
-            return $ if i == 0 then Nothing else Just (DBName "", ref))
+        cntrs <- with (stmtQuery stmt [PersistText $ unDBName tableName'
+                                      ,PersistText $ unDBName cname])
+                      (\src -> runConduit $ src .| CL.consume)
+        case cntrs of
+          [] -> return Nothing
+          [[PersistText table, PersistText constraint]] ->
+            return $ Just (DBName table, DBName constraint)
+          _ -> error "Postgresql.getColumn: error fetching constraints"
```
jkeuhlen: @robbassi So this is the place where the error is thrown. Is there a reason you assume a single constraint per table here and throw the error? A simple idea of a schema that this breaks is a table that records sales like: 
```
Sales 
  buyer PersonId 
  seller PersonId

Person
  name Text
```

robbassi:
@jkeuhlen I beleive the issue here is with multiple foreign key contraints for a single column, not the table. This was actually ported from the MySQL implementation (#937) for custom constraint names. I  just verified that this issue also exists in the MySQL backend. 

To fix this we'll need to come up with a better way of fetching the columnm metadata when building migrations, let me know if you have any ideas.

jkeuhlen:
@robbassi Yeah that makes sense. I was just writing a test for this case, and it was surprised that it passed (since I was testing the table not the column). Would it be sufficient to change the definition of `Column`? 

```
module Database.Persist.Sql.Types
    ( module Database.Persist.Sql.Types
    , SqlBackend (..), SqlReadBackend (..), SqlWriteBackend (..)
    , Statement (..), LogFunc, InsertSqlResult (..)
    , readToUnknown, readToWrite, writeToUnknown
    , SqlBackendCanRead, SqlBackendCanWrite, SqlReadT, SqlWriteT, IsSqlBackend
    ) where

import Control.Exception (Exception)
import Control.Monad.Logger (NoLoggingT)
import Control.Monad.Trans.Reader (ReaderT (..))
import Control.Monad.Trans.Resource (ResourceT)
import Control.Monad.Trans.Writer (WriterT)
import Data.Pool (Pool)
import Data.Text (Text)
import Data.Typeable (Typeable)

import Database.Persist.Types
import Database.Persist.Sql.Types.Internal

data Column = Column
    { cName      :: !DBName
    , cNull      :: !Bool
    , cSqlType   :: !SqlType
    , cDefault   :: !(Maybe Text)
    , cDefaultConstraintName   :: !(Maybe DBName)
    , cMaxLen    :: !(Maybe Integer)
    , cReference :: !(Maybe (DBName, DBName)) -- table name, constraint name
    }
    deriving (Eq, Ord, Show)
```

Specifically, the column reference could be changed to `[(DBName, DBName)]`. I'm not familiar enough with persistent to know the full ramifications of that decision though.

robbassi:
@jkeuhlen I think something like that should work. Though, I'm trying to better understand the use case here. I don't believe persistent can generate tables like this through migrations. Which leads me to believe that this was some pre-exisiting table that @MaxGabriel was trying to run a migration against. @MaxGabriel could you let us know if this table was created through a persistent migration?

jkeuhlen:
oh @robbassi I work with @MaxGabriel so I can confirm (and test with his original issue). We aren't creating any of these through the persistent migrations. But we do check our table definitions against the models with persistent and that's where we run into the problem.

robbassi:
@jkeuhlen Ok cool. So when you say you check the table definitions against the model, is that essentially running a migration and seeing what changes?

jkeuhlen:
@robbassi Exactly, we run a migration and capture the output, then do some tweaking internally since we know of a few mismatches, then fail if the generated migrations don't match the models (minus our expected mismatches).

jkeuhlen:
And the failure occurs during migration generation at the line I pointed out here: https://github.com/yesodweb/persistent/pull/979/files#r363880641

robbassi:
@jkeuhlen Ok got it. I'm going through the code now to see if there is an easier fix, otherwise I'll look into modifying the `Column` definition itself.

jkeuhlen:
@robbassi Awesome, thanks! Let me know if I can help in any way (failing test cases maybe?).

robbassi:
@jkeuhlen That would be great! I created a branch for the fix here: [robbassi/persistent/tree/fix-migration-multiple-fk](https://github.com/robbassi/persistent/tree/fix-migration-multiple-fk)

jkeuhlen:
@robbassi  So I have a pretty reliable way to make things break, but I don't really understand why they are breaking. 

Using this schema:
```haskell
share [mkPersist sqlSettings, mkMigrate "customConstraintMigrate"] [persistLowerCase|
CustomConstraint1
    some_field Text
    deriving Show

CustomConstraint2
    cc_id CustomConstraint1Id constraint=custom_constraint
    deriving Show

CustomConstraint3
    cc_id1 CustomConstraint1Id constraint=cc_id1_custom_constraint1
    cc_id2 CustomConstraint1Id constraint=cc_id2_custom_constraint2
    deriving Show
|]
```

This all seems to work fine actually. However, when the constraints aren't specified in the schema, things go awry. In this case, I just comment out the `constraint=` syntax and run the generated migrations manually:

```haskell
    it "allows constraints not expressed in persistent" $ runDb $ do
      runMigration customConstraintMigrate
      rawExecute "ALTER TABLE \"custom_constraint3\" ADD CONSTRAINT \"cc_id1_custom_constraint1\" FOREIGN KEY(\"cc_id1\") REFERENCES \"custom_constraint1\"(\"id\")" []
      rawExecute "ALTER TABLE \"custom_constraint3\" ADD CONSTRAINT \"cc_id2_custom_constraint1\" FOREIGN KEY(\"cc_id2\") REFERENCES \"custom_constraint1\"(\"id\")" []
      vals <- getMigration customConstraintMigrate
      liftIO $ print vals
      pure ()
```

Leads to: 
```
       uncaught exception: ErrorCall
       Postgresql.getColumn: error fetching constraints. Expected a single table and a single constraint for table: custom_constraint3 and a column: cc_id1 but got: [[PersistText "custom_constraint1",PersistText "cc_id1_custom_constraint1"],[PersistText "custom_constraint1",PersistText "custom_constraint3_cc_id1_fkey"]]
       CallStack (from HasCallStack):
         error, called at ./Database/Persist/Postgresql.hs:823:13 in persistent-postgresql-2.10.1.1-JVkLBlDLwV8GsrinRLi7sQ:Database.Persist.Postgresql
```

Any idea what's going on here? It seems like the manual `ALTER TABLE` command is generating an extra `custom_constraint3_cc_id1_fkey` but I don't know where that one comes from. I'm also just running the output of the migration that persistent generated in the first place.

robbassi:
@jkeuhlen It looks like `custom_constraint3_cc_id1_fkey` is referencing `cc_id1` from `custom_constraint3`. It's possible it was left over from another migration. If you drop it manually does it come back?

jkeuhlen:
@robbassi Yeah, it still shows up with an entirely fresh DB (I dropped `test`, recreated it, and ran only this test case).

jkeuhlen:
@robbassi `custom_constraint3_cc_id1_fkey` is the default constraint name that you get if you just do `constraint` instead of `constraint=`, so I wonder if that's where it comes from?

robbassi:
@jkeuhlen Ah, so if you changed:

```
CustomConstraint3
    cc_id1 CustomConstraint1Id constraint=cc_id1_custom_constraint1
    cc_id2 CustomConstraint1Id constraint=cc_id2_custom_constraint2
    deriving Show
```

to something like:

```
CustomConstraint3
    cc_id1 CustomConstraint1Id
    cc_id2 CustomConstraint1Id constraint=cc_id2_custom_constraint2
    deriving Show
```

That would explain it.

jkeuhlen:
@robbassi Yup. I realized after the fact that persistent adds the first foreign key automatically. I think this test case is a good illustration of the problem then, even if it is a trivial example. I'll PR it over to your branch.

The more complicated example would be like the one Max mentioned before where you have two different foreign keys on the same field because one is in a compound key.

robbassi:
@jkeuhlen I updated my branch with the fix. I may still clean things up a bit, but the issue should be fixed now. I verified your test case passes now, but could you try out the original use case just to make sure we caught everything?

jkeuhlen:
@robbassi The original use case passes now, thanks so much for fixing this!

robbassi:
@jkeuhlen Glad I could help! I'll make a PR for this shortly, would you mind creating an issue for this? Just want to make sure this gets tracked properly.

PR #980, opened 2019-11-04
Changed persistent-mysql to use 'utf8mb4' instead of 'utf8' in migrations

charukiewicz:
As stated in title, this changes migration behavior for the `VARCHAR` and `TEXT` column types to use the `uft8mb4` character set as opposed to `uft8`. This is because [`utf8` is an alias for `utf8mb3` which does not properly handle all UTF-8 strings](https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8.html). Here's another [StackOverflow answer that discusses this](https://stackoverflow.com/questions/202205/how-to-make-mysql-handle-utf-8-properly).

Not sure if this is worth a version number bump, as `utf8mb4` is fully backwards compatible. Let me know if I should bump the version.

-------

After submitting your PR:


parsonsmatt:
I'm happy with this change. Is it a drop-in 100% improvement with no possible breaking change behavior?

charukiewicz:
As far as I know, yes. `utf8mb4` is a superset of what MySQL calls `utf8`, which is an alias for `utf8mb3`. This means a maximum of 4 bytes instead of 3 bytes. Everything else will work the same way.

parsonsmatt:
awesome! I'll try to get this released this weekend.

parsonsmatt:
Released as `persistent-2.10.2.1`

Issue #981, opened 2019-11-04
Absolute lack of thread-safety `SqlBackend` should be documented.


merijn:
Right now `SqlBackend` has almost no documentation, there should be some big, bright warning saying "No way, no how can you ever under any conditions assume that using a single SqlBackend is remotely safe to use from separate threads"

I've been bitten by cryptic errors in the past and only NOW by browsing the source do I realise how utterly unsafe SqlBackend is in this regard.

parsonsmatt:
Dang. That's a surprising thing to hear! thanks for reporting it.

merijn:
@parsonsmatt the issue boils down most (all?) query functions being build on top of `rawQueryRes`, which in turn uses [`getStmtConn`](https://github.com/yesodweb/persistent/blob/master/persistent/Database/Persist/Sql/Raw.hs#L75-L103), which as you can see maintains a map of `Statement` in an `IORef` indexed by the query text. This means that all queries with the same text will map to a single `Statement` that is used for stepping and preparing. This means if any threads ever run the same query they will interfere with each other's preparation and stepping and generally just cause arbitrary failures due to racing eachother.

merijn:
I'm actually somewhat baffled what prompted this design in the first place...

parsonsmatt:
Hell, you're entirely right.

I think I get the intent - we want to prepare a statement once, then execute it multiple times. But sharing it across multiple threads is problematic.

Gating access behind an `MVar` or `TVar` would introduce locking/blocking on queries, which also seems bad.

Changing it to `Map (ThreadId, Text) (IORef Statement)` would be a relatively easy change to make and should stop 'bad' access patterns:

```diff
- pure $ Map.lookup sqlText stmtMap
+ threadId <- myThreadId
+ pure $ Map.lookup (threadId, sqlText) stmtMap
```

This fixes the multi-threaded access problem, but it isn't obvious to me that this is "good" in the initial design goals for the `SqlBackend` to store a map of `Statement`s.

Could aslo have:

```diff
- Map Sql (IORef Statement)
+ Map Sql (IO Statement)
```

where the `IO` action has a way of mapping the `Statement`s to appropriate `ThreadId`s.

merijn:
Ok, bad news first:

[...]

PR #982, opened 2019-11-05
Make thread-safe wrapper around `getStmtConn`.

merijn:
Refer to #981.

### DIFF
--- a/persistent/Database/Persist/Sql/Raw.hs
+++ b/persistent/Database/Persist/Sql/Raw.hs
@@ -1,14 +1,17 @@
 module Database.Persist.Sql.Raw where
 
-import Control.Exception (throwIO)
-import Control.Monad (when, liftM)
+import Control.Concurrent.MVar
+import Control.Exception (mask_, throwIO)
+import Control.Monad (join, when, liftM)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Control.Monad.Logger (logDebugNS, runLoggingT)
 import Control.Monad.Reader (ReaderT, ask, MonadReader)
 import Control.Monad.Trans.Resource (MonadResource,release)
 import Data.Acquire (allocateAcquire, Acquire, mkAcquire, with)
 import Data.Conduit
-import Data.IORef (writeIORef, readIORef, newIORef)
+import Data.Foldable (traverse_)
+import Data.IORef (IORef, atomicModifyIORef', readIORef, newIORef)
+import Data.Map (Map)
 import qualified Data.Map as Map
 import Data.Int (Int64)
 import Data.Text (Text, pack)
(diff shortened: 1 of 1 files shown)


merijn:
Running concurrent queries is threadsafe and works with this PR, but concurrent inserts result in utter garbage due to the two stage ROWID based method of returning the key that was just inserted (for SQLite and presumably the same of MySQL).

parsonsmatt:
Thanks for all the work on this stuff. I'll try to get to it soon!

parsonsmatt:
This change in behavior is complicated enough that I'd want to gate it behind a major version bump, which means I also want to collect other breaking changes into a big release.

I could be convinced to release it as a patch version, but only with sufficient testing or correctness guarantees. I'd *hate* to release a version and have someone's application deadlock unexpectedly. However, if there application would have otherwise blown up or done something bad, then I might be okay with it too.

merijn:
Yeah, this PR was/is more of a proof-of-concept/starting point

parsonsmatt:
@merijn Now that it's the new year, how are you feeling about this? I've reviewed it a few times and I feel good about the change.

merijn:
I think this one is still a bit dicy, since as I commented before concurrent inserts basically corrupt the "safe" keys you get back. I think it can be made to work, but it might require overhauling the way inserts are done in all the SQL backend, so that's quite some work.

parsonsmatt:
@merijn Do you think you'll have time to do this in the next week or two?

I'm happy to accept an improvement to the status quo, even if it doesn't completely solve the problem, provided that the remaining issues can be documented.

ygale:
I am opposed to merging this PR.

This is a breaking change for all persistent users to implement a backend-specific feature. Please implement this as part of the SQLite backend, not as part of persistent.

merijn:
This PR (as is) can't be merged, since it doesn't actually work on any backend and I don't have time to get things working anytime soon. That said, this is also wrong:

> to implement a backend-specific feature.

I haven't tested to confirm, but I can't see a way the current implementation isn't broken for *all* persistent backends. Now, admittedly the other backends rely on connection pools more, so are less likely to run into breakage as they don't run into issues with concurrent queries from separate threads. However, that's not *all* that's broken, running multiple queries in a conduit from the same thread is also horribly broken, and half the `ConduitT` based API in `persistent` will happily leak unfinished queries wasting memory in the database.

ygale:
You are referring to the `selectSource` family of methods, I suppose. Hmm, those are indeed problematic. But the prepared statement cache is not the issue for most backends, and changing it won't help. There needs to be a way to block re-use of a connection until the active query finishes. This is not a simple problem.

codygman:
> There needs to be a way to block re-use of a connection until the active query finishes. This is not a simple problem.

My team is affected by this and it results in the postgres error:

```
libpq: failed (another command is already in progress
```

From code like:

```haskell
withDB1
  . Conduit.runConduit
  $ (Persist.selectSource [] [] .| Conduit.mapM_
      (\(userExternalId, loginCount) ->
        runMyMonadT env . withDB2 . void $ Persist.upsert myModel)
    )
```

PR #983, opened 2019-11-05
Add pool creation functions for `RawSqlite`.

merijn:
Before submitting your PR, check that you've:

- [x] Bumped the version number
- [x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

- [x] Update the Changelog.md file with a link to your PR
- [x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

PR #984, opened 2019-11-06
Add Acquire based interfaces for connections.

merijn:
Add Acquire based interface for starting transactions and getting
connections from a Pool for use in MonadResource instances that are not
MonadUnliftIO, such as Conduit.

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Sql/Run.hs
+++ b/persistent/Database/Persist/Sql/Run.hs
@@ -23,11 +23,18 @@ import Database.Persist.Sql.Types
 import Database.Persist.Sql.Types.Internal (IsolationLevel)
 import Database.Persist.Sql.Raw
 
-rawAcquireSqlConnFromPool
+-- | The returned 'Acquire' gets a connection from the pool, but does __NOT__
+-- start a new transaction. Used to implement 'acquireSqlConnFromPool' and
+-- 'acquireSqlConnFromPoolWithIsolation', this is useful for performing actions
+-- on a connection that cannot be done within a transaction, such as VACUUM in
+-- Sqlite.
+--
+-- @since 2.10.5
+unsafeAcquireSqlConnFromPool
     :: forall backend m
      . (MonadReader (Pool backend) m, BackendCompatible SqlBackend backend)
     => m (Acquire backend)
-rawAcquireSqlConnFromPool = do
+unsafeAcquireSqlConnFromPool = do
     pool <- MonadReader.ask
 
     let freeConn :: (backend, LocalPool backend) -> ReleaseType -> IO ()
@@ -56,7 +63,7 @@ acquireSqlConnFromPool
     :: (MonadReader (Pool backend) m, BackendCompatible SqlBackend backend)
     => m (Acquire backend)
 acquireSqlConnFromPool = do
-    connFromPool <- rawAcquireSqlConnFromPool
+    connFromPool <- unsafeAcquireSqlConnFromPool
     return $ connFromPool >>= acquireSqlConn
 
 -- | Like 'acquireSqlConnFromPool', but lets you specify an explicit isolation
@@ -67,7 +74,7 @@ acquireSqlConnFromPoolWithIsolation
     :: (MonadReader (Pool backend) m, BackendCompatible SqlBackend backend)
     => IsolationLevel -> m (Acquire backend)
 acquireSqlConnFromPoolWithIsolation isolation = do
-    connFromPool <- rawAcquireSqlConnFromPool
+    connFromPool <- unsafeAcquireSqlConnFromPool
     return $ connFromPool >>= acquireSqlConnWithIsolation isolation
 
 -- | Get a connection from the pool, run the given action, and then return the
--- a/persistent/Database/Persist/Sql/Run.hs
+++ b/persistent/Database/Persist/Sql/Run.hs
@@ -6,9 +6,13 @@ import Control.Monad (liftM)
 import Control.Monad.IO.Unlift
 import qualified UnliftIO.Exception as UE
 import Control.Monad.Logger.CallStack
+import Control.Monad.Reader (MonadReader)
+import qualified Control.Monad.Reader as MonadReader
 import Control.Monad.Trans.Reader hiding (local)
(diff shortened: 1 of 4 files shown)


merijn:
The LTS-9 looks to be due to a different with being in scope/`resourcet` having a different type in that version?

I don't get the MySQL error in nightly, seems to be unrelated?

merijn:
LTS-9 failure is indeed due to ResourceT not being based on `MonadUnliftIO` in that version, not really sure how we could work around that.

MySQL error in nightly appears to be a parse error in timestamps (which doesn't seem remotely related to anything I changed).

parsonsmatt:
> MySQL error in nightly appears to be a parse error in timestamps (which doesn't seem remotely related to anything I changed).

I suspect that it's because the `instance Read UTCTime` became more strict, requiring the `Z` suffix. I had to fix similar errors in `esqueleto` recently.

> LTS-9 failure is indeed due to ResourceT not being based on MonadUnliftIO in that version, not really sure how we could work around that.

`lts-9` is GHC 8.0.2, which is out of the "support 3 major versions" release policy. I'd be OK with dropping support for this.

A compatibility module that had some CPP and used `MonadBaseControl`.. would be disgusting but could work.

merijn:
> A compatibility module that had some CPP and used MonadBaseControl.. would be disgusting but could work.

Not sure that'd even work, because this is part of the visible persistent API, since there's no way to guarantee that every `MonadBaseControl IO` instance has a corresponding `MonadUnliftIO` and vice versa, afaict.

merijn:
@parsonsmatt What would dropping lts-9 support entail, just deleting the lts-9 line from `.travis.yml`?

I'd like to see this and #983 merged and making it onto Hackage (although, since I'd also like to bump the SQLite version shipped, perhaps a new Hackage release should wait until I open a PR for that).

merijn:
@parsonsmatt Ping?

I'm not really sure what dropping lts-9 support entails beyond deleting the tests from `.travis.yml`, additionally I'd (ideally) like to see #983 and #991 merged and released before the busy holiday season starts :)

persistent/Database/Persist/Sql/Run.hs
```diff
@@ -28,15 +86,16 @@ import Database.Persist.Sql.Raw
 runSqlPool
     :: (MonadUnliftIO m, BackendCompatible SqlBackend backend)
     => ReaderT backend m a -> Pool backend -> m a
-runSqlPool r pconn = withRunInIO $ \run -> withResource pconn $ run . runSqlConn r
+runSqlPool r pconn = with (acquireSqlConnFromPool pconn) $ runReaderT r
```
parsonsmatt: The change from `runSqlConn` to `runReaderT` drops the logic around transactions and exception handling. Why'd this get changed?

```suggestion
runSqlPool r pconn = with (acquireSqlConnFromPool pconn) $ runSqlConn r
```

persistent/Database/Persist/Sql/Run.hs
```diff
@@ -28,15 +86,16 @@ import Database.Persist.Sql.Raw
...
 
 -- | Like 'runSqlPool', but supports specifying an isolation level.
 --
 -- @since 2.9.0
 runSqlPoolWithIsolation
     :: (MonadUnliftIO m, BackendCompatible SqlBackend backend)
     => ReaderT backend m a -> Pool backend -> IsolationLevel -> m a
-runSqlPoolWithIsolation r pconn i = withRunInIO $ \run -> withResource pconn $ run . (\conn -> runSqlConnWithIsolation r conn i)
+runSqlPoolWithIsolation r pconn i =
+    with (acquireSqlConnFromPoolWithIsolation i pconn) $ runReaderT r
```
parsonsmatt: Same here - `runSqlConn` does all the transaction stuff, so `runReaderT` here voids that safety. (Man I really need to `newtype` SqlPersistT ...)

```suggestion
    with (acquireSqlConnFromPoolWithIsolation i pconn) $ \conn -> runSqlConnWithIsolation r conn i
```

parsonsmatt: Wait, no, I think I see what's going on here. You've moved that responsibility into the `Acquire` interface. So `runSqlConn` has it's own logic for handling exceptions, and that logic is now sort of duplicated into the `Acquire` interface.

This is fine. Ignore the suggestions. I've reviewed the relevant code and as far as I can tell the behavior will be identical.

persistent/Database/Persist/Sql/Run.hs
```diff
@@ -60,30 +119,62 @@ withResourceTimeout ms pool act = withRunInIO $ \runInIO -> mask $ \restore -> d
...
-runSqlConn r conn = withRunInIO $ \runInIO -> mask $ \restore -> do
-    let conn' = projectBackend conn
-        getter = getStmtConn conn'
-    restore $ connBegin conn' getter Nothing
-    x <- onException
-            (restore $ runInIO $ runReaderT r conn)
-            (restore $ connRollback conn' getter)
-    restore $ connCommit conn' getter
-    return x
+runSqlConn r conn = with (acquireSqlConn conn) $ runReaderT r
```
parsonsmatt: :joy: teaches me to review a PR before at least skimming it fully

parsonsmatt:
@merijn OK, we've dropped support for `lts-9` in the latest master - can you merge that in and get CI green? Then we can get this + your other SQLite PR out the door.

Thanks!

merijn:
@parsonsmatt hmmm, mongodb seems to fail for some reason?

merijn:
Which seems weird, as it works in your master and the failure seems to be caused by MonadFail, which I certainly didn't touch...

parsonsmatt:
yeah the `mongodb` library probably just changed and since we don't have an upper bound it is going bonkers.

merijn:
> yeah the `mongodb` library probably just changed and since we don't have an upper bound it is going bonkers.

I can make another PR with an upperbound on mongodb to get the tests passing and rebase on top of that...

parsonsmatt:
Released in `persistent-2.10.5`, thanks so much! :smile:

PR #996, opened 2019-12-24
Handle foreign key constraint names over 63 characters for Postgres

MaxGabriel:
Currently, persistent-postgresql suggests foreign key constraint names that are over Postgres' length limit of 63 bytes. Postgres helpfully truncates these names, but this causes persistent-postgresql to think the migration to create the constraint has not been applied, and keep recommending that the migration be applied.

To solve this issue, this PR has persistent-postgresql truncate foreign key constraint names using the same algorithm that Postgres itself uses.

MySQL should also have something like this, but for a 64 character limit. It could use the same one Postgres does, or custom logic for itself.

<hr>

- [x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html) — I did not do this, but because the new APIs are in an `Internal` module I'm not sure this is necessary? They're only meant to be used by Persistent backends, is my understanding.
- [x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock — I did not do this, but because the new APIs are in an `Internal` module I'm not sure this is necessary? They're only meant to be used by Persistent backends, is my understanding.

After submitting your PR:


parsonsmatt:
> I did not do this, but because the new APIs are in an Internal module I'm not sure this is necessary? They're only meant to be used by Persistent backends, is my understanding.

I'd prefer to add documentation and `@since` annotations anyway - while most users of the library are working with database apps, some users are implementing database backends, and we should support them as well.

Ideally I'd like to factor out the internal stuff into a `persistent-core` which can evolve separately rather than have the "Internal changes don't count(ish)" policy we have now.

persistent-mysql/test/main.hs
```diff
@@ -181,6 +183,9 @@ main = do
 
     MigrationIdempotencyTest.specsWith db
     CustomConstraintTest.specs db
+    -- MySQL currently fails this test, because it has a limit of 64 characters for identifiers
+    -- TODO: implement automatic truncation for too long foreign keys
+    -- LongIdentifierTest.specsWith db
```
parsonsmatt: Instead of commenting this out, please use a [`pendingWith`](https://www.stackage.org/haddock/lts-14.19/hspec-2.7.1/Test-Hspec.html#v:pendingWith) or [`xdescribe`](https://www.stackage.org/haddock/lts-14.19/hspec-2.7.1/Test-Hspec.html#v:xdescribe) call to make this more clear at runtime :)

```suggestion
    xdescribe "MySQL currently fails because of 64 character limit for identifiers (# issue number)" $
        LongIdentifierTest.specsWith db
```

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE MultiWayIf #-}
```
parsonsmatt: I'd rather not introduce this extension - the use-site doesn't require it.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1151,6 +1153,38 @@ instance PersistConfig PostgresConf where
...
+      -- I believe this will also be an issue for extremely long table names, 
+      -- but it's just much more likely to exist with foreign key constraints because they're usually tablename * 2 in length
+
+      -- Approximation of the algorithm Postgres uses to truncate identifiers
+      -- See makeObjectName https://github.com/postgres/postgres/blob/5406513e997f5ee9de79d4076ae91c04af0c52f6/src/backend/commands/indexcmds.c#L2074-L2080
+      shortenNames :: Int -> (Int, Int) -> (Int, Int)
+      shortenNames overhead (x, y) =
+        if | x + y + overhead <= maximumIdentifierLength -> (x, y)
+           | x > y -> shortenNames overhead (x - 1, y)
+           | otherwise -> shortenNames overhead (x, y - 1)
```
parsonsmatt: This can be accomplished with ordinary guards.

```haskell
shortenNames overhead (x, y)
    | x + y + overhead <= maximumIdentifierLenght = (x, y)
    | x > y = shortenNames overhead (x - 1, y)
    | otherwise = shortenNames overhead (x, y - 1)
```

persistent/Database/Persist/Sql/Internal.hs
```diff
@@ -1,8 +1,11 @@
 {-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE RecordWildCards #-}
```
parsonsmatt:
```suggestion
{-# LANGUAGE RecordWildCards #-}

```

persistent/Database/Persist/Sql/Internal.hs
```diff
@@ -52,13 +63,15 @@ mkColumns allDefs t =
...
+    refNameFn = fromMaybe refName (backendSpecificForeignKeyName overrides)
+
     ref :: DBName
         -> ReferenceDef
         -> [Attr]
         -> Maybe (DBName, DBName) -- table name, constraint name
     ref c fe []
         | ForeignRef f _ <- fe =
-            Just (resolveTableName allDefs f, refName tn c)
+            Just (resolveTableName allDefs f, refNameFn tn c)
```
parsonsmatt: Ah! We have been ignoring the user-provided name of the foreign key reference this entire time?? Dang. Welp. Fixing that is out-of-scope for this PR.

MaxGabriel: Oh, hah, didn’t even notice

MaxGabriel:
Ok, I think this is good to go (?)

parsonsmatt:
Since this is a breaking change for `persistent`, I want to see if there is another change we can roll-up with this one.

persistent/Database/Persist/Sql/Internal.hs
belevy: maybe the constraint name override can be pushed to master now since that is a backwards compatible workaround. We only implemented it for MySQL since that is the backend we use and were running into the length issue but didn't have the resources to dedicate towards a proper solution.

parsonsmatt:
Alright, I'm getting `persistent-2.11.0.0` ready for release. This is a great candidate. Mind getting it up-to-date with the master branch?

MaxGabriel:
Oh thanks for getting this in! Sorry I often miss github notifications, feel free to ping me on FPChat slack if I don't respond

PR #998, opened 2019-12-28
Remove overlapping Lift instance

RyanGlScott:
This patch removes the overlapping `instance Lift' a => Lift a` by replacing the three `Lift'` instances with `Lift` instances:

* There has been an `instance Lift a => Lift [a]` in `template-haskell` for a while now, so that one is already covered.
* The `instance (Lift k, Lift v) => Lift (Map k v)` is now obtained from the `th-lift-instances` library, which is the canonical home for that particular instance.
* The `text` library defines an `instance Lift Text` since version 1.2.4.0. On earlier versions of `text`, an equivalent instance is obtained from the `th-lift-instances` library, which backports it.

After submitting your PR:

### DIFF
--- a/persistent-template/ChangeLog.md
+++ b/persistent-template/ChangeLog.md
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 2.7.4
+
+* Remove an overlapping instance for `Lift a`. [#998](https://github.com/yesodweb/persistent/pull/998)
+
 ## 2.7.3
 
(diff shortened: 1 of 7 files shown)


parsonsmatt:
awesome, thanks!

RyanGlScott:
CI is fixed now (after tweaking several `stack.yaml` files).

MaxGabriel:
Released!

Issue #869, opened 2019-04-02
`TypeApplications` friendly types


parsonsmatt:
Many functions in `persistent` use types to determine which tables to talk to. Now that GHC 8 has widespread adoption, we should consider refactoring the types to be more friendly to type application syntax.

As an example, `deleteWhere` currently specifies the `m` first and the `rec` second, so deleting *all* records must be `deleteWhere @_ @SomeRecord`. `selectList` specifies the `m`, then `backend`, then `record`, so you must write `selectList @_ @_ @SomeRecord [] []` to load all rows from the database.

Reordering the type variables so that the `record` is always first would make the API better for `TypeApplications`.

Vlix:
This would just mean adding `forall record a b c.` in front of the type signatures, right?

parsonsmatt:
Pretty much, yeah

PR #1006, opened 2020-01-05
added explicit forall notation to make TypeApplications easier to use

Vlix:
This is a PR for issue #869 

---

After submitting your PR:


Vlix:
It's failing because of something in MongoDB?

parsonsmatt:
That's a really strange failure...

Vlix:
All builds using stack works no problem, but built with cabal, it breaks on a missing `MonadFail` constraint.

parsonsmatt:
I suspect we're missing an upper bound somewhere...

parsonsmatt:
Try merging master - there's an upper bound in place to prevent the cabal solver from picking that version.

persistent/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for persistent
 
+## 2.10.5
```
parsonsmatt: This is a breaking change and requires a major version bump
```suggestion
## 2.11.0
```

persistent/persistent.cabal
```diff
@@ -1,5 +1,5 @@
 name:            persistent
-version:         2.10.4
+version:         2.10.5
```
parsonsmatt:
```suggestion
version:         2.11.0
```

Vlix:
I'm trying to add some tests, but it's all really annoying.
Also, all of the methods from typeclasses have the `backend` type variable as number one, because that's what's in the typeclass definition. Don't really know if there'd be a way to make that more user friendly...

Vlix:
I guess one way to do it, would be to not export the class methods in `Database.Persist`, and make new functions that have the same name so anyone importing `Database.Persist` will be able to use the @-notation and anyone making their own (e.g. `PersistStoreWrite`) instance can use `Database.Persist.Class`

parsonsmatt:
I wouldn't worry about tests for this. It's not like it can be broken.

You'd only really want to specify the `rec` where it could be ambiguous, like `selectList` and `deleteWhere`. I wouldn't worry too much about duplicating the names. eg `insert` can't be used polymorphically so it wouldn't make sense to specify the type.

Vlix:
Well, I did anyway.

But yeah, `selectList` will have to still be used as `selectList @_ @Table`, since the `backend` parameter is first, because of the fact that it's a typeclass method. Same with `deleteWhere @_ @Table`. :/

parsonsmatt:
I think `selectList` is fine since it's not a class method. `selectFirst`, `count`, `selectSource` are all affected though :\ `PersistQueryWrite` is also affected, annoying.

Well.

Hmm! Maybe it is worthwhile to have the class methods separate from the functions in the API. It's a breaking change, but *all* of this is a breaking change anyway.

Vlix:
Ah true, `selectList` was outside of the typeclass.

Do you want me to add the `select*` and `delete*` functions in `Database.Persist`?

parsonsmatt:
looking forward to merging this <3

PR #1032, opened 2020-02-11
Remove the Natural instances

### LINKED ISSUE #1031: Natural values don't fit into an sql database
https://github.com/yesodweb/persistent/blob/6665283c837d66f649d4a13c48b273d109017bdd/persistent/Database/Persist/Sql/Class.hs#L493-L494

This fails to round-trip for naturals bigger than a Word64.
I would argue that this instance should be removed.
If you don't need naturals that big, then just use the Word64 instance instead.
If you do, then you expect them to roundtrip.

parsonsmatt:

After submitting your PR:

Fixes #1031

### DIFF
--- a/persistent/Database/Persist/Class/PersistField.hs
+++ b/persistent/Database/Persist/Class/PersistField.hs
@@ -1,11 +1,12 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE PatternGuards, DataKinds, TypeOperators, UndecidableInstances #-}
 module Database.Persist.Class.PersistField
     ( PersistField (..)
     , SomePersistField (..)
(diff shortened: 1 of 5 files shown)


erikd:
I should stop using it then :).

persistent/Database/Persist/Class/PersistField.hs
```diff
@@ -322,11 +324,43 @@ instance PersistField UTCTime where
...
+
+instance
+  TypeError
+    ( 'Text "The instance of PersistField for the Natural type was removed."
+    ':$$: 'Text "Please see the documentation for OverflowNatural if you want to "
+    ':$$: 'Text "continue using the old behavior or want to see documentation on "
+    ':$$: 'Text "why the instance was removed."
+    ':$$: 'Text ""
+    ':$$: 'Text "This error instance will be removed in a future release."
+    )
```
MaxGabriel: This is pretty sick

erikd: That is the good kind of sick right?

PR #1036, opened 2020-02-12
JSON instances require an extraneous id field for entities with a Primary

### LINKED ISSUE #1035: JSON instances require an extraneous id field for entities with a Primary
For a model definition:

```haskell
share [mkPersist sqlSettings] [persistLowerCase|
JsonEncoding json
    name Text
    age  Int
    Primary name
    deriving Show Eq
|]
```

This will generate an `instance {To,From}JSON (Entity JsonEncoding)` that requires an `"id"` field in the resulting object.

Serializing always puts a blank string.

parsonsmatt:

After submitting your PR:

Fixes #1035

### DIFF
--- a/persistent-test/src/PersistentTest.hs
+++ b/persistent-test/src/PersistentTest.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE RecordWildCards, UndecidableInstances #-}
+
 module PersistentTest
     ( module PersistentTest
     , cleanDB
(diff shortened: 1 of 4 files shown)


persistent-template/Database/Persist/TH.hs
```diff
@@ -443,8 +443,8 @@ data MkPersistSettings = MkPersistSettings
     --
     -- @
     --  Just EntityJSON
-    --      { entityToJSON = 'keyValueEntityToJSON
-    --      , entityFromJSON = 'keyValueEntityFromJSON
+    --      { entityToJSON = 'entityIdToJSON
+    --      , entityFromJSON = 'entityIdFromJSON
```
parsonsmatt: not sure when this was changed lol

Issue #776, opened 2018-01-28
Per entity field prefix?


gurgl:
I found you about mpsPrefixFields setting making it possible to control prefixing of fields by entity name. I wonder if its possible to prefix field accessors / lenses, per entity, with something else than entity name? Ive a bunch of types with accessors prefixed by "_" <acronym> "_". Like User's "name" field is _u_name and GameLog "command" field is _gl_command.

parsonsmatt:
We could support this by adding something like a `mpsFieldLabelModifier :: String -> String -> String` where the first parameter is the entity name.

PR #1044, opened 2020-02-15
First attempt at customising prefixes #776

olivierdeckers:
I gave #776 a try, and wanted to get some initial feedback.

This implementation doesn't interact very nicely with setting mpsPrefixFields to False. An empty entity name will then be passed to the function, which is surprising and limiting.
Maybe I could instead make the default implementation smarter to get around this? 

After submitting your PR:


persistent-test/src/MpsCustomPrefixTest.hs
```diff
@@ -0,0 +1,44 @@
...
+
+    -- mcp1a <- get cp1a
+    -- liftIO $ mcp1a @?= Just (CustomPrefix1 2)
+    -- liftIO $ fmap _customFieldName mcp1a @?= Just 2
+    -- mcp2 <- get cp2
+    -- liftIO $ fmap _customPrefixedRef mcp2 @?= Just cp1b
+    -- liftIO $ fmap _otherCustomFieldName mcp2 @?= Just 5
+
+    -- insert_ $ CustomPrefixedLeftSum 5
+    -- insert_ $ CustomPrefixedRightSum "Hello"
```
parsonsmatt: This is a great integration test for the feature! Thanks for writing it up.

persistent-test/src/PersistentTestModels.hs
```diff
@@ -133,8 +137,32 @@ deriving instance Eq (BackendKey backend) => Eq (NoPrefix1Generic backend)
 deriving instance Show (BackendKey backend) => Show (NoPrefix2Generic backend)
 deriving instance Eq (BackendKey backend) => Eq (NoPrefix2Generic backend)
 
-deriving instance Show (BackendKey backend) => Show (RelationshipGeneric backend)
-deriving instance Eq (BackendKey backend) => Eq (RelationshipGeneric backend)
+share [mkPersist persistSettings { 
+          mpsFieldLabelModifier = \entity field -> case entity of
+            "customPrefix1" -> append "_cp1" field
+            "customPrefix2" -> append "_cp2" field
```
parsonsmatt: It is - IMO - surprising that the `HaskellName` for the *table* is lowercased! I would assume that that parameter would be identical to what's on the table. One thing I love about tests is that they surface this kind of thing.

persistent-template/Database/Persist/TH.hs
```diff
@@ -490,9 +494,10 @@ sqlSettings = mkPersistSettings $ ConT ''SqlBackend
 
 recNameNoUnderscore :: MkPersistSettings -> HaskellName -> HaskellName -> Text
 recNameNoUnderscore mps dt f
-  | mpsPrefixFields mps = lowerFirst (unHaskellName dt) ++ upperFirst ft
-  | otherwise           = lowerFirst ft
+  | mpsPrefixFields mps = modifier (lowerFirst $ unHaskellName dt) (upperFirst ft)
```
parsonsmatt: I think that we could have a more consistent and reliable ordering if we did:

```
lowerFirst (modifier (unHaskellName dt) (upperFirst ft))
```

This also protects the user against providing an invalid field name. Consider if their modifier function was:

```haskell
\entity field -> "SchemaThree" <> entity <> field
```

For a `User; name String` model, this would make a field name of `SchemaThreeuserName`. Which isn't a valid Haskell identifier for a record field. But with that re-ordering, we'd get `schemaThreeUserName`, which is valid.

persistent-template/Database/Persist/TH.hs
```diff
@@ -437,6 +438,8 @@ data MkPersistSettings = MkPersistSettings
     -- False.
     , mpsPrefixFields :: Bool
     -- ^ Prefix field names with the model name. Default: True.
+    , mpsFieldLabelModifier :: Text -> Text -> Text
+    -- ^ Customise the field accessors and lens names using the entity and field name. Default: appends both
```
parsonsmatt: I'd prefer for things to just keep working as expected as much as possible. The corner case that you mention is that someone has `mpsPrefixFields = False`, *and* they go to use the `mpsFieldLabelModifier`. Obviously, they'd expect the entity name passed in with `mpsFieldLabelModifier` function, and not just empty string. 

I'm tempted to say that if they have specified `mpsPrefixFields = False`, then we just pass both things in to the function. The user is free to have a function like:

```haskell
mpsFieldLabelModifier = \entity field -> "_schema_x_" <> field
```

With this design, we don't know if the user has *provided* an implementation or is using the default. So perhaps we should set the type here to `mpsFieldLabelModifier ::Maybe (Text -> Text -> Text)` and a default of `Nothing`.

Then the logic looks like:

```haskell
case mpsFieldLabelModifier mps of
  Nothing -> do the logic as currently written
  Just fn -> provide both entity name and field to the function
```

This basically makes `mpsPrefixFields` redundant, in the case that the user specifies `mpsPrefixFields = False, mpsFieldLabelModifier = Just ...`. 

Another thing we'll want to consider: if they have `mpsPrefixFields = False`, do we want to uppercase the first character of the field name for the user? Whatever choice we make should be documented.

persistent-template/Database/Persist/TH.hs
olivierdeckers: Good idea, will fix

persistent-template/Database/Persist/TH.hs
olivierdeckers: I've thought about this some more. I make two observations:

1. We want the change to be backwards compatible for users using `mpsPrefixFields = False` who don't want to use the `mpsFieldLabelModifier`
2. Other than for this backward compatibility case, `mpsPrefixFields` is redundant since you can discard `entity` when specifying the `mpsFieldLabelModifier`.

I propose that we deprecate `mpsPrefixFields` and document that `mpsFieldLabelModifier` will be ignored if `mpsPrefixFields` is set to False. 
This way, we avoid complicating the type of the fieldLabelModifier.

parsonsmatt: That sounds good to me!

olivierdeckers: Observation 2 is not completely correct. mpsPrefixFields is also used for disabling the entity prefix in the constraint name, which should be uppercase and not start with an underscore.
One solution is to reuse the modifier function, drop any leading underscore and uppercase the result, but maybe an additional modifier function for the constraints would make more sense?

olivierdeckers: I went ahead and implemented a second modifier function (mpsConstraintLabelModifier) for the constraints which need to be upper case. Let me know what you think

persistent-template/Database/Persist/TH.hs
```diff
@@ -437,6 +439,24 @@ data MkPersistSettings = MkPersistSettings
     -- False.
     , mpsPrefixFields :: Bool
     -- ^ Prefix field names with the model name. Default: True.
+    --
+    -- Note: this field is deprecated. Use the mpsFieldLabelModifier  and mpsConstraintLabelModifier instead.
```
parsonsmatt: We can use a [`DEPRECATED` pragma](https://downloads.haskell.org/~ghc/7.2.1/docs/html/users_guide/pragmas.html#warning-deprecated-pragma) to make this official.

persistent-template/Database/Persist/TH.hs
```diff
@@ -437,6 +439,24 @@ data MkPersistSettings = MkPersistSettings
...
+    , mpsFieldLabelModifier :: Text -> Text -> Text
+    -- ^ Customise the field accessors and lens names using the entity and field name. 
+    -- Both arguments are upper cased. 
+    --
+    -- Default: appends entity and field.
+    --
+    -- Note: this setting is ignored if mpsPrefixFields is set to False.
+    -- @since 2.10.5.3
```
parsonsmatt: Technically, as a breaking change, this will be part of 2.11.0.0.

```suggestion
    -- @since 2.11.0.0
```

persistent-template/Database/Persist/TH.hs
```diff
@@ -437,6 +439,24 @@ data MkPersistSettings = MkPersistSettings
...
+    , mpsConstraintLabelModifier :: Text -> Text -> Text
+    -- ^ Customise the Constraint names using the entity and field name. The result
+    -- should be a valid haskell type (start with an upper cased letter).
+    --
+    -- Default: appends entity and field
+    --
+    -- Note: this setting is ignored if mpsPrefixFields is set to False.
+    -- @since 2.10.5.3
```
parsonsmatt: Technically, as a breaking change, this will be part of 2.11.0.0.

```suggestion
    -- @since 2.11.0.0
```

persistent/ChangeLog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog for persistent
 
+## 2.10.5.3
```
parsonsmatt:
```suggestion
## 2.11.0.0
```

persistent/persistent.cabal
```diff
@@ -1,5 +1,5 @@
 name:            persistent
-version:         2.10.5.2
+version:         2.10.5.3
```
parsonsmatt:
```suggestion
version:         2.11.0.0
```

Issue #910, opened 2019-05-22
Foreign Key Cascade


parsonsmatt:
Persistent users should be able to specify how delete/update work on foreign key references.

```
User
  name String
  UniqueName name

Dog
  user UserId OnDeleteCascade
```

parsonsmatt:
This is getting fixed in 2.11

PR #1056, opened 2020-03-19
Foreign key cascade

kderme:
I tried to fix https://github.com/yesodweb/persistent/issues/910. It works only for table constaints and not for field constraints. Also it only works and is tested on SQLite, but I think making postgres work is just a matter of copy-pase. No existing migrations break.

Some of the options, like 'OnDeleteSetNull' are not useful at the moment, because of the limitation of persistent to only make possible to refer to primary keys of other tables (which can't be NULL anyway).

Not to be confused with what `mkDeleteCascade` does. The approach here is to let the db do all the cascading changes, instead of manually updating other tables. 

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -272,6 +272,7 @@ import Data.Monoid (mappend)
 import Data.Text (Text)
 import qualified Data.Text as T
 import Database.Persist.Types
+import Text.Read (readEither)
 
 data ParseState a = PSDone | PSFail String | PSSuccess a Text deriving Show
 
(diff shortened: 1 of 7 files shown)


persistent/Database/Persist/Quasi.hs
```diff
@@ -908,28 +909,48 @@ takeForeign :: PersistSettings
...
-            , foreignAttrs =
-                attrs
-            , foreignNullable =
-                False
-            }
+takeForeign ps tableName _defs = takeRefTable
   where
-    (fields,attrs) = break ("!" `T.isPrefixOf`) rest
+    errorPrefix :: String
+    errorPrefix = "invalid foreign key constraint on table[" ++ show tableName ++ "] "
```
kderme:
```suggestion
    mkError :: String -> a
    mkError srt = error $ conat ["invalid foreign key constraint on table[", show tableName, "] ", str
```
Maybe this is cleaner

PR #1058, opened 2020-03-26
mongodb: Fix `ninList` filter operator

avdv:

After submitting your PR:

===

Currently, the `inList` and `ninList` definitions are the same, I think the latter should use the `NotIn` query operator.

### DIFF
--- a/persistent-mongoDB/persistent-mongoDB.cabal
+++ b/persistent-mongoDB/persistent-mongoDB.cabal
@@ -1,5 +1,5 @@
 name:            persistent-mongoDB
-version:         2.9.0.2
+version:         2.9.0.3
 license:         MIT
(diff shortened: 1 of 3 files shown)


persistent-mongoDB/persistent-mongoDB.cabal
```diff
@@ -1,5 +1,5 @@
 name:            persistent-mongoDB
-version:         2.9.0.2
+version:         2.9.0.3
```
parsonsmatt: Because we are changing the meaning of a function, it's going to have to be a major version bump.

```suggestion
version:         2.10.0.0
```

avdv:
Thank you @parsonmatt!

parsonsmatt:
Thanks! Released :)

PR #1060, opened 2020-03-30
Foreign key cascades

parsonsmatt:

After submitting your PR:

A continuation on #1056 implementing it in Postgresql.

We can't implement this for MySQL until we figure out how to set a `maxlen` on the keys for strings. :\ Well, we could, but we'd need a different set of tests, and I don't feel like writing that out right now. Might as well get sqlite and postgres supported and released.

### DIFF
--- a/persistent/Database/Persist/Types/Base.hs
+++ b/persistent/Database/Persist/Types/Base.hs
@@ -295,16 +295,28 @@ data ForeignDef = ForeignDef
     , foreignRefTableDBName        :: !DBName
     , foreignConstraintNameHaskell :: !HaskellName
     , foreignConstraintNameDBName  :: !DBName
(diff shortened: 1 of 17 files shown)


persistent/Database/Persist/Types/Base.hs
```diff
@@ -295,12 +295,39 @@ data ForeignDef = ForeignDef
     , foreignRefTableDBName        :: !DBName
     , foreignConstraintNameHaskell :: !HaskellName
     , foreignConstraintNameDBName  :: !DBName
+    , foreignFieldCascade          :: !FieldCascade
```
parsonsmatt: Adding a field to a data type is a breaking change, so this PR will be for the 2.11.0 release.

PR #1066, opened 2020-03-31
Allow sql=id to be declared for columns

### LINKED ISSUE #1059: `sql=id` breaks migrations
```
Test
    pk Int sql=id
    Primary pk
|]
```

This schema migrates correctly the first time.

However after the initial migration it keeps trying to run the following SQL command:

```
ALTER TABLE "test" ADD COLUMN "id" INT8 NOT NULL;
```

This of course throws an exception.

parsonsmatt:

After submitting your PR:

Fixes #1059


persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -202,18 +207,18 @@ withStmt' conn query vals = do
       let getters = [ maybe PersistNull (getGetter f f . Just) | f <- fields]
           convert = use getters
             where use (g:gs) (col:cols) =
-                    let v  = g col
-                        vs = use gs cols
-                    in v `seq` vs `seq` (v:vs)
+                    let !v  = g col
+                        !vs = use gs cols
+                    in (v:vs)
```
parsonsmatt: this is really like a strict `map` but I am not entirely sure where that lives?

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -331,41 +336,92 @@ migrate' :: MySQL.ConnectInfo
...
-                                                                     Just _ -> c { cReference = Nothing }
-                                                                     Nothing -> c
-                                                    Nothing -> c) xs,ys)
-            (acs, ats) = getAlters allDefs name (newcols, udspair) $ excludeForeignKeys $ partitionEithers old'
-            acs' = map (AlterColumn name) acs
-            ats' = map (AlterTable  name) ats
-        return $ Right $ map showAlterDb $ acs' ++ ats'
-      -- Errors
-      (_, _, (errs, _)) -> return $ Left errs
+    case ([], old, partitionEithers old) of
```
parsonsmatt: Sorry for the syntactic noise here but I just couldn't read this code. I formatted it and made it a bit more spaced out to make it easier to read.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -331,41 +336,92 @@ migrate' :: MySQL.ConnectInfo
...
+                        $ AlterTable name
+                        $ AddUniqueConstraint uname
+                        $ map (findTypeAndMaxLen name) ucols
+
+            let foreigns = do
+                    Column { cName=cname, cReference=Just (refTblName, refConstraintName) } <- newcols
+                    let refTarget =
+                          addReference allDefs refConstraintName refTblName cname
+
+                    guard $ refTblName /= name && cname /= fieldDB (entityId val)
```
parsonsmatt: Now that the `id` column is part of the columns returned by `mkColumn` we have to treat the reference separately. We don't want a self-reference on the ID column treated as a foreign key.

persistent-mysql/test/main.hs
```diff
@@ -119,7 +119,7 @@ main = do
       , Recursive.recursiveMigrate
       , CompositeTest.compositeMigrate
       , PersistUniqueTest.migration
-      , RenameTest.migration
+      -- , RenameTest.migration
```
parsonsmatt: Currently broken on mysql because `DEFAULT=CURRENT_DATE` *is not supported on mysql* ???

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -622,29 +624,39 @@ mkForeignAlt name fdef =
...
+        , escape name
+        , "("
+        , idtxt
+        , if null nonIdCols then "" else ","
+        , T.intercalate "," $ map showColumn nonIdCols
+        , ")"
+        ]
+  where
+    nonIdCols =
+        filter (\c -> cName c /= fieldDB (entityId entity) ) cols
```
parsonsmatt: We filter out the id column here,because it is treated specially in `idtxt`. And unfortunately it would be pretty difficult/annoying to set a default at the `Column` level because that bit isn't database agnostic.

persistent-test/src/MigrationTest.hs
```diff
@@ -17,6 +17,11 @@ Target
 Source
     field3 Int
     field4 TargetId
+
+CustomSqlId
+    pk      Int   sql=id
+    Primary pk
```
parsonsmatt: This is the actual test case for the issue in question.

persistent-test/src/MigrationTest.hs
```diff
@@ -38,6 +43,7 @@ specsWith runDb = describe "Migration" $ do
       again <- getMigration migrationMigrate
       liftIO $ again @?= []
     it "really is idempotent" $ runDb $ do
+      runMigrationSilent migrationMigrate
```
parsonsmatt: reall,y really, REALLY is idempotent

parsonsmatt:
@tysonzero want to give this a shot?

tysonzero:
It seems to have fixed the issue.

Although now it's trying to add a lot of foreign keys from tables to themselves:

```
ALTER TABLE "<table>" ADD CONSTRAINT "<table>_id_fkey" FOREIGN KEY("id") REFERENCES "<table>"("id") ;
...
```

parsonsmatt:
I think I've fixed the self-reference hack, thanks for noticing that :D

PR #1072, opened 2020-04-12
Refactor PostgresJSON Tests

JonathanLorimer:
I wasn't sure how to put all of the db entry keys into scope. I made some ad hoc data structures to do so, but if anyone has any suggestions on how to do this better I would be open to them!

> Note: I used `runIO` instead of `before` or `beforeAll` because I wanted to have access to the db entry keys. Additionally, I didn't insert the `edgeCases` when constructing the spec tree because they mess up some of the earlier tests. I had to insert the `edgeCases` in each `it` block afterwards, so I would still have access to the keys and they wouldn't pollute the db state.

- ~~Bumped the version number~~
- ~~Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)~~
- ~~Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock~~

After submitting your PR:

- ~~Update the Changelog.md file with a link to your PR~~

### DIFF
--- a/persistent-postgresql/persistent-postgresql.cabal
+++ b/persistent-postgresql/persistent-postgresql.cabal
@@ -1,5 +1,5 @@
 name:            persistent-postgresql
-version:         2.11.0.0
+version:         2.11.0.1
 license:         MIT
 license-file:    LICENSE
 author:          Felipe Lessa, Michael Snoyman <michael@snoyman.com>
--- a/persistent-postgresql/test/JSONTest.hs
+++ b/persistent-postgresql/test/JSONTest.hs
@@ -85,18 +85,19 @@ data TestDBKeys record =
              , bools   :: [Key record]
              , numbers :: [Key record]
              , strings :: [Key record]
-             , arrays  :: [Key record]
+             , lists  :: [Key record]
              , objects :: [Key record] }
 
 specs :: Spec
 specs =
-    describe "tests for @>. operator"
+    describe "Testing JSON operators"
     $ beforeAll migrateJSON
     $ before preHook
     $ afterAll_ cleanUp $ do
 
       -- Setup DB and get ahold of keys
       TestDBKeys {..} <- runIO $ runConn_ $ do
+          liftIO $ putStrLn "-------------------- JSON Test Setup -------------------------"
           nullK <- insert' Null
           let nulls = [nullK]
 
@@ -129,7 +130,7 @@ specs =
           arrListK <- insert' $ toJSON ([emptyArr,emptyArr,toJSON [emptyArr,emptyArr]])
           arrList2K <- insert' $ toJSON [emptyArr,toJSON [Number 3,Bool False],toJSON [emptyArr,toJSON [Object mempty]]]
           arrFilledK <- insert' $ toJSON [Null, Number 4, String "b", Object mempty, emptyArr, object [ "test" .= [Null], "test2" .= String "yes"]]
-          let arrays = [ arrNullK, arrListK, arrList2K, arrFilledK ]
+          let lists = [ arrNullK, arrListK, arrList2K, arrFilledK ]
 
           objNullK <- insert' $ Object mempty
           objTestK <- insert' $ object ["test" .= Null, "test1" .= String "no"]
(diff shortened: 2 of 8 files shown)


persistent-postgresql/persistent-postgresql.cabal
```diff
@@ -1,5 +1,5 @@
 name:            persistent-postgresql
-version:         2.11.0.0
+version:         2.11.0.1
```
parsonsmatt, line 2: We don't need a version bump here because we haven't released 2.11.0.0 yet.
```suggestion
version:         2.11.0.0
```

persistent-postgresql/test/JSONTest.hs
```diff
@@ -40,431 +42,654 @@ insert' :: (MonadIO m, PersistStoreWrite backend, BaseBackend backend ~ SqlBacke
...
+  arrList3K <- insert' $ toJSON [toJSON [String "a"], Number 1]
+  arrList4K <- insert' $ toJSON [String "a", String "b", String "c", String "d"]
+  objEmptyK <- insert' $ object ["" .= Number 9001]
+  objFullK  <- insert' $ object ["a" .= Number 1, "b" .= Number 2
+                                , "c" .= Number 3, "d" .= Number 4 ]
+  return $ EdgeCaseKeys {..}
+
+globalKeys :: (Monad m, MonadIO m) => ReaderT SqlBackend m (GlobalKeys TestValue)
+globalKeys = do
+          nullK <- insert' Null
```
parsonsmatt, line 90: We should stick to a consistent indentation. The project uses 4 space or 2 space depending on the module (sigh) but nothing uses the 8 space.

persistent-postgresql/test/JSONTest.hs
```diff
@@ -40,431 +42,654 @@ insert' :: (MonadIO m, PersistStoreWrite backend, BaseBackend backend ~ SqlBacke
...
+               , objEmptyK  :: Key record
+               , objFullK   :: Key record }
+
+data GlobalKeys record =
+  GlobalKeys { nulls   :: [Key record]
+             , bools   :: [Key record]
+             , numbers :: [Key record]
+             , strings :: [Key record]
+             , lists   :: [Key record]
+             , objects :: [Key record] }
```
parsonsmatt, line 147: Formatting wise, It hink this should be:

```haskell
data GlobalKeys record = GlobalKeys
  { nulls :: [Key record]
  , bools :: [Key record]
  , ...
  }
```

I'm curious why it's polymorphic.

persistent-postgresql/test/JSONTest.hs
```diff
@@ -40,431 +42,654 @@ insert' :: (MonadIO m, PersistStoreWrite backend, BaseBackend backend ~ SqlBacke
...
+
 specs :: Spec
-specs = describe "postgresql's JSON operators behave" $ do
+specs =
+    describe "Testing JSON operators"
+    $ beforeAll setup
+    $ afterAll_ teardown $ do
+
+      -- Setup DB and get ahold of keys
+      GlobalKeys {..} <- runIO $ runConn_ $ globalKeys
```
parsonsmatt, line 156: I don't think we want to execute this in the `runIO` part of the process. This runs it while constructing the spec tree, not during spec execution.

persistent-postgresql/test/JSONTest.hs
```diff
@@ -40,431 +42,654 @@ insert' :: (MonadIO m, PersistStoreWrite backend, BaseBackend backend ~ SqlBacke
...
+      let [ nullK ] = nulls
+      let [ boolTK, boolFK ] = bools
+      let [ num0K, num1K, numBigK, numFloatK, numSmallK, numFloat2K, numBigFloatK ] = numbers
+      let [ strNullK, strObjK, strArrK, strAK, strTestK, str2K, strFloatK ] = strings
+      let [ arrNullK, arrListK, arrList2K, arrFilledK ] = lists
+      let [ objNullK, objTestK, objDeepK ] = objects
```
parsonsmatt, line 162: I'm *really* not a fan of getting the patterns like this, it's extremely fragile. It's more repetitive but it makes sense IMO to capture each key as a record field instead of constructing lists and then deconstructing them.

persistent-postgresql/test/JSONTest.hs
```diff
@@ -40,431 +42,654 @@ insert' :: (MonadIO m, PersistStoreWrite backend, BaseBackend backend ~ SqlBacke
...
+
+      describe "@>. object queries" $ do
+        it "matches an empty Object with any object" $ db $ do
+            vals <- selectList [TestValueJson @>. Object mempty] []
+            objects `matchKeys`  vals
```
parsonsmatt, line 167: Ahhh yes these look great!

persistent-postgresql/test/JSONTest.hs
```diff
@@ -40,431 +42,654 @@ insert' :: (MonadIO m, PersistStoreWrite backend, BaseBackend backend ~ SqlBacke
...
+             , numbers :: [Key record]
+             , strings :: [Key record]
+             , lists   :: [Key record]
+             , objects :: [Key record] }
+
 specs :: Spec
-specs = describe "postgresql's JSON operators behave" $ do
+specs =
+    describe "Testing JSON operators"
+    $ beforeAll setup
```
parsonsmatt, line 152: So the 'magic' of `beforeAll` is that you can run a function `IO SomeDataType` and then the `SomeDataType` is provided to every `it`, like:

```haskell
beforeAll (pure "hello world") $ do
  it "gives us hello world" $ \string -> do
    string `shouldBe` "hello world"
```

So we can write `beforeAll (setup :: IO GlobalKeys)` and then each spec item can look like:

```
it "lmao" $ \GlobalKeys{..} -> ...
```

When you get to the edge cases, you can write `beforeAll` *again*, but this time, because it's under a `beforeAll (setup :: IO Thing)`, the type of the function `beforeAll (setupExtra :: Thing -> IO ExtraThing)`, and now the `it` specs under that would be 

```haskell
it "message" $ \ExtraThing {..} -> ...
```

This may be a cleaner way to structure this.

JonathanLorimer: Oh, I didn't know that about `beforeAll`, I like that way better!

persistent-postgresql/test/JSONTest.hs
JonathanLorimer: Yeah, I agree with you. At first a field for each seemed like a lot of work, but I think you are right!

persistent-postgresql/test/JSONTest.hs
JonathanLorimer: I am getting a type error when I try and nest a `beforeAll` beneath another `beforeAll`. I think it is because the `beforeAll` returns a `Spec`, but the do block that the nested `beforeAll` resides in expects a `SpecWith GlobalKeys`. I am sure I am missing something simple, but if you have any advice that would be appreciated!

JonathanLorimer: I am just going to make all the tests work with the same initial state. I think that will simplify everything.

JonathanLorimer:
@parsonsmatt I made the changes you request. I ended up just making all the tests work with the same setup, as I was having trouble adding a second `beforeAll` beneath the first one (types weren't checking). I think having the same setup is actually a cleaner approach anyways.

parsonsmatt:
Fantastic, thanks! This is great work :smile:

parsonsmatt:
awesome thanks!

JonathanLorimer:
Thanks for your guidance on this Matt, I really appreciate it.

PR #1076, opened 2020-04-19
Add cabal flag and CPP option to use Loc from monad-logger.

dmjio:
If template haskell is disabled in monad-logger, this needs to
be propagated to persistent. Otherwise, type errors ensue. 

```
Database/Persist/Sql/Raw.hs:43:18: error:
    • Couldn't match type ‘Control.Monad.Logger.Loc’
                     with ‘Language.Haskell.TH.Syntax.Loc’
      NB: ‘Language.Haskell.TH.Syntax.Loc’
            is defined in ‘Language.Haskell.TH.Syntax’
                in package ‘template-haskell-2.14.0.0’
          ‘Control.Monad.Logger.Loc’
            is defined in ‘Control.Monad.Logger’
                in package ‘monad-logger-0.3.30’
      Expected type: Control.Monad.Logger.Loc
                     -> Control.Monad.Logger.LogSource
                     -> Control.Monad.Logger.LogLevel
                     -> fast-logger-2.4.15:System.Log.FastLogger.LogStr.LogStr
                     -> IO ()
        Actual type: LogFunc
    • In the second argument of ‘runLoggingT’, namely
        ‘(connLogFunc conn)’
      In a stmt of a 'do' block:
        runLoggingT
          (logDebugNS (pack "SQL") $ T.append sql $ pack $ "; " ++ show vals)
          (connLogFunc conn)
      In the expression:
        do runLoggingT
             (logDebugNS (pack "SQL") $ T.append sql $ pack $ "; " ++ show vals)
             (connLogFunc conn)
           getStmtConn conn sql
   |
43 |                 (connLogFunc conn)
   |                  ^^^^^^^^^^^^^^^^

Database/Persist/Sql/Raw.hs:65:10: error:
    • Couldn't match type ‘Control.Monad.Logger.Loc’
                     with ‘Language.Haskell.TH.Syntax.Loc’
      NB: ‘Language.Haskell.TH.Syntax.Loc’
            is defined in ‘Language.Haskell.TH.Syntax’
                in package ‘template-haskell-2.14.0.0’
          ‘Control.Monad.Logger.Loc’
            is defined in ‘Control.Monad.Logger’
                in package ‘monad-logger-0.3.30’
      Expected type: Control.Monad.Logger.Loc
                     -> Control.Monad.Logger.LogSource
                     -> Control.Monad.Logger.LogLevel
                     -> fast-logger-2.4.15:System.Log.FastLogger.LogStr.LogStr
                     -> IO ()
        Actual type: LogFunc
    • In the second argument of ‘runLoggingT’, namely
        ‘(connLogFunc conn)’
      In a stmt of a 'do' block:
        runLoggingT
          (logDebugNS (pack "SQL") $ T.append sql $ pack $ "; " ++ show vals)
          (connLogFunc conn)
      In the expression:
        do conn <- projectBackend `liftM` ask
           runLoggingT
             (logDebugNS (pack "SQL") $ T.append sql $ pack $ "; " ++ show vals)
             (connLogFunc conn)
           stmt <- getStmt sql
           res <- liftIO $ stmtExecute stmt vals
           ....
   |
65 |         (connLogFunc conn)
   |          ^^^^^^^^^^^^^^^^
```

*Note*: this change will not affect any code by default, only when the flag is disabled.

After submitting your PR:

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -20,6 +20,8 @@
 * Fix a bug where unsafe migration error messages were being shown using `Show` prior to printing, resulting in less helpful output. [#1080](https://github.com/yesodweb/persistent/pull/1080)
 * [#1087](https://github.com/yesodweb/persistent/pull/1087)
   * `RawSql` now has tuple instances up to GHC's max tuple size (62)
+* [#1076](https://github.com/yesodweb/persistent/pull/1076)
+  * `Loc` is now imported from `monad-logger` as opposed to `template-haskell`. Removes `template-haskell` as an explicit dependency.
 
 ## 2.10.5.2
 
--- a/persistent/Database/Persist/Sql/Types/Internal.hs
+++ b/persistent/Database/Persist/Sql/Types/Internal.hs
@@ -23,15 +23,9 @@ module Database.Persist.Sql.Types.Internal
     , IsSqlBackend
     ) where
 
-#ifdef MONAD_LOGGER_TH
-import Language.Haskell.TH.Syntax (Loc)
-import Control.Monad.Logger (LogSource, LogLevel)
-#else
-import Control.Monad.Logger (LogSource, LogLevel, Loc)
-#endif
-
 import Data.List.NonEmpty (NonEmpty(..))
 import Control.Monad.IO.Class (MonadIO (..))
+import Control.Monad.Logger (LogSource, LogLevel, Loc)
 import Control.Monad.Trans.Class (lift)
 import Control.Monad.Trans.Reader (ReaderT, runReaderT, ask)
 import Data.Acquire (Acquire)
--- a/persistent/Database/Persist/Sql/Types/Internal.hs
+++ b/persistent/Database/Persist/Sql/Types/Internal.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE CPP #-}
 module Database.Persist.Sql.Types.Internal
     ( HasPersistBackend (..)
     , IsPersistBackend (..)
--- a/persistent/Database/Persist/Sql/Types/Internal.hs
+++ b/persistent/Database/Persist/Sql/Types/Internal.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE RankNTypes #-}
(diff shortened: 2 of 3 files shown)


MaxGabriel:
I'm not sure if I'm reading this right, but it kind of looks like `Loc` is always exported from monad-logger (the fake one if you have template haskell turned off, the TH one if you have it turned one). Could we instead always import `Loc` from monad-logger, and then we'd just get the correct version depending on what flag you have on monad-logger?

MaxGabriel:
If I'm correct about that, it should probably be documented in monad-logger that that's the intended approach

dmjio:
I'm trying to understand exactly why I did this. It does seem that `monad-logger` exports `Loc` from either `template-haskell` or its own defined `Loc` [here](https://github.com/snoyberg/monad-logger/blob/master/Control/Monad/Logger.hs#L103-L109). So it seems as if the changes I proposed would be redundant. If that were the case I would not expect the error message above to occur. 

I honestly think `Loc` has the correct type, but for some reason GHC is not allowing the type being exported to be used since it comes from a different package. It would seem that nobody has tried to disable `template_haskell` in `monad-logger` and then use `persistent` with it, but I could be wrong here.

Here is a minimal reproduction of what I'm experiencing using the latest `persistent-2.10.5.2` and `monad-logger-0.3.34`

```nix
{ pkgs ? import <nixpkgs> {} }: # nix rev 96069f7d890b90cbf4e8b4b53e15b036210ac146
with pkgs.haskell.lib;
let
  monad-logger = disableCabalFlag pkgs.haskellPackages.monad-logger "template_haskell";
  persistent = pkgs.haskellPackages.persistent.override { inherit monad-logger; };
in
  persistent
```

MaxGabriel:
Oh, what I think the issue is is that Persistent does not import `Loc` from `monad-logger`, it always imports it from template haskell:

https://github.com/yesodweb/persistent/blob/be8901eac2714e09a083c0d46bb9a2073650c497/persistent/Database/Persist/Sql/Types/Internal.hs#L39

Would you be able to try Persistent always importing like this:

```
import Control.Monad.Logger (LogSource, LogLevel, Loc)
```

and seeing if that compiles for you?

dmjio:
Nice, worked for both `template_haskell` enabled and disabled on `monad-logger`. I reverted the flag stuff, and we can squash the other commits to clean up the history.

MaxGabriel:
Nice, do you mind adding a note to the changelog file about this?

dmjio:
@MaxGabriel looks like you beat me to it. Sorry about that.

MaxGabriel:
Thanks, merged! I also added documentation about this to monad-logger

dmjio:
Awesome, thanks a ton 🍻

PR #1077, opened 2020-04-19
Issues/Questions about Postgres array support

MaxGabriel:
Hey all, we started trying to use the Postgres array support at work and ran into some issues. I'm not sure which of these are considered bugs, vs expected limitations.

1. Documentation on using `PersistArray` is pretty limited and we could probably improve that. Happy to PR this. Some notable things:

    * The docs currently say "Intended especially for PostgreSQL backend for text arrays". I don't think this is true? They seem to work fine for `int[]` for example.

    * Generally lacking a description of how to write `PersistField` or `PersistFieldSql` instances for them.

2. Serializing to `PersistArray` deserializes to `PersistList`. This isn't a big deal, and I haven't looked into if there's a solution for it, but it would be great to document it if not! See `instance PersistField RoundtripTextArray` in the draft PR for an example of this failing.

3. Trying to store data in a `jsonb[]` column fails with the following error:

```
       SqlError {sqlState = "42804", sqlExecStatus = FatalError, sqlErrorMsg = "column \"test\" is of type jsonb[] but expression is of type text[]", sqlErrorDetail = "", sqlErrorHint = "You will need to rewrite or cast the expression."}
```

4. Can/should persistent-postgresql provide any datatypes that allow people to use PersistArray without writing their own instances?

### DIFF
--- a/persistent-postgresql/persistent-postgresql.cabal
+++ b/persistent-postgresql/persistent-postgresql.cabal
@@ -52,6 +52,8 @@ test-suite test
                      JSONTest
                      CustomConstraintTest
                      PgIntervalTest
+                     ArrayTest
+                     ArrayTest.Instances
     ghc-options:     -Wall
 
     build-depends:   base                 >= 4.9 && < 5
--- a/persistent-postgresql/test/ArrayTest.hs
+++ b/persistent-postgresql/test/ArrayTest.hs
@@ -1,0 +1,92 @@
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-} -- FIXME
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module ArrayTest where
+
+import Control.Monad.IO.Class (MonadIO)
+import Data.Aeson
+import Data.List (sort)
+import qualified Data.Text as T
+import Test.Hspec.Expectations ()
+
+import PersistentTestModels
+import PgInit
+import ArrayTest.Instances
+
+share [mkPersist persistSettings,  mkMigrate "migrate"] [persistLowerCase|
+  TestRoundtrip
+    test RoundtripTextArray
+    deriving Show Eq
+  TestListHack
+    test ListHackTextArray
+    deriving Show Eq
+  TestIntArray
+    test IntArray
+    deriving Show Eq
+  TestJSONArray
+    test (JSONArray Text)
+    deriving Show Eq
+|]
+
+-- cleanDB :: (BaseBackend backend ~ SqlBackend, PersistQueryWrite backend, MonadIO m) => ReaderT backend m ()
+-- cleanDB = deleteWhere ([] :: [Filter TestValue])
+
+-- setup :: IO TestKeys
+-- setup = asIO $ runConn_ $ do
+--   void $ runMigrationSilent migrate
+
+-- teardown = cleanDB
+
+shouldBeIO :: (Show a, Eq a, MonadIO m) => a -> a -> m ()
+shouldBeIO x y = liftIO $ shouldBe x y
+
+roundTrip :: (MonadIO m, PersistStoreWrite backend,
+                    PersistEntity a, Show a, Eq a,
+                    PersistEntityBackend a ~ BaseBackend backend) =>
+                   a -> ReaderT backend m ()
+roundTrip x = do
+  xId <- insert x
+  maybeX <- get xId
+  case maybeX of
+    Nothing -> error "expected to get Just"
+    Just x2 -> x2 `shouldBeIO` x
+
+specs :: Spec
+specs = do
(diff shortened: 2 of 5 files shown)


MaxGabriel:
cc @Vlix who initially added the Postgres array support, and @parsonsmatt who reviewed the PR

also my co-workers @mitchellvitez and @danielbarter who were looking into this

MaxGabriel:
Ok, one more thing, it looks like Persistent wants to remigrate the columns each time? e.g. if you run this:

```
(1) ~/D/C/H/y/persistent> stack test persistent-postgresql --test-arguments="--match PersistArray"        14:41:53
persistent-postgresql> test (suite: test, args: --match PersistArray)

ALTER TABLE "test_roundtrip" ALTER COLUMN "test" TYPE text[];
ALTER TABLE "test_list_hack" ALTER COLUMN "test" TYPE text[];
ALTER TABLE "test_int_array" ALTER COLUMN "test" TYPE int[];
ALTER TABLE "test_j_s_o_n_array" ALTER COLUMN "test" TYPE jsonb[];
```

You will see those `ALTER TABLE`s on each run, whereas I believe they should only be on the first run. You can see in the database the columns already have that type as well:

```
                                                Table "public.test_roundtrip"
 Column |  Type  | Collation | Nullable |                  Default                   | Storage  | Stats target | Description
--------+--------+-----------+----------+--------------------------------------------+----------+--------------+-------------
 id     | bigint |           | not null | nextval('test_roundtrip_id_seq'::regclass) | plain    |              |
 test   | text[] |           | not null |                                            | extended |              |
```

This isn't explicitly tested for in these tests atm, but that would be a good one to add

parsonsmatt:
Thanks for reporting and exploring on this!

Vlix:
Hey @MaxGabriel, I added in the PersistArray constructor to be able to correctly `toField` a text array (`"{text1,text2,etc}"`) for the JSON module. Specifically the `(?|.)` and `(?&.)` operators.
I didn't intend it to be used in any other way, so anyone wants to correctly implement it can take as many liberties as they want, as long as the JSON tests pass. :)

MaxGabriel:
Ahhh, ok that is helpful info @Vlix. I saw @parsonsmatt closed the existing Postgres array issues (https://github.com/yesodweb/persistent/issues?q=is%3Aissue+postgres+array)—Matt may have thought it was considered solved?

Vlix:
@MaxGabriel I have not seen those issues. And yes, it seems like @parsonsmatt assumed I implemented arrays completely. I probably should have put big disclaimers around that constructor :/
I couldn't use `PersistList` (which semantically would make most sense) because it uses to/from `JSON` marshalling in `text` columns and it would probably break way too much, so that was just not usable.
But yes, I emphatically state that `PersistArray` was not meant to be used in anything else than the JSON operators.

I also see my JSON tests could have been written better. Apologies @ @parsonsmatt

parsonsmatt:
No worries @Vlix - your contributions are super welcome and I'd rather have something incomplete than wait for a perfection that never comes :smile:

parsonsmatt:
Fortunately we have a breaking change 2.11 coming out soon and then I want to get 3.0.0 out before too long, so we have two opportunities to implement breaking changes to behavior to provide extra functionality.

MaxGabriel:
Yes I want to second @parsonsmatt's thanks to you @Vlix. Postgres array support has been a requested feature for a long, long time so it was great you got it working for the JSON part of it. I just didn't realize that was what was intended

MaxGabriel:
Ok looked into this a little, small update on https://github.com/yesodweb/persistent/pull/1077#issuecomment-616743081

Currently we run this code to get data about columns:

```
    let sqlv=T.concat ["SELECT "
                          ,"column_name "
                          ,",is_nullable "
                          ,",COALESCE(domain_name, udt_name)" -- See DOMAINS below
                          ,",column_default "
                          ,",numeric_precision "
                          ,",numeric_scale "
                          ,",character_maximum_length "
                          ,"FROM information_schema.columns "
                          ,"WHERE table_catalog=current_database() "
                          ,"AND table_schema=current_schema() "
                          ,"AND table_name=? "
                          ,"AND column_name <> ?"]
```

The important part of the query is that it returns `_int4` for the `udt_name`. This causes Persistent to think the column needs to be migrated from an `_int4` (based on the query) to `int[]` (Based on the PersistFieldSql instance), in the `modType` function inside `findAlters`.

It looks like that table also has a `data_type` parameter, which is `ARRAY` for the arrays. When the value is array, we can join with the `element_types` table to get additional info about the type, per this: https://www.postgresql.org/docs/11/infoschema-element-types.html

MaxGabriel:
I did a little investigation into postgresql-simple and came up with a way to support arrays of enums with it https://github.com/haskellari/postgresql-simple/issues/57

MaxGabriel:
Hm, ok but even if inserts are solved, selects are still a problem. My understanding is: when custom types are used, they get assigned an OID by Postgres identifying their type. Persistent gets access to the OID like so:

```
oids <- forM [0..cols-1] $ \col -> fmap ((,) col) (LibPQ.ftype ret col)
```

Then it checks the OID against a handful of hard-coded options (`builtinGetters`), and if it doesn't find a suitable match it defaults to `defaultGetter`, which uses `PersistLiteralEscaped`. This works fine for string-based custom types like enums or e.g. an email type, but not others (e.g. an array).

To get this additional data, Persistent can use [getTypeInfo](https://hackage.haskell.org/package/postgresql-simple-0.6.3/docs/Database-PostgreSQL-Simple-TypeInfo.html#v:getTypeInfo) (or query Postgres itself). The results of that query are cached on a per-connection basis, which is helpful.

(`postgresql-simple` notes that sharing this cache among a pool of connections would be more efficient. postgresql-simple could maybe also make `getTypeInfo` work on a list of OIDs, since the current setup requires e.g. 5 queries to Postgres if a result row contains 5 new types).

Overall I think this could be a good move to use this, even though the caching behavior could be better.

PR #1080, opened 2020-04-23
Fix display of exception error messages, #1079

### LINKED ISSUE #1079: Migration error messages are not printed correctly
#969 unfortunately made it so that errors from migrations are not printed in a very helpful way. Repro script:
```haskell
#!/usr/bin/env stack
-- stack --resolver lts-15.0 script
{-# LANGUAGE EmptyDataDecls             #-}
{-# LANGUAGE FlexibleContexts           #-}
{-# LANGUAGE GADTs                      #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE MultiParamTypeClasses      #-}
{-# LANGUAGE OverloadedStrings          #-}
{-# LANGUAGE QuasiQuotes                #-}
{-# LANGUAGE TemplateHaskell            #-}
{-# LANGUAGE TypeFamilies               #-}
import Control.Monad.Logger (runStdoutLoggingT)
import Control.Monad.IO.Class  (liftIO)
import Database.Persist
import Database.Persist.Postgresql
import Database.Persist.TH
import Control.Monad.Reader

share [mkMigrate "migrate1"] [persistLowerCase|
Person1 sql=person
    name String
    age Int Maybe
    deriving Show Eq
|]

share [mkMigrate "migrate2"] [persistLowerCase|
Person2 sql=person
    name String
    deriving Show Eq
|]

main :: IO ()
main = runStdoutLoggingT $ withPostgresqlPool "dbname=persistent" 1 $ liftSqlPersistMPool $ do
  runMigration migrate1
  runMigration migrate2
```

Desired output:
```
Migrating: ALTER TABLE "person" ALTER COLUMN "age" TYPE INT8
[Debug#SQL] ALTER TABLE "person" ALTER COLUMN "age" TYPE INT8; []

Database migration: manual intervention required.
The unsafe actions are prefixed by '***' below:

*** ALTER TABLE "person" DROP COLUMN "age";
```

Actual output:
```
Migrating: ALTER TABLE "person" ALTER COLUMN "age" TYPE INT8
[Debug#SQL] ALTER TABLE "person" ALTER COLUMN "age" TYPE INT8; []
migration-errors.hs: PersistError "\n\nDatabase migration: manual intervention required.\nThe unsafe actions are prefixed by '***' below:\n\n*** ALTER TABLE \"person\" DROP COLUMN \"age\";\n"
```
I'm happy to include OS versions etc if they would help but I'm pretty sure this is not affected by those.

hdgarrood:
Fixes #1079. Running the test script from that issue now prints

    Migrating: ALTER TABLE "person" ALTER COLUMN "age" TYPE INT8
    [Debug#SQL] ALTER TABLE "person" ALTER COLUMN "age" TYPE INT8; []
    persistent-repro-exe:

    Database migration: manual intervention required.
    The unsafe actions are prefixed by '***' below:

    *** ALTER TABLE "person" DROP COLUMN "age";

This is arguably a breaking change, since if people were catching this exception in their code before, they'll need to change their uses of `catch` to catch this type. If #969 was previously released without a breaking-level version bump and nobody complained, though, this is probably okay to release without a breaking-level version bump too?

After submitting your PR:

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,9 +1,5 @@
 # Changelog for persistent
 
-## 2.11.1.0
-
-* Fix a bug where unsafe migration error messages were being shown using `Show` prior to printing, resulting in less helpful output. [#1080](https://github.com/yesodweb/persistent/pull/1080)
-
 ## 2.11.0.0
 
 * Introduces a breaking change to the internal function `mkColumns`, which can now be passed a record of functions to override its default behavior. [#996](https://github.com/yesodweb/persistent/pull/996)
@@ -21,6 +17,7 @@
     record` from a `record` if it was defined with `Primary`.
 * [#1036](https://github.com/yesodweb/persistent/pull/1036):
   * The method `entityIdFromJSON` that is used to parse entities now correctly works for entities that define a custom `Primary` key.
+* Fix a bug where unsafe migration error messages were being shown using `Show` prior to printing, resulting in less helpful output. [#1080](https://github.com/yesodweb/persistent/pull/1080)
 
 ## 2.10.5.2
 
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for persistent
 
+## 2.11.1.0
+
+* Fix a bug where unsafe migration error messages were being shown using `Show` prior to printing, resulting in less helpful output. [#1080](https://github.com/yesodweb/persistent/pull/1080)
+
 ## 2.11.0.0
 
 * Introduces a breaking change to the internal function `mkColumns`, which can now be passed a record of functions to override its default behavior. [#996](https://github.com/yesodweb/persistent/pull/996)
--- a/persistent/Database/Persist/Sql/Types.hs
+++ b/persistent/Database/Persist/Sql/Types.hs
@@ -7,13 +7,13 @@ module Database.Persist.Sql.Types
     , OverflowNatural(..)
     ) where
 
-import Control.Exception (Exception)
+import Control.Exception (Exception(..))
 import Control.Monad.Logger (NoLoggingT)
 import Control.Monad.Trans.Reader (ReaderT (..))
 import Control.Monad.Trans.Resource (ResourceT)
 import Control.Monad.Trans.Writer (WriterT)
 import Data.Pool (Pool)
-import Data.Text (Text)
+import Data.Text (Text, unpack)
 import Data.Typeable (Typeable)
 
 import Database.Persist.Types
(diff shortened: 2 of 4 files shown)


persistent/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for persistent
 
+## 2.11.1.0
```
parsonsmatt, line 3: I've *got* to update the changelog to make it less confusing - can you put this in `2.11.0.0`?

persistent/Database/Persist/Sql/Types.hs
```diff
@@ -114,3 +114,30 @@ type ConnectionPool = Pool SqlBackend
...
+  show (PersistUnsafeMigrationException mig) =
+    concat
+      [ "\n\nDatabase migration: manual intervention required.\n"
+      , "The unsafe actions are prefixed by '***' below:\n\n"
+      , unlines $ map displayMigration mig
+      ]
+    where
+      displayMigration :: (Bool, Sql) -> String
+      displayMigration (True,  s) = "*** " ++ unpack s ++ ";"
+      displayMigration (False, s) = "    " ++ unpack s ++ ";"
```
parsonsmatt, line 141: I'd rather have a derived `Show` instance and a custom implementation of `displayException`.

persistent/persistent.cabal
```diff
@@ -1,5 +1,5 @@
 name:            persistent
-version:         2.11.0.0
+version:         2.11.1.0
```
parsonsmatt, line 2:
```suggestion
version:         2.11.0.0
```

persistent/Database/Persist/Sql/Types.hs
hdgarrood: I tried that to start with, but unfortunately GHC uses Show, not displayException, when exiting due to an uncaught exception 🤷‍♂️ I’m not sure how else to handle it. Maybe the exception type could just be `data UnsafeMigrationException = UnsafeMigrationException` and we could instead print the human readable error message before throwing it?

hdgarrood: That’s not super helpful if people do want to catch the error, though.

persistent/ChangeLog.md
hdgarrood: Sure thing! I’ve seen people use a heading “unreleased changes” at the top of changelogs, which seems to work okay.

persistent/Database/Persist/Sql/Types.hs
parsonsmatt: are you serious, GHC? okay, this is fine then

hdgarrood: haha yeah that was more or less my reaction too

PR #1085, opened 2020-05-25
Allow aeson 1.5

felixonmars:
Builds fine and all tests pass here.

### DIFF
--- a/persistent-template/ChangeLog.md
+++ b/persistent-template/ChangeLog.md
@@ -1,5 +1,7 @@
 ## Unreleased changes
 
+* Allow aeson 1.5. [#1085](https://github.com/yesodweb/persistent/pull/1085)
+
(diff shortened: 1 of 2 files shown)


parsonsmatt:
Why'd you delete all the "things to do before/after submitting a PR"? I'd like to have a CHANGELOG entry and version bump for this - patch should do fine. Thanks!

felixonmars:
I thought it is not worth a version bump but only a revision, so I removed all of them. I'll update now.

jkachmar:
Ping re. commercialhaskell/stackage#5393, anything blocking this from being merged that I could help out on?

parsonsmatt:
Thanks for the ping - this LGTM and I'm happy to incorporate + make a HAckage revision.

Issue #1095, opened 2020-07-01
DB representation for `Word64` values greater than `maxBound :: Int64` are negative


erikd:
Defining a column as `Word64` and writing a value in the range `[maxBoundInt64 +1 .. maxBoundWord64]` results in a negative value (ie the `Int64` bit pattern of the `Word64`) being inserted into the database.

The reason for this is that the `Word64` type in the schema definition is converted to `SqlInt64` which is then written to the database. The reverse is done when read the value is read, so the value is correct in Haskell land but the value in the database is negative. This is a problem for interop when the database is used from languages other then Haskell.

The obvious solution is to add an `SqlWord64` constructor to the type to `SqlType`.

@parsonsmatt Is this a sane idea? Volunteering to do this if it is.

erikd:
I should add that we do not need `SqlWord32` because the Persistent SQL representation for that is `SqlInt64`.

parsonsmatt:
Yeah this sounds perfectly reasonable, thanks!

erikd:
I have something that compiles. I want to test it in my application before raising a PR.

PR #1096, opened 2020-07-05
Add proper support for Word64

erikd:
Previously when the schema used `Word64` as the column type, Persistent
would would use `SqlInt64` as the SQL representation which means that
`Word64` values above `maxBound :: Int64` would be stored as negative
values in the database. That is fine for a database only accessed from
Haskell but is a pain in the neck when the database is used as an
interop layer for other languages.

This commit fixes these issues by adding `SqlWord64` and `PersistWord64`.

Closes: https://github.com/yesodweb/persistent/issues/1095

This works in my application. Specifically it fixes https://github.com/input-output-hk/cardano-db-sync/issues/163 but has not been tested beyond that. I have not fixed or even tested this for the other database back ends. I will see what CI says.

After submitting your PR:


erikd:
@parsonsmatt Once I fix all the CI errors is there anything else required to get this merged?

parsonsmatt:
I'm going to be out for the next week so you may want to ping another maintainer if you want it in soon :) I generally trust your judgement and I don't expect there will be any issues aside from the checklist.

erikd:
Currently getting the following failure in CI:
```
 src/LargeNumberTest.hs:25:3: 

  1) Large Numbers preserves their values in the database

       uncaught exception: PersistException

       PersistMarshalError "Couldn't parse field `word64` from table `number`.
       Failed to parse Haskell type `Word64`; expected integer from database, but received:
       PersistDouble 1.8446744073709552e19. Potential solution: Check that your database schema matches your
       Persistent model definitions."
```
I am not even sure which database backend this is coming from.

Clues?

erikd:
How does one run the tests? I use Postgres and was trying `cabal test persistent-postgresql` and I get:
```
Test suite test: RUNNING...
)
Test suite test: FAIL
```

persistent/Database/Persist/Class/PersistField.hs
```diff
@@ -226,8 +228,13 @@ instance PersistField Word32 where
     fromPersistValue x = Left $ fromPersistValueError "Word32" "integer" x
 
 instance PersistField Word64 where
-    toPersistValue = PersistInt64 . fromIntegral
+    toPersistValue = PersistWord64 . fromIntegral
```
parsonsmatt, line 231: is `fromIntegral` here a no-op?

persistent/Database/Persist/Types/Base.hs
```diff
@@ -534,6 +538,7 @@ data SqlType = SqlString
              | SqlTime
              | SqlDayTime -- ^ Always uses UTC timezone
              | SqlBlob
+             | SqlWord64 -- @since 2.11.0
```
parsonsmatt, line 541:
```suggestion
             | SqlWord64 -- ^ @since 2.11.0
```

so it is a proper doc comment

persistent/Database/Persist/Types/Base.hs
```diff
@@ -368,6 +368,7 @@ instance Error PersistException where
 data PersistValue = PersistText Text
                   | PersistByteString ByteString
                   | PersistInt64 Int64
+                  | PersistWord64 Word64  -- @since 2.11.0
```
parsonsmatt, line 371:
```suggestion
                  | PersistWord64 Word64  -- ^ @since 2.11.0
```

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -382,6 +382,7 @@ showSqlType :: SqlType -> Text
 showSqlType SqlString = "VARCHAR"
 showSqlType SqlInt32 = "INTEGER"
 showSqlType SqlInt64 = "INTEGER"
+showSqlType SqlWord64 = "INTEGER"
```
parsonsmatt, line 385: I wonder if this is what's going on with the failing test you're finding. Based on [this](https://sqlite.org/datatype3.html) saying that `INTEGER` is (at most) an 8 byte size, I'd expect that `maxBound :: Word64` would be troublesome.

parsonsmatt (approved):
I think the testing setup relied on older behavior of postgres authentication. It should probably require a specific user and a password.

persistent-sqlite/Database/Persist/Sqlite.hs
parsonsmatt: Which, uh. Hm. Would suggest that the current behavior is Correct, since the test isn't failing right now, at least for SQLite.

parsonsmatt:
@erikd So in my investigation on this problem, it looks like there isn't going to be a way to get SQLite to have the "proper" behavior - the `INTEGER` type in SQLite just doesn't support values that large, so the negation trick is the only way to make it work out. I think we can just back that behavior out to the older behavior, throw a doc comment somewhere as to why `sqlite` doesn't use the Word64 constructor, and get this out the door :smile:

Issue #775, opened 2018-01-26
createSqlPool's idle timeout seems much too short


ocharles:
https://hackage.haskell.org/package/persistent-2.7.3.1/docs/src/Database-Persist-Sql-Run.html#createSqlPool calls `createPool` with 1 stripe and a 20s idle timeout on resources. I would say that

a) This is much too small

and, to contradict myself somewhat

b) There is no one-size-fits-all.

I would at least bump this up to 5 minutes. Anything but a heavily loaded site essentially negates any benefit of using a pool, constantly forming new connections. I've instrumented the use of a 20s timeout in a pool at work, and while we are constantly serving requests, in the last hour we've opened 160 new database connections. 

Here is our traffic (average requests/s with a 1h look back)
![traffic](https://user-images.githubusercontent.com/20878/35457267-34cce962-02d0-11e8-8119-b24771543517.png)

Here is a a counter of new database connections (average new connections per second).
![connections](https://user-images.githubusercontent.com/20878/35457268-34e91650-02d0-11e8-91c2-4e8042495584.png)

And here is a graph showing every time the new connection counter increases:

![counter](https://user-images.githubusercontent.com/20878/35457387-a7b9891c-02d0-11e8-965e-262393329bf2.png)

ocharles:
The defaults in https://pgbouncer.github.io/config.html might provide some good defaults for `persistent`. They are somewhat PostgreSQL specific.

> server_idle_timeout
> If a server connection has been idle more than this many seconds it will be dropped. If 0 then timeout is disabled. [seconds]
> Default: 600.0

MaxGabriel:
My instinct is that the idle timeout should be increased and that it should be made configurable. Personally I can only think of really weird edge case scenarios where the increased timeout could cause a problem; can anyone think of anything serious?

Are you also suggesting that the stripe count should be increased and/or be configurable? I haven't used anything but 1 before, but if we're making the idle timeout configurable then may as well make the stripe count configurable as well.

Hope things are going well at CircuitHub btw!

ocharles:
> My instinct is that the idle timeout should be increased and that it should be made configurable.

Agreed and agreed.

> Personally I can only think of really weird edge case scenarios where the increased timeout could cause a problem; can anyone think of anything serious?

On both sides your resource usage is going to be higher. However, you should be anticipating using every connection in the pool anyway, so if you're unable to operate in those constraints then you already had a problem!

[...]

Issue #920, opened 2019-06-22
Make the definition types abstract


parsonsmatt:
Currently the `SqlBackend` and `*Def` types are all completely exposed. Adding extra functionality to the `persistent` library requires a breaking change version bump.

We should make these types abstract from the general interface so that additions to the data types are not a breaking change, and we can add functionality in minor version bumps.

We can put them in a `persistent-core` package per ["internal convention is a mistake"](https://www.reddit.com/r/haskell/comments/a0ioo3/internal_convention_is_a_mistake/) or we can do the `Internal` module. I don't have a strong preference here.

This is a pretty big change, so I anticipate it'll go in as 3.0. I'll find some other issues/enhancements to slot in there.

PR #1098, opened 2020-07-13
Configurable idle timeout

MaxGabriel:
This PR ~is a draft~ shows what it might look like to have a configurable idle timeout and stripe size for Persistent's SQL backends. This would resolve the issues brought up in #775.

~This PR is just a draft~. It's untested and needs a few small additions, but most things are in place. (Edit: added a test)

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -107,7 +107,7 @@ instance Show PostgresServerVersionError where
       "Unexpected PostgreSQL server version, got " <> uniqueMsg
 instance Exception PostgresServerVersionError
 
--- | Create a PostgreSQL connection pool and run the given action.  The pool is
+-- | Create a PostgreSQL connection pool and run the given action. The pool is
 -- properly released after the action finishes using it.  Note that you should
 -- not use the given 'ConnectionPool' outside the action since it may already
 -- have been released.
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -276,14 +276,16 @@ getServerVersionNonEmpty :: PG.Connection -> IO (NonEmpty Word)
 getServerVersionNonEmpty conn = do
   [PG.Only version] <- PG.query_ conn "show server_version";
   case AT.parseOnly parseVersion (T.pack version) of
-    Left err -> throwIO $ PostgresServerVersionError $ "Failed to parse Postgres version. Got: " <> version <> ". Error: " <> err
+    Left err -> throwIO $ PostgresServerVersionError $ "Parse failure on: " <> version <> ". Error: " <> err
     Right versionComponents -> case NEL.nonEmpty versionComponents of
-      Nothing -> throwIO $ PostgresServerVersionError $ "Empty Postgres version string. Got: " <> version
+      Nothing -> throwIO $ PostgresServerVersionError $ "Empty Postgres version string: " <> version
       Just neVersion -> pure neVersion
 
   where
     -- Partially copied from the `versions` package
-    parseVersion = AT.decimal `AT.sepBy` AT.char '.' <* AT.endOfInput
+    -- Typically server_version gives e.g. 12.3
+    -- In Persistent's CI, we get "12.4 (Debian 12.4-1.pgdg100+1)", so we ignore the trailing data.
+    parseVersion = AT.decimal `AT.sepBy` AT.char '.'
 
 -- | Choose upsert sql generation function based on postgresql version.
 -- PostgreSQL version >= 9.5 supports native upsert feature,
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -11,9 +11,11 @@ module Database.Persist.Postgresql
     , withPostgresqlPoolWithVersion
     , withPostgresqlConn
     , withPostgresqlConnWithVersion
+    , withPostgresqlPoolWithConf
     , createPostgresqlPool
     , createPostgresqlPoolModified
     , createPostgresqlPoolModifiedWithVersion
+    , createPostgresqlPoolWithConf
     , module Database.Persist.Sql
     , ConnectionString
     , PostgresConf (..)
@@ -24,6 +26,8 @@ module Database.Persist.Postgresql
     , fieldName
     , mockMigration
     , migrateEnableExtension
+    , PostgresConfHooks(..)
+    , defaultPostgresConfHooks
     ) where
 
 import qualified Debug.Trace as Debug
@@ -51,6 +55,7 @@ import qualified Blaze.ByteString.Builder.Char8 as BBB
 import Data.Acquire (Acquire, mkAcquire, with)
 import Data.Aeson
 import Data.Aeson.Types (modifyFailure)
+import qualified Data.Attoparsec.Text as AT
 import qualified Data.Attoparsec.ByteString.Char8 as P
 import Data.Bits ((.&.))
 import Data.ByteString (ByteString)
@@ -136,7 +141,25 @@ withPostgresqlPoolWithVersion :: (MonadUnliftIO m, MonadLogger m)
                               -- ^ Action to be executed that uses the
                               -- connection pool.
                               -> m a
-withPostgresqlPoolWithVersion getVer ci = withSqlPool $ open' (const $ return ()) getVer ci
+withPostgresqlPoolWithVersion getVerDouble ci = do
+  let getVer = oldGetVersionToNew getVerDouble
+  withSqlPool $ open' (const $ return ()) getVer ci
+
+-- | Same as 'withPostgresqlPool', but can be configured with 'PostgresConf' and 'PostgresConfHooks'.
+--
+-- @since 2.11.0.0
+withPostgresqlPoolWithConf :: (MonadUnliftIO m, MonadLogger m)
+                           => PostgresConf -- ^ Configuration for connecting to Postgres
+                           -> PostgresConfHooks -- ^ Record of callback functions
+                           -> (Pool SqlBackend -> m a)
+                           -- ^ Action to be executed that uses the
+                           -- connection pool.
+                           -> m a
+withPostgresqlPoolWithConf conf hooks = do
+  let getVer = pgConfHooksGetServerVersion hooks
+      modConn = pgConfHooksAfterCreate hooks
+  let logFuncToBackend = open' modConn getVer (pgConnStr conf)
+  withSqlPoolWithConfig logFuncToBackend (postgresConfToConnectionPoolConfig conf)
 
 -- | Create a PostgreSQL connection pool.  Note that it's your
 -- responsibility to properly close the connection pool when
@@ -179,8 +202,30 @@ createPostgresqlPoolModifiedWithVersion
     -> ConnectionString -- ^ Connection string to the database.
     -> Int -- ^ Number of connections to be kept open in the pool.
     -> m (Pool SqlBackend)
-createPostgresqlPoolModifiedWithVersion getVer modConn ci =
+createPostgresqlPoolModifiedWithVersion getVerDouble modConn ci = do
+  let getVer = oldGetVersionToNew getVerDouble
   createSqlPool $ open' modConn getVer ci
+
+-- | Same as 'createPostgresqlPool', but can be configured with 'PostgresConf' and 'PostgresConfHooks'.
+--
+-- @since 2.11.0.0
+createPostgresqlPoolWithConf
+    :: (MonadUnliftIO m, MonadLogger m)
+    => PostgresConf -- ^ Configuration for connecting to Postgres
+    -> PostgresConfHooks -- ^ Record of callback functions
+    -> m (Pool SqlBackend)
+createPostgresqlPoolWithConf conf hooks = do
+  let getVer = pgConfHooksGetServerVersion hooks
+      modConn = pgConfHooksAfterCreate hooks
+  createSqlPoolWithConfig (open' modConn getVer (pgConnStr conf)) (postgresConfToConnectionPoolConfig conf)
+
+postgresConfToConnectionPoolConfig :: PostgresConf -> ConnectionPoolConfig
+postgresConfToConnectionPoolConfig conf =
+  ConnectionPoolConfig
+    { connectionPoolConfigStripes = pgPoolStripes conf
+    , connectionPoolConfigIdleTimeout = fromInteger $ pgPoolIdleTimeout conf
+    , connectionPoolConfigSize = pgPoolSize conf
+    }
 
 -- | Same as 'withPostgresqlPool', but instead of opening a pool
 -- of connections, only one connection is opened.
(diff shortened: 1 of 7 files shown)


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1261,19 +1275,27 @@ escape (DBName s) =
 data PostgresConf = PostgresConf
     { pgConnStr  :: ConnectionString
       -- ^ The connection string.
+    
+    , pgPoolStripes :: Int
+    -- ^ How many stripes to divide the pool into. See "Data.Pool" for details.
+    , pgPoolIdleTimeout :: Integer
```
MaxGabriel, line 1281: Ideally this would be `NominalDiffTime`. However, NominalDiffTime lacks a Read instance, and `PostgresConf` derives `Read`. I requested a Read instance here: https://github.com/haskell/time/issues/130

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -180,6 +180,20 @@ createPostgresqlPoolModifiedWithVersion
 createPostgresqlPoolModifiedWithVersion getVer modConn ci =
   createSqlPool $ open' modConn getVer ci
 
+createPostgresqlPoolWithConf
+    :: (MonadUnliftIO m, MonadLogger m)
+    => (PG.Connection -> IO (Maybe Double)) -- ^ Action to perform to get the server version.
+    -> (PG.Connection -> IO ()) -- ^ Action to perform after connection is created.
```
MaxGabriel, line 186: These two functions are kind of annoying, naming wise. Something like `createPostgresqlPoolModifiedWithVersionWithConf` would be gross (`createPostgresqlPoolModifiedWithVersion` is pretty bad on its own tbh). I'm also somewhat worried that another callback will eventually be added here (Perhaps a callback for when a connection is destroyed, or some other Postgres specific thing?)

Possibly a separate data type is called for to hold these functions (`PostgresConfCallbacks` isn't quite right, but something like that), with a function to create a default data structure?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1261,19 +1275,27 @@ escape (DBName s) =
 data PostgresConf = PostgresConf
     { pgConnStr  :: ConnectionString
       -- ^ The connection string.
+    
+    , pgPoolStripes :: Int
```
MaxGabriel, line 1279: Adding new fields could be considered a breaking change for persistent-postgresql

persistent/Database/Persist/Sql/Types.hs
```diff
@@ -56,6 +59,16 @@ type Migration = WriterT [Text] (WriterT CautiousMigration (ReaderT SqlBackend I
 
 type ConnectionPool = Pool SqlBackend
 
+data ConnectionPoolConfig = ConnectionPoolConfig
+    { connectionPoolConfigStripes :: Int
+    , connectionPoolConfigIdleTimeout :: NominalDiffTime
+    , connectionPoolConfigSize :: Int
+    }
+    deriving (Show)
```
MaxGabriel, line 67: I think having a separate data type for the connection pool configuration is nicer than 3 separate function parameters. Thoughts?

MaxGabriel: In an ideal world for me, there is a field of `PostgresConf` named `poolConfig :: ConnectionPoolConfig`. Unfortunately, `PostgresConf` already has an existing field for the pool size, so I broke out stripes + idle timeout to separate fields there.

persistent/Database/Persist/Sql/Types.hs
```diff
@@ -56,6 +59,16 @@ type Migration = WriterT [Text] (WriterT CautiousMigration (ReaderT SqlBackend I
 
 type ConnectionPool = Pool SqlBackend
 
+data ConnectionPoolConfig = ConnectionPoolConfig
+    { connectionPoolConfigStripes :: Int
+    , connectionPoolConfigIdleTimeout :: NominalDiffTime
+    , connectionPoolConfigSize :: Int
```
MaxGabriel, line 65: I've tried to keep the order of stripes + idle timeout + pool size consistent. I think it's somewhat unfortunate that it's easy to mix up stripes and pool size though.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1282,11 +1304,11 @@ instance FromJSON PostgresConf where
                    , PG.connectDatabase = database
                    }
             cstr = PG.postgreSQLConnectionString ci
-        return $ PostgresConf cstr pool
+        return $ PostgresConf cstr poolStripes poolIdleTimeout poolSize
 instance PersistConfig PostgresConf where
     type PersistConfigBackend PostgresConf = SqlPersistT
     type PersistConfigPool PostgresConf = ConnectionPool
-    createPoolConfig (PostgresConf cs size) = runNoLoggingT $ createPostgresqlPool cs size -- FIXME
+    createPoolConfig conf = runNoLoggingT $ createPostgresqlPoolWithConf getServerVersion (const $ pure ()) conf -- FIXME
```
MaxGabriel, line 1311: I'm not sure what this FIXME is about.

persistent/Database/Persist/Sql/Types.hs
MaxGabriel: But if adding the two new fields is a breaking change, possibly we could go all-out and replace the fields with a ConnectionPoolConfig field?

parsonsmatt: there are breaking changes, and then there is Breaking Changes, and, weeeeeell, I'd rather the easiest breaking change as possible. Making these types more abstract is a goal of #920 so we can leave a `TODO` here to clean it up with `persistent-3.0` possibly.

MaxGabriel: Ok cool, that's helpful guidance 👍

MaxGabriel:
Ok, I think this is pretty ready for review. It's just missing changelog + cabal bumps at this point. But I've left a few TODOs where I was uncertain about things

MaxGabriel:
Ok, added to the Changelog and `@since 2.11.0.0`

persistent/Database/Persist/Sql/Types.hs
```diff
@@ -56,6 +59,23 @@ type Migration = WriterT [Text] (WriterT CautiousMigration (ReaderT SqlBackend I
 
 type ConnectionPool = Pool SqlBackend
 
+-- | Values to configure a pool of database connections. See "Data.Pool" for details.
+--
+-- @since 2.11.0.0
+data ConnectionPoolConfig = ConnectionPoolConfig
+    { connectionPoolConfigStripes :: Int -- ^ How many stripes to divide the pool into. See "Data.Pool" for details. Default: 1.
+    , connectionPoolConfigIdleTimeout :: NominalDiffTime -- ^ How long connections can remain idle before being disposed of, in seconds. Default: 600l
```
MaxGabriel, line 67:
```suggestion
    , connectionPoolConfigIdleTimeout :: NominalDiffTime -- ^ How long connections can remain idle before being disposed of, in seconds. Default: 600
```

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: `Hooks` is another term that gets used for this concept a bit - 

```haskell
data PostgresConfHooks = PostgresConfHooks
  { getServerVersion :: IO (Maybe Double)
  , afterCreate :: IO ()
  }

createPostgresqlPoolWithHooks :: (PG.Connection -> PostgresConfHooks) -> ...
```

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: If they're exported, yeah, it's technically a breaking change.

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: Yeah looks fine to me. Let's delete it?

parsonsmatt:
Everything looks great to me! I'm happy to merge now or we can incorporate some of the changes I mentioned. Either way is good for me.

MaxGabriel:
I'll take a look at your suggestions 👍

persistent-postgresql/Database/Persist/Postgresql.hs
MaxGabriel:
```suggestion
    createPoolConfig conf = runNoLoggingT $ createPostgresqlPoolWithConf getServerVersion (const $ pure ()) conf
```

persistent-postgresql/Database/Persist/Postgresql.hs
MaxGabriel: So, I was investigating this today and I was thinking this could be an opportunity to clean up the server version API. I have these complaints:

1. Double is a weird way to represent a version. It doesn't work for 3 digit versions (x.y.z) (though Postgres moved to just 2 digit versions recently which helps). I think something like `NonEmpty Integer` makes more sense, e.g. `[10, 2]`

2. The `Maybe` behavior feels weird to me. If you return `Nothing`, Persistent assumes the lowest possible version, not trying to do things like upsert. This isn't documented afaict, and would subtly degrade behavior if someone happened to return `Nothing`.

As long as we're adding a new API, I would propose something like:

```
getServerVersion :: IO (NonEmpty Integer)
```

In the case of Redshift users who can't get the server version (assuming that's still a thing), they can return a hard-coded value that matches whatever Postgres capabilities Redshift has, e.g. `pure (10 :| [2])`.

If there was something better suited than `NonEmpty Integer` I'd be fine with that too.

MaxGabriel: Implementation would be something like:

```
getServerVersion2 :: PG.Connection -> IO (NonEmpty Word)
getServerVersion2 conn = do
  [PG.Only version] <- PG.query_ conn "show server_version";
  case AT.parseOnly parseVersion (T.pack version) of
    Left err -> throwIO $ PostgresServerVersionError $ "Failed to parse Postgres version. Got: " <> version <> ". Error: " <> err
    Right versionComponents -> case NEL.nonEmpty versionComponents of
      Nothing -> throwIO $ PostgresServerVersionError $ "Empty Postgres version string. Got: " <> version
      Just neVersion -> pure neVersion

  where
    -- Partially copied from the `versions` package
    parseVersion = AT.decimal `AT.sepBy` AT.char '.' <* AT.endOfInput
```

MaxGabriel:
Ok, I pushed a change which moves to using `NonEmpty Word` as the new way of representing Postgres versions, while being backwards compatible with `Maybe Double`. See https://github.com/yesodweb/persistent/pull/1098#discussion_r507198177.

Is this an approach we want to go with? If so, I can just do final touchups on this PR and we should be good to go

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: Sounds great to me

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -244,23 +322,23 @@ openSimpleConn = openSimpleConnWithVersion getServerVersion
 --
 -- @since 2.9.1
 openSimpleConnWithVersion :: (PG.Connection -> IO (Maybe Double)) -> LogFunc -> PG.Connection -> IO SqlBackend
-openSimpleConnWithVersion getVer logFunc conn = do
+openSimpleConnWithVersion getVerDouble logFunc conn = do
     smap <- newIORef $ Map.empty
-    serverVersion <- getVer conn
+    serverVersion <- (oldGetVersionToNew getVerDouble) conn
```
parsonsmatt, line 327:
```suggestion
    serverVersion <- oldGetVersionToNew getVerDouble conn
```

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -244,23 +322,23 @@ openSimpleConn = openSimpleConnWithVersion getServerVersion
...
     return $ createBackend logFunc serverVersion smap conn
 
 -- | Create the backend given a logging function, server version, mutable statement cell,
 -- and connection.
-createBackend :: LogFunc -> Maybe Double
+createBackend :: LogFunc -> (NonEmpty Word)
```
parsonsmatt, line 332:
```suggestion
createBackend :: LogFunc -> NonEmpty Word
```

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1356,6 +1448,29 @@ instance PersistConfig PostgresConf where
         addPass     = maybeAddParam "password" "PGPASS"
         addDatabase = maybeAddParam "dbname"   "PGDATABASE"
 
+-- | Hooks for configuring the Persistent/its connection to Postgres
+--
+-- TODO: API guarantees? Consider not exporting constructor and/or fields
```
parsonsmatt, line 1453:
```suggestion
-- TODO: API guarantees? Consider not exporting constructor and/or fields
--
-- @since 2.11.0
```

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1356,6 +1448,29 @@ instance PersistConfig PostgresConf where
...
+      --
+      -- The default implementation queries the server with "show server_version".
+      -- Some variants of Postgres, such as Redshift, don't support showing the version.
+      -- It's recommended you return a hardcoded version in those cases.
+  , pgConfHooksAfterCreate :: PG.Connection -> IO ()
+      -- ^ Action to perform after a connection is created.
+      --
+      -- Typical uses of this are modifying the connection (e.g. to set the schema) or logging a connection being created.
+  }
+
```
parsonsmatt, line 1466:
```suggestion

-- | 
--
-- @since 2.11.0
```

parsonsmatt (approved):
Looks great :smile: A few suggestions but I'm excited to get this in.

MaxGabriel:
Talked with Matt a little on Slack: he's good with no specific attempt make to make data types backwards compatible with future updates. In Persistent 3.0, we'll move to some model like that.

persistent-postgresql/test/PgInit.hs
```diff
@@ -89,24 +90,50 @@ runConn :: MonadUnliftIO m => SqlPersistT (LoggingT m) t -> m ()
...
+runConnInternal :: MonadUnliftIO m => RunConnType -> SqlPersistT (LoggingT m) t -> m t
+runConnInternal connType f = do
   travis <- liftIO isTravis
   let debugPrint = not travis && _debugOn
-  let printDebug = if debugPrint then print . fromLogStr else void . return
+      printDebug = if debugPrint then print . fromLogStr else void . return
+      poolSize = 1
+  connString <- if travis
+    then do
+      pure "host=localhost port=5432 user=postgres dbname=persistent"
```
parsonsmatt, line 109:
```suggestion
      pure "host=localhost port=5432 user=perstest password=perstest dbname=persistent"
```

MaxGabriel:
Looks like the postgres server version on CI is formatted differently; working on a patch to make the parser more flexible

Issue #579, opened 2016-06-14
No builtin support for UUID


pseudonom:
Should support for UUIDs be built in to `persistent` or should it be its own package? FYI, the own package version might look something like

```haskell

{-# OPTIONS_GHC -fno-warn-orphans #-}

module Instances where

import Control.Error.Util
import Data.UUID (UUID)
import qualified Data.UUID as UUID
import qualified Data.ByteString.Lazy as BSL
import Database.Persist.Sql
import Web.HttpApiData

instance PersistField UUID where
  toPersistValue = PersistDbSpecific . BSL.toStrict . UUID.toByteString
  fromPersistValue (PersistDbSpecific t) =
    case UUID.fromByteString $ BSL.fromStrict t of
      Just x  -> Right x
      Nothing -> Left "Invalid UUID"
  fromPersistValue _ = Left "Not PersistDBSpecific"
instance PersistFieldSql UUID where
  sqlType _ = SqlOther "uuid"
instance PathPiece UUID where
  toPathPiece = toUrlPiece
  fromPathPiece = hush . parseUrlPiece
instance ToHttpApiData UUID where
  toUrlPiece = UUID.toText
instance FromHttpApiData UUID where
  parseUrlPiece uuid = note ("Error parsing UUID " <> uuid) . UUID.fromText $ uuid
instance ToJSON UUID where
  toJSON = toJSON . UUID.toText
instance FromJSON UUID where
  parseJSON uuid = maybe (fail $ "Error parsing UUID " <> show uuid) pure . UUID.fromText <=< parseJSON $ uuid
```

gregwebs:
I think it should be added.
* it's lightweight
* uuid is an important (even if overused) data type for databases

Want to send

bergey:
I'd like to work on this over the next couple of weeks.  Is the approach shown above good by the maintainers, or would you hold out for a solution that supports more backends?  I'm only familiar with Postgres, so I will need some help if folks want the more general solution.

Incidentally, the HTTP & JSON instances have been added upstream, so only the 2 Persist instances are needed now.

MaxGabriel:
I'm good with this approach. This is what we're using at work right now if you want to base it off that:

```haskell
--------------------------------------------------------------------------------
instance PersistField UUID where
  toPersistValue = PersistDbSpecific . toASCIIBytes
  fromPersistValue (PersistDbSpecific uuid) =
    case fromASCIIBytes uuid of
[...]

PR #1102, opened 2020-07-29
Add UUID instances for PersistField

jappeace:
Issue: https://github.com/yesodweb/persistent/issues/579

As discussed this only works for postgres. This suggested PR has no support for MySQL.  
An alternative could be creating a newtype as suggested by @parsonsmatt 
I'm not sure what's best. But at least now we have something to point at.

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Class/PersistField.hs
+++ b/persistent/Database/Persist/Class/PersistField.hs
@@ -32,6 +32,7 @@ import Numeric.Natural (Natural)
 import Text.Blaze.Html
 import Text.Blaze.Html.Renderer.Text (renderHtml)
 import GHC.TypeLits
+import Data.UUID.Types
 
 import Database.Persist.Types.Base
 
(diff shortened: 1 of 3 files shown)


MaxGabriel:
So, I think this should probably be a newtype and be in the persistent-postgresql package, so other databases could have their own instances.

Possibly UUID could have a PersistField instance in regular `persistent`, and a newtype of _that_ could be in the postgres package for a  `PersistFieldSql` instance. This would work if most databases use the same format for UUIDs (I'm seeing MySQL recommend a binary version of UUIDs, though, so this might not work https://mysqlserverteam.com/mysql-8-0-uuid-support/)

jappeace:
@MaxGabriel 

I'm considering extending the `SqlType` value instead with a SqlUUID constructor. First class support for UUID seems like something many users want, doing it like this essentially gives us the ability to make the best default choice. If a user wants to use a different choice they can still choose to use a newtype downstream.
Using a newtype in this repository works, but it feels more like a workaround. What do you think?

parsonsmatt:
`SqlUUID` is the right point to add it, IMO, since it's not consistent among databases how it is represented.

jessekempf:
@MaxGabriel: Do you really anticipate it would be that difficult for us to infer the encoding, and then decode that? The possible wire encodings boil down to strings, strings without `-` separators, and 16 bytes in either host or network byte order.

Issue #1104, opened 2020-08-06
Regression: Show instance for keys no longer includes the name of the key


MaxGabriel:
Previously, `show`ing a `Key` would include the name of the key;

```
git checkout persistent-template-2.7.4

stack ghci

import CustomPrimaryKeyReferenceTest

> TweetKey 1
TweetKey {unTweetKey = 1}

```

But later, `show`ing a `Key` only shows the underlying value of the newtype:

```
git checkout persistent-template-2.8.0.1

stack ghci

> import CustomPrimaryKeyReferenceTest
> TweetKey 1
1
```

I believe this occurred because persistent-template-2.8.0 (there isn't a tag for it, so I checked out 2.8.0.1 above) started deriving the Show instance with the `Newtype` strategy, rather than `Stock`. I think this should probably be changed back to `Stock` for both `Show` and `Read`, for a few reasons:

1. This is an unintentional change from previous behavior
2. By removing the constructor from the show instance, you can no longer tell what you're looking up when you `show` the key. For example, if an error is thrown from `getJust`:

```
getJust :: ( PersistStoreRead backend
           , PersistRecordBackend record backend
           , MonadIO m
           ) => Key record -> ReaderT backend m record
getJust key = get key >>= maybe
  (liftIO $ throwIO $ PersistForeignConstraintUnmet $ T.pack $ show key)
  return
```

you won't see that you failed to lookup, say a `TweetKey 1`, you'll just see you failed to lookup `1`. Pretty unhelpful to users!

3. This is a minor reason, but if someone is really relying on the Read/Show instances, it's probably helpful to have the extra safety that `Show`ing a FooKey can't be `read` as a `BarKey`

OS version: MacOS 
stack version: Version 2.3.1, Git revision de2a7b694f07de7e6cf17f8c92338c16286b2878 (8103 commits) x86_64 hpack-0.33.0
Persistent versions: see above

PR #1106, opened 2020-08-09
persistent-template: Use the stock strategy when deriving Show/Read f…

MaxGabriel:
…or keys

This closes https://github.com/yesodweb/persistent/issues/1104

If people are amenable to this PR, thoughts on the version for it? This could be a breaking change, as described in the Changelog (idk how people feel about that though—if changing the content of a Show/Read instance is considered to be a breaking change)

- [ ] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html) (n/a)
- [ ] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock (n/a)

After submitting your PR:

### DIFF
--- a/persistent-template/ChangeLog.md
+++ b/persistent-template/ChangeLog.md
@@ -1,6 +1,6 @@
 ## Unreleased changes
 
-* Always use the "stock" strategy when deriving Show/Read for keys
+* Always use the "stock" strategy when deriving Show/Read for keys [#1106](https://github.com/yesodweb/persistent/pull/1106)
 	* This fixes a regression from 2.8.0, which started using the `newtype` strategy when deriving `Show`/`Read` for keys
 	* In practice, this means that from 2.8.0–2.8.3.1, for the following schema:
 
--- a/persistent-template/ChangeLog.md
+++ b/persistent-template/ChangeLog.md
@@ -1,5 +1,7 @@
 ## Unreleased changes
 
+## 2.9
+
 * Always use the "stock" strategy when deriving Show/Read for keys [#1106](https://github.com/yesodweb/persistent/pull/1106)
 	* This fixes a regression from 2.8.0, which started using the `newtype` strategy when deriving `Show`/`Read` for keys
 	* In practice, this means that from 2.8.0–2.8.3.1, for the following schema:
--- a/persistent-template/ChangeLog.md
+++ b/persistent-template/ChangeLog.md
@@ -1,5 +1,28 @@
 ## Unreleased changes
 
+* Always use the "stock" strategy when deriving Show/Read for keys
+	* This fixes a regression from 2.8.0, which started using the `newtype` strategy when deriving `Show`/`Read` for keys
+	* In practice, this means that from 2.8.0–2.8.3.1, for the following schema:
+
+	```
+	Person
+		name Text
+	CustomPrimary
+		anInt Int
+		Primary anInt
+		name Text
+	```
+
+	`PersonKey 1` would show as `"SqlBackendKey {unSqlBackendKey = 1}"`
+	and `CustomPrimaryKey 1` would show as `"1"`
+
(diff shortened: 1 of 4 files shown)


MaxGabriel:
Cool, do you have thoughts on what the version number should be?

parsonsmatt:
well, it's a breaking change, but since it fixes a bug that is *also* a breaking change that came in a non-breaking manner, I think we're fine to release a patch version and deprecate the versions that contained the breaking behavior.

MaxGabriel:
Sounds good

lf-:
Hi, as someone whose software has been broken by this, and not because we were relying on the output of the show instance, but actually because it broke a relation, requiring manual database hacking (I will double check later today but a LTS update from 14.27->16.11 *did* break our data), this should be a minor release number change, so people who assumed it was an intentional change and migrated their data accordingly already can see it more easily since a new minor release will likely break compilation on the next LTS update. 

Here's what happened to https://github.com/Carnap/Carnap, specifically Carnap-Server:

signal-2020-09-04-091936

gleachkr:
To add a little detail to the above: we were using `derivePersistField` with a datatype that included a key as a field in one of its constructors. So, I think what happened is exactly what you'd expect, given the description of the issue above. This PR fixes the bad behavior.

MaxGabriel:
Ah ok, I think this is a strong case to do a breaking version bump for this

parsonsmatt:
@MaxGabriel I'm confused by what you mean - it seems to me that @lf- and @gleachkr so far is saying a minor version bump is appropriate, which is my intuition as well.

MaxGabriel:
Ok I'm kinda confused, this is how I was reading people:

@parsonsmatt 
> well, it's a breaking change, but since it fixes a bug that is also a breaking change that came in a non-breaking manner, I think we're fine to release a patch version and deprecate the versions that contained the breaking behavior.

I am taking that to mean, in the PVP versioning schema (A.B.C.D/MAJOR.MAJOR.MINOR.PATCH), you were saying to just bump PATCH.

@lf- 
>this should be a minor release number change, so people who assumed it was an intentional change and migrated their data accordingly already can see it more easily since a new minor release will likely break compilation on the next LTS update

This part confuses me. LTS Haskell automatically picks up minor version bumps (https://github.com/commercialhaskell/lts-haskell#how-it-works), but it sounds like what @lf- wanted was for this new version to not be picked up until the next LTS version (where people would assume breaking changes). But for that to happen, it needs a major (either A or  B in A.B.C.D)

lf-:
@MaxGabriel: What I meant is that I wanted it to cause version bounds to break, encouraging reading the changelog, but I was not aware of the implications for LTS Haskell users as I primarily work with Nix. I believe the fix should require a major upgrade, even though it would delay LTS availability since, for example, if an app already on the buggy version newly introduces one of the affected use cases, thereby becoming affected by this bug, it would only have data needing migration following the fix release. Regardless of our choice, migrations/fixes will need to happen for all apps using show that have used the affected version, so our role is to most effectively notify people that these migrations are needed.

MaxGabriel:
Ok cool, so I think I’m in favor of a major (B in A.B.C.D to be specific) bump. That said, I am more open to making lots of major version bumps than most people.

Now that it’s clear there is a disagreement, can the people who prefer C or D version bumps weigh in? I am inclined to agree with @lf-

parsonsmatt:
I don't have a strong preference. We can put this in with 2.11, or we can get it out-the-door quicker with a 2.10 minor release.

MaxGabriel:
Ok cool, let's just do 2.11. I'll update the PR accordingly

MaxGabriel:
Ok, updated to 2.9 (persistent template's versions aren't synced with persistent proper's)

parsonsmatt (approved):

MaxGabriel:
Ok cool this is released. The big caveat is: the previous version of persistent-template depends on 2.11, which isn't release, so this isn't available yet in practice

PR #1108, opened 2020-08-11
Test Stackage lts-15 and lts-16 on CI

sol:


MaxGabriel:
This seems good to me. Possibly we should think about dropping support for 8.2 as well

sol:
Is there anything specific that I can do to help move this forward?

MaxGabriel:
ah, yes, sorry. Do you mind bumping the version of Persistent-mongoDB in the .cabal file, and adding a line to the changelog?

parsonsmatt:
I'm fine publishing a revision and accepting as-is

PR #1111, opened 2020-08-18
persistent-qq: Add interpolation support for multirow VALUES syntax

sol:
Before submitting your PR, check that you've:

- [x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

- [x] Update the Changelog.md file with a link to your PR
- [x] Bumped the version number if there isn't an `(unreleased)` on the Changelog
- [x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

PR #1112, opened 2020-08-18
persistent-qq: Fix Haddock markup + fix warnings + refactor

sol:
This are all the commits from #1111 that don't change any behavior; as a separate PR.  I assume these require less discussion + they are useful on their own.  So let's get these merged first!

### DIFF
--- a/persistent-qq/src/Database/Persist/Sql/Raw/QQ.hs
+++ b/persistent-qq/src/Database/Persist/Sql/Raw/QQ.hs
@@ -162,15 +162,15 @@ makeQQ x = QuasiQuoter
 -- We can now execute this raw query:
 --
 -- @
--- let lft = 10 :: Int
---     rgt = 20 :: Int
---     width = rgt - lft
---     nams = "first" :| ["second", "third"]
+-- let lft = 10 :: `Int`
+--     rgt = 20 :: `Int`
+--     width = rgt `-` lft
+--     nams = "first" `:|` ["second", "third"]
 --  in [sqlQQ|
---       DELETE FROM ^{Category} WHERE @{CategoryLft} BETWEEN #{lft} AND #{rgt};
---       UPDATE category SET @{CategoryRgt} = @{CategoryRgt} - #{width} WHERE @{CategoryRgt} > #{rgt};
---       UPDATE category SET @{CategoryLft} = @{CategoryLft} - #{width} WHERE @{CategoryLft} > #{rgt};
---       SELECT ?? FROM ^{Category} WHERE ^{Category}.@{CategoryNam} IN %{nams};
+--       DELETE FROM ^{Category} WHERE \@{CategoryLft} BETWEEN \#{lft} AND \#{rgt};
+--       UPDATE category SET \@{CategoryRgt} = \@{CategoryRgt} - \#{width} WHERE \@{CategoryRgt} > \#{rgt};
+--       UPDATE category SET \@{CategoryLft} = \@{CategoryLft} - \#{width} WHERE \@{CategoryLft} > \#{rgt};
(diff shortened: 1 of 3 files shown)


sol:
@parsonsmatt @snoyberg ping.

persistent-qq/src/Database/Persist/Sql/Raw/QQ.hs
```diff
@@ -69,6 +70,21 @@ parseStr a ('^':'{':xs) = Literal (reverse a) : parseHaskell TableName  [] xs
 parseStr a ('@':'{':xs) = Literal (reverse a) : parseHaskell ColumnName [] xs
 parseStr a (x:xs)       = parseStr (x:a) xs
 
+interpolateValues :: PersistField a => NonEmpty a -> (Text, [[PersistValue]]) -> (Text, [[PersistValue]])
+interpolateValues xs =
+    first (mkPlaceholders values <>) .
+    second (NonEmpty.toList values :)
+    where
```
parsonsmatt, line 77: Minor style comment 
```suggestion
  where
```

persistent-qq/src/Database/Persist/Sql/Raw/QQ.hs
```diff
@@ -69,6 +70,21 @@ parseStr a ('^':'{':xs) = Literal (reverse a) : parseHaskell TableName  [] xs
...
+    values = NonEmpty.map toPersistValue xs
+
+mkPlaceholders :: NonEmpty a -> Text
+mkPlaceholders values = "(" <> n `timesCommaSeparated` "?" <> ")"
+    where
```
parsonsmatt, line 82:
```suggestion
  where
```

persistent-qq/test/Spec.hs
```diff
@@ -19,33 +19,27 @@ import PersistTestPetType
...
-sqlite_database = mkSqliteConnectionInfo sqlite_database_file
-
 runConn :: MonadUnliftIO m => SqlPersistT (LoggingT m) t -> m ()
 runConn f = do
-  let debugPrint = _debugOn
-  let printDebug = if debugPrint then print . fromLogStr else void . return
+  let printDebug = when _debugOn . B.putStrLn . fromLogStr
   flip runLoggingT (\_ _ _ s -> printDebug s) $ do
-    _ <- withSqlitePoolInfo sqlite_database 1 $ runSqlPool f
+    _ <- withSqliteConn ":memory:" $ runSqlConn f
```
parsonsmatt, line 31: What motivated this change?

sol: @parsonsmatt creating a file in the source tree interferes with file watching (specifically, `hspec/sensei`).  If you don't like the in-memory approach, I can create the sqlite-file somewhere under `/tmp` instead.

persistent-qq/src/Database/Persist/Sql/Raw/QQ.hs
sol: @parsonsmatt I followed the odd indentation for `where` to match the style used in the rest of the file.  I pushed 
e12934c to fix it everywhere within that source file.  Feel free to cherry-pick.

persistent-qq/test/Spec.hs
parsonsmatt: It's totally fine, I'm just curious for the change. In memory seems like it'd be preferable anyway, and I'm not sure why it was setup like that.

persistent-qq/src/Database/Persist/Sql/Raw/QQ.hs
parsonsmatt: Ah! Thank you. I appreciate it <3

sol:
@parsonsmatt an other friendly ping.

PR #1121, opened 2020-08-31
Foreign key improvements

### LINKED ISSUE #957: persistent's quasi quoter wrongly rejects foreign keys that are not primary keys.
[This line](https://github.com/yesodweb/persistent/blob/master/persistent/Database/Persist/Quasi.hs#L227) checks that the foreign key matches the primary key of the referred table. This is needlessly restrictive as foreign keys are allowed to point to unique constraints too.

I was trying to work around persistent's inability to define unique constraints that involve the id key, but now I've been hamstrung by the fact that even if I define them myself I can't define the proper foreign keys because they are different from the primary key...

### LINKED ISSUE #999: Foreign key syntax improvements
Currently, the `persistent` quasiquoter entity definition syntax allows you to define a foreign key with the `Foreign` keyword. The QQ definition and associated table are given here:

```
User
  otherField String
  Foreign OtherTable fkothertable otherField

==>

CREATE TABLE user (
  id PRIMARY AUTOINCREMENT,
  other_field TEXT NOT NULL,
  FOREIGN KEY (other_field) REFERENCES other_table(id)
);
```

This syntax has some important restrictions:

- You can't specify a `CASCADE` behavior
- You can only make a foreign key to the *primary* key of another table, while most SQL databases only require a `Unique` key.

## Proposed Solution:

Immediately, documentation should be updated with these restrictions, and also suggest a means to write these things in raw SQL to work-around the behavior.

Add a `References` keyword to the syntax. After this keyword, accept a non-empty list of field name identifies. If the keyword is missing, assume a reference to the primary key column to retain backwards compatibility.

Add a pair of keywords `OnDelete` and `OnUpdate` which accept another argument, one of `NoAction`, `Restrict`, or `Cascade`. These will translate to the expected migrations.

A user would then be able to write:

```
User
  otherField String
  Foreign OtherTable fkothertable otherField References name OnDelete Cascade OnUpdate Cascade

===>

CREATE TABLE user (
  id PRIMARY AUTOINCREMENT,
  other_field TEXT NOT NULL,
  FOREIGN KEY (other_field) REFERENCES other_table(name) ON DELETE CASCADE ON UPDATE CASCADE
);
```

## Related Issues:

- #909 
- #910
- #957

### LINKED ISSUE #1118: Foreign Key Cascade Actions on "plain" foreign keys
Currently, the `OnCascade` functionality only works on explicit foreign keys.

```
User
  name String
  Primary name

Post
  user UserId
  Foreign User OnDeleteCascade fkuser user
```

We should also be able to write:

```
User
  name String
  Primary name

Post
  user UserId OnDeleteCascade
```

Problem surfaced in #1113 

To implement this, we'll need to start with `takeCols` to parse the `OnDelete`and `OnUpdate` behavior into the column type. However, we don't currently track whether or not this field is a reference yet, so we'll need to figure some other way to propagate the cascade behavior.

- How should we handle cascade behavior on non-foreign key fields? ie:
    ```
    User
        name String OnDeleteCascade
    ```
    Easiest thing to do is ignore it silently. But throwing an `error` may work as well.
- We will probably need to update `FieldDef` to accept the cascade behavior separately from the reference. Then, in `fixForeignkeys`, we'll copy that into the new reference.

### LINKED ISSUE #1119: Explicit Foreign Key to non-Primary table
Currently, if we are writing an explicit foreign key, it is assumed that we're writing to a composite table:

```
User
    name String
    email String
    
    Primary name email

Post
    userName String
    userEmail String
   
    Foreign User fkpostuser userName userEmail
```

This also works (somewhat by accident) using a custom singleton primary key:

```
User
    name String

    Primary name

Post
    user UserId
    Foreign User fkpostuser user
```

However, we can't do this with an automatically generated primary key.

```
User
    name String

Post
    user UserId
    Foreign User fkpostuser user
```

This restriction is artificial. It shouldn't *really* matter since writing `UserId` should cause the relevant migrations to generate foreign keys anyway, but as #1113 surfaced (and #1118 can solve) the only way to write a `FieldCascade` is on an explicit foreign key.

From #1113:

> We should be able to define explicit Foreign declarations to tables that lack an explicit primary key. Database.Persist.Quasi.takeForeign is where this gets parsed out. "no explicit primary key" is found in fixForeignKey. If we get Nothing back from entityPrimaryKey, then we have an automatically generated ID. This means we want to check for the field name being id and generate a reference to the entityId field.

kderme:
This pr adds some new feautures related to Foreign Keys that are commonly needed:
- allows a reference to a table without an explicit Primary key (fixes #1119 ), 
- allows a reference to non Primary keys, using the keyword `References` (fixes #999, fixes #957 )
- allows simple/field Foreign Keys to have Cascade Options (fixes #1118)

The extended syntax for Foreign Key is like this:
`Foreign <otherTable> [OnDelete...] [OnUpdate...] <fkeyName> list_of_foreign_fields [References list_of_parent_fields]`

It's still a _WIP_ because:
- Some postgres specific implementation is missing (see commit Support Cascade options for SQLite )
- Currently when the `References` keyword is used, there is no check for Unique constraint, so it's the user's responsibility to only refer to unique fields. Should we change this? ie we could only reference Unique keys or change the syntax after `References` to accept a unique key name instead of a list of fields. This can be too restrictive though, since the user may ensure uniqueness through other ways (ie unique indexes) which we can't check. Any opinions?

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE ViewPatterns #-}
 {-# LANGUAGE NamedFieldPuns #-}
@@ -849,7 +850,7 @@ getColumns getter def cols = do
     us <- with (stmtQuery stmt' vals) (\src -> runConduit $ src .| helperU)
     return $ cs ++ us
   where
-    refMap = Map.fromList $ foldl' ref [] cols
+    refMap = fmap (\cr -> (crTableName cr, crConstraintName cr)) $ Map.fromList $ foldl' ref [] cols
         where ref rs c = case cReference c of
                   Nothing -> rs
                   (Just r) -> (unDBName $ cName c, r) : rs
@@ -942,7 +943,8 @@ getColumn getter tableName' [PersistText columnName, PersistText isNullable, Per
                           , cDefault = fmap stripSuffixes d''
                           , cDefaultConstraintName = Nothing
                           , cMaxLen = Nothing
-                          , cReference = ref
+                          -- TODO: Fix cascade reference is ignored
+                          , cReference = fmap (\(a,b) -> ColumnReference a b noCascade) ref
                           }
   where
     stripSuffixes t =
@@ -1053,9 +1055,9 @@ findAlters defs edef col@(Column name isNull sqltype def _defConstraintName _max
             ([(name, Add' col)], cols)
         Just (Column _oldName isNull' sqltype' def' _defConstraintName' _maxLen' ref') ->
             let refDrop Nothing = []
-                refDrop (Just (_, cname)) = [(name, DropReference cname)]
+                refDrop (Just ColumnReference {crConstraintName=cname}) = [(name, DropReference cname)]
                 refAdd Nothing = []
-                refAdd (Just (tname, a)) =
+                refAdd (Just ColumnReference {crTableName=tname, crConstraintName=a}) =
                     case find ((==tname) . entityDB) defs of
                         Just refdef
                             | entityDB edef /= tname
@@ -1066,6 +1068,7 @@ findAlters defs edef col@(Column name isNull sqltype def _defConstraintName _max
                                     a
                                     [name]
                                     (Util.dbIdColumnsEsc escape refdef)
+                                    -- TODO: Fix cascade reference is ignored
                                     noCascade
                               )
                             ]
@@ -1073,7 +1076,7 @@ findAlters defs edef col@(Column name isNull sqltype def _defConstraintName _max
                         Nothing ->
                             error $ "could not find the entityDef for reftable[" ++ show tname ++ "]"
                 modRef =
-                    if fmap snd ref == fmap snd ref'
+                    if fmap crConstraintName ref == fmap crConstraintName ref'
                         then []
                         else refDrop ref' ++ refAdd ref
                 modNull = case (isNull, isNull') of
@@ -1113,13 +1116,14 @@ getAddReference
     :: [EntityDef]
     -> EntityDef
     -> DBName
-    -> (DBName, DBName)
+    -> ColumnReference
     -> Maybe AlterDB
-getAddReference allDefs entity cname (s, constraintName) = do
+getAddReference allDefs entity cname ColumnReference {crTableName = s, crConstraintName=constraintName} = do
     guard $ table /= s && cname /= fieldDB (entityId entity)
     pure $ AlterColumn
         table
         ( s
+        -- TODO: Fix cascade reference is ignored
         , AddReference constraintName [cname] id_ noCascade
         )
   where
--- a/persistent-template/Database/Persist/TH.hs
+++ b/persistent-template/Database/Persist/TH.hs
@@ -290,7 +290,7 @@ data FieldSqlTypeExp = FieldSqlTypeExp FieldDef SqlTypeExp
 
 instance Lift FieldSqlTypeExp where
     lift (FieldSqlTypeExp FieldDef{..} sqlTypeExp) =
-        [|FieldDef fieldHaskell fieldDB fieldType $(lift sqlTypeExp) fieldAttrs fieldStrict fieldReference fieldComments|]
+        [|FieldDef fieldHaskell fieldDB fieldType $(lift sqlTypeExp) fieldAttrs fieldStrict fieldReference fieldComments fieldCascadeOpts|]
 #if MIN_VERSION_template_haskell(2,16,0)
     liftTyped = unsafeTExpCoerce . lift
 #endif
@@ -1694,18 +1694,22 @@ liftAndFixKeys entityMap EntityDef{..} =
     |]
 
 liftAndFixKey :: EntityMap -> FieldDef -> Q Exp
-liftAndFixKey entityMap (FieldDef a b c sqlTyp e f fieldRef mcomments) =
-    [|FieldDef a b c $(sqlTyp') e f fieldRef' mcomments|]
+liftAndFixKey entityMap (FieldDef a b c sqlTyp e f fieldRef mcomments casc) =
+    [|FieldDef a b c $(sqlTyp') e f fieldRef' mcomments casc|]
   where
     (fieldRef', sqlTyp') = fromMaybe (fieldRef, lift sqlTyp) $
       case fieldRef of
         ForeignRef refName _ft -> case M.lookup refName entityMap of
-          Nothing -> Nothing
+          Nothing -> checkCascade
           Just ent ->
             case fieldReference $ entityId ent of
               fr@(ForeignRef _Name ft) -> Just (fr, lift $ SqlTypeExp ft)
-              _ -> Nothing
-        _ -> Nothing
+              _ -> checkCascade
+        _ -> checkCascade
+    checkCascade = case casc of
+      FieldCascade Nothing Nothing -> Nothing
+      _ -> error $ "cascade field is not allown for field " <> show a
+           <> ". It doesn't reference any other tables."
 
 deriving instance Lift EntityDef
 
--- a/persistent-template/Database/Persist/TH.hs
+++ b/persistent-template/Database/Persist/TH.hs
@@ -1335,7 +1335,8 @@ mkLenses mps ent = fmap mconcat $ forM (entityFields ent) $ \field -> do
         ]
 
 mkForeignKeysComposite :: MkPersistSettings -> EntityDef -> ForeignDef -> Q [Dec]
-mkForeignKeysComposite mps t ForeignDef {..} = do
+mkForeignKeysComposite mps t ForeignDef {..} =
+    if not foreignToPrimary then return [] else do
     let fieldName f = mkName $ unpack $ recName mps (entityHaskell t) f
     let fname = fieldName foreignConstraintNameHaskell
     let reftableString = unpack $ unHaskellName foreignRefTableHaskell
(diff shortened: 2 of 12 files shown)


parsonsmatt:
:raised_hands: my man :raised_hands:

persistent/ChangeLog.md
```diff
@@ -2,6 +2,9 @@
 
 ## (Unreleased) 2.11.0.0
 
+* It is now supported to refer to a table with an auto generated Primary Kay
+* It is now supported to refer to non-primary fields, using the keyword `References`
+* It is now supported to have cascade options for simple/single-field Foreign Keys
```
parsonsmatt: For these changelog entries, can we group by the PR?

persistent/Database/Persist/Types/Base.hs
```diff
@@ -300,6 +313,10 @@ data ForeignDef = ForeignDef
     , foreignFields                :: ![(ForeignFieldDef, ForeignFieldDef)] -- this entity plus the primary entity
     , foreignAttrs                 :: ![Attr]
     , foreignNullable              :: Bool
+    , foreignToPrimary             :: Bool
+    -- ^ Determines if the reference is towards a Primary Key or not.
+    --
+    -- @since 2.11.0
```
parsonsmatt, line 319: :tada:

persistent/Database/Persist/Sql/Types.hs
```diff
@@ -25,7 +27,7 @@ data Column = Column
     , cDefault   :: !(Maybe Text)
     , cDefaultConstraintName   :: !(Maybe DBName)
     , cMaxLen    :: !(Maybe Integer)
-    , cReference :: !(Maybe (DBName, DBName)) -- table name, constraint name
+    , cReference :: !(Maybe ((DBName, DBName), FieldCascade)) -- table name, constraint name, cascade options
```
parsonsmatt, line 30: I kinda wish this were a different datatype rather than a mash of tuples. Any opposition to writing a new datatype with field accessors instead of a tuple?

erikd:
Thanks for this @kderme !

This actually compiles for me, but I have since run into another problem.

I am storing a block chain in PostgreSQL and I have a `Block` table, the simplified version of which is:
```
  Block
    hash                ByteString
    previous            BlockId Maybe
    UniqueBlock         hash
```
Note that there is no delete/cascade defined for this table, but the table does reference iteself.

When I try to **write** something to this table using Psersistents `insertBy` I get an error:
```
Exception:
  SqlError
    { sqlState = "23503"
    , sqlExecStatus = FatalError
    , sqlErrorMsg = "update or delete on table \"block\" violates foreign key constraint \"block_previous_fkey\" on table \"block\""
    , sqlErrorDetail = "Key (id)=(89) is still referenced from table \"block\"."
    , sqlErrorHint = ""
    }
```
Using the old Persistent `mkDeleteCascade` method on the schema instead of the new `OnDeleteCascade`, this worked correctly.

parsonsmatt:
That's intriguing. `insertBy` does not call any update or delete methods. Can you post a more complete example?

erikd:
How do I even run the `persistent-postgresql` tests? I am currently getting:
```
Test suite test: RUNNING...
)
Test suite test: FAIL
```
As for reproducing this I *think* it should be a simple as:
```
diff --git a/persistent-test/src/ForeignKey.hs b/persistent-test/src/ForeignKey.hs
index aaa039e7..081a743b 100644
--- a/persistent-test/src/ForeignKey.hs
+++ b/persistent-test/src/ForeignKey.hs
@@ -18,6 +18,7 @@ share [mkPersist persistSettings { mpsGeneric = False }, mkMigrate "compositeMig
 
   ParentImplicit
     name String
+    backRef ParentImplicit Maybe
 
   ChildImplicit
     pname String
```

kderme:
@erikd I think your best bet is to use `OnDeleteSetNull` since you have nullable foreign keys. If there are no cascade options and the parent key is deleted, the error you see will be thrown (although I'm not sure why it's thrown when you insert a Block).

Unfortunately there seems to be an issue with nullable foreign keys, I'll try to fix later today.

kderme:
I fixed the postgres tests. The bug was that with https://github.com/yesodweb/persistent/issues/1119 2 foreign keys are added. I had to add the `noreference` attribute for the tests to pass. SQLite seems to work properly even without the attribute.

erikd:
I must be doing something wrong.

I have what is basically:
```
  Block
    hash                ByteString
    previous            BlockId Maybe
    Foreign             Block               OnDeleteSetNull fkBlock previous References Id
    UniqueBlock         hash
```
and when I try the `delete` on the `block` table I get:
```
SqlError
  { sqlState = "23503"
  , sqlExecStatus = FatalError
  , sqlErrorMsg = "update or delete on table \"block\" violates foreign key constraint \"block_previous_fkey\" on table \"block\""
  , sqlErrorDetail = "Key (id)=(29989) is still referenced from table \"block\"."
  , sqlErrorHint = ""
  }
```
The generated SQL is:
```
EXECUTE 'ALTER TABLE "block" ADD CONSTRAINT "block_previous_fkey" FOREIGN KEY("previous")
    REFERENCES "block"("id")' ;
```

kderme:
@erikd I think you should add the `noreference` attribute, right after Maybe to make it work. See also the tests I added on my last commit.

parsonsmatt:
@erikd It should be possible to replace your declaration with something like:

```
  Block
    hash                ByteString
    previous            BlockId Maybe   OnDeleteSetNull fkBlock previous References Id
    UniqueBlock         hash
```

Your current declaration appears to have *two* foreign keys.

Perhaps we can have a separate check for a duplicate foreign key - eg, if we have a field `foo FooId` and also a `Foreign Foo fooId` then it won't generate a duplicate - that would solve the issue of two nearly identical keys with differeing cascade behaviors.

parsonsmatt:
> How do I even run the `persistent-postgresql` tests? I am currently getting:
> 
> ```
> > cabal test persistent-postgresql
> ....
> Test suite test: RUNNING...
> test: libpq: failed (fe_sendauth: no password supplied
> )

Edit `pg_hba.conf` and set the local connections to `trust` instead of `md5`. Apparently the config is not using a password/username in the setting and instead using peer authentication. Or we can patch the test suite to use password/usernames.

kderme:
> @erikd It should be possible to replace your declaration with something like:
> 
> ```
>   Block
>     hash                ByteString
>     previous            BlockId Maybe   OnDeleteSetNull fkBlock previous References Id
>     UniqueBlock         hash
> ```
> 



This doesn't work yet on postgres, but I plan to do it as part of this pr.

> Perhaps we can have a separate check for a duplicate foreign key - eg, if we have a field foo FooId and also a Foreign Foo > fooId then it won't generate a duplicate - that would solve the issue of two nearly identical keys with differeing cascade behaviors.

Yes probably that's the correct solution.

erikd:
There are a couple of things for me to look at here, but unfortunately I will not get back to this until later this week.

erikd:
Still have not managed to get back to testing this :/ .

parsonsmatt:
I'm seeing that we have a green check on most recent CI with a commit ~10 days ago - let me review it and see if it's all set to go.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -942,7 +943,8 @@ getColumn getter tableName' [PersistText columnName, PersistText isNullable, Per
                           , cDefault = fmap stripSuffixes d''
                           , cDefaultConstraintName = Nothing
                           , cMaxLen = Nothing
-                          , cReference = ref
+                          -- TODO: Fix cascade reference is ignored
```
parsonsmatt, line 946: Is this TODO still relevant?

persistent-template/Database/Persist/TH.hs
```diff
@@ -1689,18 +1694,22 @@ liftAndFixKeys entityMap EntityDef{..} =
...
           Just ent ->
             case fieldReference $ entityId ent of
               fr@(ForeignRef _Name ft) -> Just (fr, lift $ SqlTypeExp ft)
-              _ -> Nothing
-        _ -> Nothing
+              _ -> checkCascade
+        _ -> checkCascade
+    checkCascade = case casc of
+      FieldCascade Nothing Nothing -> Nothing
+      _ -> error $ "cascade field is not allown for field " <> show a
```
parsonsmatt, line 1711:
```suggestion
      _ -> reportError $ "cascade field is not allowed for field " <> show a
```

`reportError` is a slightly nicer facility for reporting errors in the `Q` Template Haskell type. Would require refactoring this up a bit - probably:

```haskell
liftAndFixKey ... casc) = do
    (fieldRef', sqlTyp') <- fromMaybe (fieldRef, lift sqlTyp) <$> checkCascade
    [|FieldDef 
  where
    checkCascade casc =
        case fieldRef of
            ForeignRef refName _ft -> ...
                Nothing -> 
                    case casc of ...
```

persistent-test/src/ForeignKey.hs
```diff
@@ -33,6 +42,45 @@ share [mkPersist persistSettings { mpsGeneric = False }, mkMigrate "compositeMig
...
+
+  Chain
+    name String
+    previous ChainId Maybe noreference
+    Foreign Chain OnDeleteSetNull fkChain previous References Id
+    deriving Show Eq
+
+  Chain2
+    name String
+    previous Chain2Id Maybe noreference
```
parsonsmatt, line 81: Okay, so we're still using `noreference` as a work-around to get the cascades working I take it

persistent-test/src/ForeignKey.hs
```diff
@@ -64,3 +119,50 @@ specsWith runDb = describe "foreign keys options" $ do
...
+    let expected = [Entity {entityKey = ChainKey 2, entityVal = Chain "B" Nothing}]
+    cs @== expected
+  it "deletes cascades with self reference to the whole chain" $ runDb $ do
+    k1 <- insert $ Chain2 "A" Nothing
+    k2 <- insert $ Chain2 "B" (Just k1)
+    k3 <- insert $ Chain2 "C" (Just k2)
+    delete k1
+    cs <- selectList [] []
+    let expected = [] :: [Entity Chain2]
+    cs @== expected
```
parsonsmatt, line 168: Wonderful tests, thanks <3

persistent/Database/Persist/Types/Base.hs
```diff
@@ -139,6 +139,14 @@ data EntityDef = EntityDef
     }
     deriving (Show, Eq, Read, Ord)
 
+entitiesPrimary :: EntityDef -> Maybe [FieldDef]
+entitiesPrimary t = case fieldReference primaryField of
+    CompositeRef c -> Just $ (compositeFields c)
```
parsonsmatt, line 144:
```suggestion
    CompositeRef c -> Just (compositeFields c)
```

Issue #995, opened 2019-12-19
Generalizing `insert` and gracefully handling defaults


parsonsmatt:
Raise your hand if you've been frustrated by needing to specify a time or use `Maybe` for `createdAt UTCTime` fields.

If you're one of the few folks that hasn't, well, let me introduce the problem.

```
User
  name String
```

This is fine and easy. To insert a new user into the database, we write `insert User { userName = "Matt" }`. There is an implied surrogate key that is associated, and it should be an auto-incrementing integer. Because it has a `default` in the database, we don't need to specify it. This pattern is so common that we have the `Entity` type, which includes the `Key entity` for the `entity`.

Then we want to record when a user is created.

```
User
  name String
  createdAt UTCTIme
```

Database users are accustomed to writing a schema like:

```sql
CREATE TABLE user (
  id   SERIAL PRIMARY KEY,
  name VARCHAR NOT NULL,
  created_at TIMESTAMP NOT NULL DEFAULT=NOW()
);
```

And the `persistent` library happily supports the `default=` syntax, which does The Right Thing with migrations.

```
User
  name String
  createdAt UTCTIme default=NOW()
```

Unfortunately, we reach a problem when we go to insert a new value. `insert`'s type is `insert :: entity -> SqlPersistT m (Key entity)`. And `userCreatedAt :: UTCTime` - it's a required field!! So now we have two options:

1. Make the timestamp in Haskell and forego the database default, writing:

    ```haskell
    newUser :: String -> SqlPersistT m UserId
    newUser userName = do
      userCreatedAt <- liftIO getCurrentTime
      insert User {..}
    ```

  But this gets really annoying as the `User` gets additional arguments, and people *really* don't like writing this out when the database defaulting mechanism is designed to provide exactly this.
2. Make the definition nullable and provide `Nothing`:
    
    ```haskell
    insert User { userName = "Matt", userCreatedAt = Nothing }
    ```
    The database defaulting mechanism works out here, hooray. But now we have to care about `Maybe` at every use site! Gross.

So here's my plan:

1. Extend the `PersistEntity` class with an associated type `New`:
    ```haskell
    class PersistEntity entity where
        type New entity :: *
    ```
2. Change the signature of `insert` to be: `insert :: (PersistEntity entity) => New entity -> SqlPersistT m (Key entity)
3. For a definition with no `default=` clauses, define `type New User = User`
4. For a definition with `default=` clauses, 
[...]

Issue #875, opened 2019-04-02
Experimental/unmaintained backends: couchdb/zookeeper not currently supported/built


parsonsmatt:
`persistent-zookeeper` is not in the `stack.yaml` and was last touched 3 years ago.

@junjihashimoto is listed as the maintainer for the package `persistent-zookeeper`. Would you be willing to put `persistent-zookeeper` in a repository you control/maintain? Currently, the `.cabal` file points to your fork on GitHub.

@MaxGabriel looks like you're the last person to touch the `experimental/couchdb` stuff. Any opposition to deleting it?

parsonsmatt:
I am going to be deleting `persistent-zookeeper` and `experimental/couchdb` out of this repository on April 2nd, 2020.

junjihashimoto:
@parsonsmatt 
Thank you very much for contacting me, and sorry for bothering you.
I will maintain `persistent-zookeeper` in my own repository.

junjihashimoto:
I think it is good to test persistent-zookeeper by using persistent-test, but I'm sorry I couldn't maintain it. 
It seems that persistent-test is listed in hackage, so I will try to test using it.

parsonsmatt:
Thank you for maintaining it @junjihashimoto! There is no bother. If you have any problems with `persistent-test`, please let me know. I want `persistent-test` to be useful to everyone :)

PR #1122, opened 2020-09-09
PersistLiteral support for SQL keywords

ldub:
When mapping to a generated column such as the following, the `toPersistValue` must output `DEFAULT`, without quotes. 

```
example_column text GENERATED ALWAYS AS (COALESCE(field_one, field_two)) STORED
```

Currently, `PersistDbSpecific` will quote things it outputs (in persistent-postgresql), but an insert of `'DEFAULT'` would be rejected by Postgres.

This PR adds support for a `PersistLiteral` constructor on `PersistValue`, meant to explicitly never escape its output in any database system. I've added an implementation for MySQL, PostgreSQL, and sqlite, as well as a test that uses a schema with a GENERATED column. My colleague @friedbrice said he would be happy to look into implementing this for the other db systems that I haven't added support for.

A note on tests:
`GENERATED` columns are supported in MySQL, PostgreSQL, and sqlite. While sqlite's [documentation](https://sqlite.org/gencol.html) says its available from 3.31.0, my testing shows that it does not actually work in 3.31.0:

<details> <summary>sqlite version testing output</summary>

does not work with current sqlite 3.31.0

```
❯ /usr/local/opt/sqlite/bin/sqlite3 --version
3.30.1 2019-10-10 20:19:45 18db032d058f1436ce3dea84081f4ee5a0f2259ad97301d43c426bc7f3df1b0b

❯ /usr/local/opt/sqlite/bin/sqlite3
SQLite version 3.30.1 2019-10-10 20:19:45
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> CREATE TABLE persist_literal_field_test_table (id serial primary key, field_one text, field_two text, field_three text GENERATED ALWAYS AS (COALESCE(field_one, field_two)) STORED);
Error: near "AS": syntax error
sqlite>
```

works with newer sqlite 3.33.0
```
❯ /usr/local/opt/sqlite/bin/sqlite3 --version
3.33.0 2020-08-14 13:23:32 fca8dc8b578f215a969cd899336378966156154710873e68b3d9ac5881b0ff3f

❯ sqlite3
SQLite version 3.33.0 2020-08-14 13:23:32
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> CREATE TABLE persist_literal_field_test_table (id serial primary key, field_one text, field_two text, field_three text GENERATED ALWAYS AS (COALESCE(field_one, field_two)) STORED);
sqlite>
```

</details>

If the sqlite.c "amalgamation" were to be updated to the latest 3.33.0, the test for sqlite will work. Currently it fails. Something need to be done about that and I would appreciate suggestions.

After submitting your PR:


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -552,6 +554,19 @@ instance PGFF.FromField Unknown where
 instance PGTF.ToField Unknown where
     toField (Unknown a) = PGTF.Escape a
 
+newtype UnknownLiteral = UnknownLiteral { unUnknownLiteral :: ByteString }
+  deriving (Eq, Show, Read, Ord, Typeable)
```
MaxGabriel, line 558:
```suggestion
  deriving (Eq, Show, Read, Ord)
```

Typeable is automatically derived since iirc GHC 7.10

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -552,6 +554,19 @@ instance PGFF.FromField Unknown where
...
+
+instance PGFF.FromField UnknownLiteral where
```
MaxGabriel, line 560: Is this instance used?

persistent-test/src/PersistLiteralTestSQL.hs
```diff
@@ -0,0 +1,30 @@
...
+    fieldTwo Text
+    fieldThree (NullableGenerated Text)
+    deriving Show Eq
+|]
+
+specsWith :: Runner SqlBackend m => RunDb SqlBackend m -> Spec
+specsWith runDB = describe "PersistLiteral field" $ do
+  it "should read a GENERATED column" $ runDB $ do
+    -- we have to manage the table manually because of the generated field
+    _ <- rawExecute "CREATE TABLE persist_literal_field_test_table (id serial primary key, field_one text, field_two text, field_three text GENERATED ALWAYS AS (COALESCE(field_one, field_two)) STORED);" []
```
MaxGabriel, line 23: Possibly do a DROP TABLE IF EXISTS before this line, rather than a DROP TABLE after? If this test failed part way through, future runs of the test would fail in the current code, if I understand right

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -237,6 +237,7 @@ instance MySQL.Param P where
       MySQL.Plain $ BBB.fromString $ show (fromRational r :: Pico)
       -- FIXME: Too Ambiguous, can not select precision without information about field
     render (P (PersistDbSpecific s))    = MySQL.Plain $ BBS.fromByteString s
+    render (P (PersistLiteral l))     = MySQL.Plain $ BBS.fromByteString l
```
MaxGabriel, line 240: I’m confused how this can be the same implementation as PersistDbSpecific

persistent-sqlite/Database/Sqlite.hs
```diff
@@ -468,6 +468,7 @@ bind statement sqlData = do
             PersistList l -> bindText statement parameterIndex $ listToJSON l
             PersistMap m -> bindText statement parameterIndex $ mapToJSON m
             PersistDbSpecific s -> bindText statement parameterIndex $ decodeUtf8With lenientDecode s
+            PersistLiteral l -> bindText statement parameterIndex $ decodeUtf8With lenientDecode l
```
MaxGabriel, line 471: I’m confused how this can be the same implementation as PersistDbSpecific

MaxGabriel:
This would also work for calling functions, right? eg if I want to set a column to the database’s NOW() value or some other function?

MaxGabriel:
This is a fairly dangerous tool, since any dynamic value opens you up to SQL injection, right? Thoughts on adding more warning flags, or a name like UnsafeUnescapedLiteral?

friedbrice:
> This would also work for calling functions, right? eg if I want to set a column to the database’s NOW() value or some other function?

Yes, it would be governed by specific instances of `PersistField`. For example

```haskell
newtype ComputedColumn a = ComputedColumn a

instance PersistField a => PersistField (ComputedColumn a) where
    toPersistValue _ = PersistLiteral "DEFAULT"
    fromPersistValue x = ComputedColumn <$> fromPersistValue x

instance PersistFieldSql a => PersistFieldSql (ComputedColumn a) where
    sqlType _ = sqlType (Proxy @a)
```

And that's just because that's how Postgres allows you to insert a row into a table that has a computed column.

If you wanted to use `now()`, it'd be

```
newtype Timestamp = Timestamp UTCTime

instance PersistField Timestamp where
    toPersistValue _ = PersistLiteral "now()"
    fromPersistValue x = Timestamp <$> fromPersistValue x

instance PersistFieldSql Timestamp where
    sqlType _ = sqlType (Proxy @UTCTime)
```

friedbrice:
> This is a fairly dangerous tool, since any dynamic value opens you up to SQL injection, right? Thoughts on adding more warning flags, or a name like UnsafeUnescapedLiteral?
> 
> (Is it possible to call the escaping function at all for this, eg to implement something like ARRAY(‘a’), or can this only be used for static values (eg DEFAULT)?

In both use cases I give above, the literal happens to be completely static. I wonder if this suggests that a more-limited API would be better. Perhaps there's a way to use a quasiquote to ensures that the implementation of `toPersistValue` is completely static.

persistent-test/src/PersistLiteralTestSQL.hs
ldub: For some reason I thought that `IF EXISTS` was not standard SQL. Just checked and it is at least supported by Postgres, MySQL, and SQLite, so I will implement this.

friedbrice:
@parsonsmatt I see you lurking in this thread. IIRC, you were working on something for timestamps. I'm interested to hear how this concept relates to your work. Does your work subsume this concept? Are they orthogonal?

ldub:
In persistent currently, PersistDbSpecific only escapes values in persistent-postgresql but not persistent-mysql or persistent-sqlite. This explains @MaxGabriel 's questions about "I’m confused how [PersistLiteral can have] the same implementation as PersistDbSpecific" in the SQLite and MySQL implementations.

It may be that is instead a bug in persistent-postgresql's implementation of PersistDbSpecific. I thought this was not the case because the documentation shows an [example of using it with PostGIS](https://hackage.haskell.org/package/persistent-2.10.5/docs/Database-Persist-Types.html#t:PersistValue), however on second reading I think that implementation is broken because both `toPersistValue` and `toPoint` will escape the `POINT(44 44)` literal with single quotes.

I will try to demonstrate that this is broken by adding a test.

persistent-sqlite/Database/Sqlite.hs
ldub: see comment below

persistent-mysql/Database/Persist/MySQL.hs
ldub: see comment below https://github.com/yesodweb/persistent/pull/1122#issuecomment-689707949

parsonsmatt:
So, having a `UnsafeUnescaped` thing in `PersistValue` is necessary and useful, IMO. 

Using it for computed columns is very, very, very interesting, but I think we need to be cautious about exactly how it is used. `toPersistValue` is not only used with `insert` but also in a few other cases. Consider an embedded entity:

```
Thing
    name String
    createdAt Timestamp

HasStuff
  thing Thing
```

The embedded `Thing` uses the `PersistMap` constructor:

```
entityToPersistValueHelper :: (PersistEntity record) => record -> PersistValue
entityToPersistValueHelper entity = PersistMap $ zip columnNames fieldsAsPersistValues
    where
        columnNames = map (unHaskellName . fieldHaskell) (entityFields (entityDef (Just entity)))
        fieldsAsPersistValues = map toPersistValue $ toPersistFields entity
```

Which means we get `toPersistValue` called on the entity before embedding it. This means that performing this insert:

```
insert (HasStuff (Thing "hello" (Timestamp dummyUTCTime)))
```

will result in the following record in the database:

```
sql> select * from has_stuff;

id | thing
---+-----
1  | {"thingName":"hello","thingCreatedAt":"now()"}
```

which is really not what we want.

friedbrice:
> Using it for computed columns is very, very, very interesting, but I think we need to be cautious about exactly how it is used. `toPersistValue` is not only used with `insert` but also in a few other cases.

My first instinct is the following workaround:

```
class PersistField a where
    toPersistValue :: a -> PersistValue
    fromPersistValue :: a -> Either Text PersistValue

    insertPersistValue :: a -> PersistValue
    insertPersistValue = toPersistValue
```

So users could provide their own `insertPersistValue` that does special database logic. I feel like this is a Bad Idea™, but I can't really see why immediately. What are people's thoughts?

friedbrice:
With the above class definition, we'd change `insert` so that it used `insertPersistValue` instead of `toPersistValue`, and we'd write our `PersistField` instances as so:

```haskell
newtype ComputedColumn a = ComputedColumn a

instance PersistField a => PersistField (ComputedColumn a) where
    toPersistValue (ComputedColumn x) = toPersistValue x
    fromPersistValue x = ComputedColumn <$> fromPersistValue x
    insertPersistValue _ = PersistLiteral "DEFAULT"

instance PersistFieldSql a => PersistFieldSql (ComputedColumn a) where
    sqlType _ = sqlType (Proxy @a)


newtype Timestamp = Timestamp UTCTime

instance PersistField Timestamp where
    toPersistValue (Timestamp x) = toPersistValue x
    fromPersistValue x = Timestamp <$> fromPersistValue x
    insertPersistValue _ = PersistLiteral "now()"

instance PersistFieldSql Timestamp where
    sqlType _ = sqlType (Proxy @UTCTime)
```

parsonsmatt:
The thing with tracking `insert` writes is in here: https://github.com/yesodweb/persistent/issues/995

I don't think that adding an additional responsibility to `PersistField` is appropriate

persistent-test/src/NullableGenerated.hs
```diff
@@ -0,0 +1,20 @@
...
+module NullableGenerated where
+
+import qualified Data.Proxy as P (Proxy(..))
+import Init
+
+newtype NullableGenerated a = NullableGenerated (Maybe a) deriving (Show, Eq)
+
+instance forall a. PersistField a => PersistField (NullableGenerated a) where
+  toPersistValue (NullableGenerated _) = PersistLiteral "DEFAULT"
+  fromPersistValue g = NullableGenerated <$> fromPersistValue g
```
parsonsmatt, line 14: This is a useful pattern for describing how this works but I really can't recommend using it in app code. Having it present in the tests is a bit worrying to me, as well, as other people might mimic it and then come across bugs for it.

parsonsmatt (approved):
I like this approach a lot.

parsonsmatt:
Travis is failing because that version of Postgres doesn't support GENERATED ,looks like. We can either upgrade Travis or not use GENERATED in the schema - slight preference to drop GENERATED since that'd make it easier to run tests locally too for older OSes.

parsonsmatt:
I've made a PR that will add GENERATED column support, if Travis accepts it: #1141 Once that has landed, if you merge `master`, then it should work, and we can get this released :smile:

parsonsmatt:
Actually, I have switched CI to GitHub Actions with postgres 12 installed. Can you merge `master`? That should kick off the right CI which should get this verified in.

persistent/Database/Persist/Types/Base.hs
```diff
@@ -377,6 +377,7 @@ data PersistValue = PersistText Text
                   | PersistMap [(Text, PersistValue)]
                   | PersistObjectId ByteString -- ^ Intended especially for MongoDB backend
                   | PersistArray [PersistValue] -- ^ Intended especially for PostgreSQL backend for text arrays
+                  | PersistLiteral ByteString -- ^ Using 'PersistLiteral' you can customize PersistField instances to output unescaped SQL
                   | PersistDbSpecific ByteString -- ^ Using 'PersistDbSpecific' allows you to use types specific to a particular backend
```
parsonsmatt, line 381: Okay, here's a migration plan that can add this feature safely *and* manage #1123

1. `PersistLiteral` is introduced as *the* "unescaping" variant.
2. We add a constructor `PersistLiteralEscaped` that has the escaping behavior consistently across DB packages.
3. Add a deprecation notice to `PersistDbSpecific` specifying what's changed and how. Mention that it will be removed in the next major version release of the library. This warning should percolate through the ecosystem and packages will update.

Then, in the next major version, we can make the behavior breaking changes in the relevant database packages, unifying how postgres/mysql/sqlite use these constructors.

friedbrice:
@parsonsmatt I'm working on getting it up to date and making sure all the tests pass. Will merge when finished. Thanks!

friedbrice:
@parsonsmatt It looks like I'm not going to be able to get this to work with Sqlite in its current state. We can do a few things: (1) merge and just say Sqlite generated columns are not supported [i'm not too keen on that one], (2) rework the feature, probably adding a field to `EntityDef` and pushing the merge out to the next major release. What are your thoughts?

parsonsmatt:
2.11 is going to be a major release and so additional fields on `EntityDef` or similar are totally fine :) Can you describe the issues you're running into with it?

friedbrice:
Yeah, we're using this `PersistLiteral` for generated fields, right? Say you have a table `table foo (foo1 text, foo2 text generated as 'always this exact text')`. By hand, you'd insert like so: `insert into foo(foo1) values ('yup');` But since we're doing things programatically, we generate a prepared statement that looks like this: `insert into foo(foo1, foo2) values (?,?);`. This seems like it should be broken right away, since we're specifying a generated column. However, Postgres and MySQL allow the literal `DEFAULT`, and so to emulate generated columns, we write the PersistValue instance to always serialize anything as `DEFAULT`, and the works for us, in MySQL and Postgres. However, in Sqlite, the prepared statement itself is an error, and that database stops your right there in your tracks. You're not allowed to specify a generate field, even if you use the literal `DEFAULT` keyword.

So, we'll probably want to keep `PersistLiteral` for other reasons, but stop using it for default columns. For default columns, we'll instead want to add a field to `EntityDef` that says whether or not a field is generated and allows the various backends to case on that field explicitly themselves.

friedbrice:
I pushed my latest working tree, but it won't get past CI. You'll find `persistent-sqlite` will fail on `PersistLiteralTestSQL`.

friedbrice:
The failed test can be fixed in `Database.Persist.Sqlite.insertSql'` if `EntityDef` had an extra field that `insertSql'` could case on when building the its statement. (See https://github.com/ldub/persistent/blob/5e9fc69ab0e80a5d6ec1cbd77e13e461671b480c/persistent-sqlite/Database/Persist/Sqlite.hs#L327)

persistent/Database/Persist/Sql/Internal.hs
```diff
@@ -28,14 +30,20 @@ data BackendSpecificOverrides = BackendSpecificOverrides
...
 -- | Creates an empty 'BackendSpecificOverrides' (i.e. use the default behavior; no overrides)
 --
 -- @since 2.11
 emptyBackendSpecificOverrides :: BackendSpecificOverrides
 emptyBackendSpecificOverrides = BackendSpecificOverrides Nothing
 
-defaultAttribute :: [Attr] -> Maybe Text
-defaultAttribute = listToMaybe . mapMaybe (T.stripPrefix "default=")
+defaultAttribute :: [FieldAttr] -> Maybe Text
+defaultAttribute = findMaybe (\case {FieldAttrDefault x -> Just x; _ -> Nothing})
```
parsonsmatt, line 43: Please write this out as a multi-line case expression

```suggestion
defaultAttribute = findMaybe $ \case 
    FieldAttrDefault x -> Just x
     _ -> Nothing
```

persistent/Database/Persist/Sql/Internal.hs
```diff
@@ -28,14 +30,20 @@ data BackendSpecificOverrides = BackendSpecificOverrides
...
+
+generatedAttribute :: [FieldAttr] -> Maybe Text
+generatedAttribute = findMaybe (\case {FieldAttrGenerated x -> Just x; _ -> Nothing})
```
parsonsmatt, line 46:
```suggestion
generatedAttribute = findMaybe $ \case 
    FieldAttrGenerated x -> Just x
    _ -> Nothing
```

persistent/Database/Persist/Sql/Internal.hs
```diff
@@ -100,39 +110,34 @@ mkColumns allDefs t overrides =
...
-    maxLen [] = Nothing
-    maxLen (a:as)
-        | Just d <- T.stripPrefix "maxlen=" a =
-            case reads (T.unpack d) of
-              [(i, s)] | all isSpace s -> Just i
-              _ -> error $ "Could not parse maxlen field with value " ++
-                           show d ++ " on " ++ show tableName
-        | otherwise = maxLen as
+    maxLen :: [FieldAttr] -> Maybe Integer
+    maxLen = findMaybe (\case {FieldAttrMaxlen n -> Just n; _ -> Nothing})
```
parsonsmatt, line 120:
```suggestion
    maxLen = findMaybe $ \case 
        FieldAttrMaxlen n -> Just n
         _ -> Nothing
```

persistent/Database/Persist/Types/Base.hs
```diff
@@ -165,14 +167,46 @@ newtype DBName = DBName { unDBName :: Text }
 
 type Attr = Text
 
+data FieldAttr
```
parsonsmatt, line 170: I like this! This is a big improvement.

```suggestion
-- | Documentation goes here please
--
-- @since 2.11.0.0
data FieldAttr
```

persistent/Database/Persist/Types/Base.hs
```diff
@@ -165,14 +167,46 @@ newtype DBName = DBName { unDBName :: Text }
...
+    | FieldAttrReference Text
+    | FieldAttrConstraint Text
+    | FieldAttrDefault Text
+    | FieldAttrGenerated Text
+    | FieldAttrSqltype Text
+    | FieldAttrMaxlen Integer
+    | FieldAttrOther Text
+    deriving (Show, Eq, Read, Ord)
+
+parseFieldAttrs :: [Text] -> [FieldAttr]
```
parsonsmatt, line 185:
```suggestion
-- |
--
-- @since 2.11.0.0
parseFieldAttrs :: [Text] -> [FieldAttr]
```

persistent/Database/Persist/Types/Base.hs
```diff
@@ -165,14 +167,46 @@ newtype DBName = DBName { unDBName :: Text }
...
+    | raw == "noreference" -> FieldAttrNoreference
+    | Just x <- T.stripPrefix "reference=" raw -> FieldAttrReference x
+    | Just x <- T.stripPrefix "constraint=" raw -> FieldAttrConstraint x
+    | Just x <- T.stripPrefix "default=" raw -> FieldAttrDefault x
+    | Just x <- T.stripPrefix "generated=" raw -> FieldAttrGenerated x
+    | Just x <- T.stripPrefix "sqltype=" raw -> FieldAttrSqltype x
+    | Just x <- T.stripPrefix "maxlen=" raw -> case reads (T.unpack x) of
+        [(n, s)] | all isSpace s -> FieldAttrMaxlen n
+        _ -> error $ "Could not parse maxlen field with value " <> show raw
+    | otherwise -> FieldAttrOther raw
```
parsonsmatt, line 200: Please use a `case` expression.

```haskell
fmap $ \case
    "Maybe" -> FieldAttrMaybe
    "nullable" -> FieldAttrNullable
    -- etc
    raw ->
        asum
            [ FieldAttrRefrence <$> T.stripPrefix "reference=" raw 
            , FieldAttrConstraint <$> T.stripPrefix "constraint=" raw
            -- etc...
            ]
```

persistent/Database/Persist/Types/Base.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE MultiWayIf #-}
```
parsonsmatt, line 1: I would honestly prefer to avoid this extension, since it is completely redundant with `case` expressions.

```suggestion
```

persistent-sqlite/test/main.hs
```diff
@@ -216,6 +217,7 @@ main = do
     TransactionLevelTest.specsWith db
     MigrationTest.specsWith db
     LongIdentifierTest.specsWith db
+    PersistLiteralTestSQL.specsWith db
```
parsonsmatt, line 220: I'm happy to accept the PR without fully supporting SQLite. We can mark this test as pending with `xdescribe`

```suggestion
    xdescribe "SQLite doesn't work, see issue #??? for discussion" $ do
        PersistLiteralTestSQL.specsWith db
```

parsonsmatt (approved):
I have a few style suggestions but I'm otherwise happy to accept the PR for inclusion with 2.11 release.

friedbrice:
The MySQL tests pass on my machine but fail on CI. I thought it might be a version issue, but it looks like CI is using the latest major version, just like I am.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -882,7 +907,7 @@ findAlters edef allDefs col@(Column name isNull type_ def _defConstraintName max
 -- | Prints the part of a @CREATE TABLE@ statement about a given
 -- column.
 showColumn :: Column -> String
-showColumn (Column n nu t def _defConstraintName maxLen ref) = concat
+showColumn (Column n nu t def _gen _defConstraintName maxLen ref) = concat
```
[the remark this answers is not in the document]
friedbrice: 0.0

Maybe a merge conflict X(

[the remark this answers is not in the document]
parsonsmatt: I suspect what's going on locally is that, your database was setup to have the GENERATED from a prior commit, and then this stuff ignores the `GENERATED` expression when making migrations, so it doesn't *un*-generate it. So your local code isn't migrating away from GENERATED, and it's already GENERATED, so it passes the test. But in CI we're making the migration from scratch, so it doesn't get GENERATED.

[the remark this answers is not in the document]
friedbrice: need to scrub the database after each test...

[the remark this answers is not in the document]
friedbrice: > and then this stuff ignores the GENERATED expression when making migrations, so it doesn't un-generate it.

That seems like a grave oversight on my part.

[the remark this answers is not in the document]
friedbrice: How do migrations currently handle a field that goes from having a default to not having a default?

friedbrice:
Okay, so. This patch is a lot different from what I originally described (https://github.com/yesodweb/persistent/pull/1122#issuecomment-689690823). The original didn't work because column generation turned out to be too backend-specific (e.g. still haven't gotten Sqlite working), and because the original plan broke things like search. I am happy with the present solution, though, as it feels closer to the spirit of the rest of Persistent.

persistent/Database/Persist/Types/Base.hs
friedbrice: Will add.

parsonsmatt:
I think we can support this in SQLite without too much fuss.

Here's the definition of `insert` for `SqlBackend`:

```haskell
    insert val = do
        conn <- ask
        let esql = connInsertSql conn t vals
        key <-
            case esql of
                ISRSingle sql -> withRawQuery sql vals $ do
                    x <- CL.head
                    case x of
                        Just [PersistInt64 i] -> case keyFromValues [PersistInt64 i] of
                            Left err -> error $ "SQL insert: keyFromValues: PersistInt64 " `mappend` show i `mappend` " " `mappend` unpack err
                            Right k -> return k
                        Nothing -> error $ "SQL insert did not return a result giving the generated ID"
                        Just vals' -> case keyFromValues vals' of
                            Left e -> error $ "Invalid result from a SQL insert, got: " ++ show vals' ++ ". Error was: " ++ unpack e
                            Right k -> return k

                ISRInsertGet sql1 sql2 -> do
                    rawExecute sql1 vals
                    withRawQuery sql2 [] $ do
                        mm <- CL.head
                        let m = maybe
                                  (Left $ "No results from ISRInsertGet: " `mappend` tshow (sql1, sql2))
                                  Right mm

                        -- TODO: figure out something better for MySQL
                        let convert x =
                                case x of
                                    [PersistByteString i] -> case readInteger i of -- mssql
                                                            Just (ret,"") -> [PersistInt64 $ fromIntegral ret]
                                                            _ -> x
                                    _ -> x
                            -- Yes, it's just <|>. Older bases don't have the
                            -- instance for Either.
                            onLeft Left{} x = x
                            onLeft x _ = x

                        case m >>= (\x -> keyFromValues x `onLeft` keyFromValues (convert x)) of
                            Right k -> return k
                            Left err -> throw $ "ISRInsertGet: keyFromValues failed: " `mappend` err
                ISRManyKeys sql fs -> do
                    rawExecute sql vals
                    case entityPrimary t of
                       Nothing -> error $ "ISRManyKeys is used when Primary is defined " ++ show sql
                       Just pdef ->
                            let pks = map fieldHaskell $ compositeFields pdef
                                keyvals = map snd $ filter (\(a, _) -> let ret=isJust (find (== a) pks) in ret) $ zip (map fieldHaskell $ entityFields t) fs
                            in  case keyFromValues keyvals of
                                    Right k -> return k
                                    Left e  -> error $ "ISRManyKeys: unexpected keyvals result: " `mappend` unpack e

        return key
      where
        tshow :: Show a => a -> Text
        tshow = T.pack . show
        throw = liftIO . throwIO . userError . T.unpack
        t = entityDef $ Just val
        vals = map toPersistValue $ toPersistFields val
```

Next up we have SQLite's `insertSql'` definition:

```haskell
insertSql' :: EntityDef -> [PersistValue] -> InsertSqlResult
insertSql' ent vals =
  case entityPrimary ent of
    Just _ ->
      ISRManyKeys sql vals
        where sql = T.concat
                [ "INSERT INTO "
                , escape $ entityDB ent
                , "("
                , T.intercalate "," $ map (escape . fieldDB) $ entityFields ent
                , ") VALUES("
                , T.intercalate "," (map (const "?") $ entityFields ent)
                , ")"
                ]
    Nothing ->
      ISRInsertGet ins sel
        where
          sel = T.concat
              [ "SELECT "
              , escape $ fieldDB (entityId ent)
              , " FROM "
              , escape $ entityDB ent
              , " WHERE _ROWID_=last_insert_rowid()"
              ]
          ins = T.concat
              [ "INSERT INTO "
              , escape $ entityDB ent
              , if null (entityFields ent)
                    then " VALUES(null)"
                    else T.concat
                      [ "("
                      , T.intercalate "," $ map (escape . fieldDB) $ entityFields ent
                      , ") VALUES("
                      , T.intercalate "," (map (const "?") $ entityFields ent)
                      , ")"
                      ]
              ]
```

The logic kinda ping-pongs here, so it can be easy to get lost.

1. We call `connInsertSql conn t (vals = map toPersistValue $ toPersistFields val)`. The `vals` argument is only used in the case that the entity has a `Primary` key definition. `insertSql'` checks for it and then stuffs it in the `ISRManyKeys` unchanged. 
2. In `insert`, we use `vals` as the value to pass to `rawExecute` and `rawQuery`. In the `ISRManyKeys` result, we execute the query and the convert the `vals` we inserted into a `Key entity` in a somewhat hacky way.

So we have three touch points:

1. Generating the list of fields to insert - `map (escape . fieldDB) $ entityFields ent`
2. Generating the list of `?` - `map (const "?") $ entityFields ent`
3. Generating the list of values to use for the placeholders: `map toPersistValue $ toPersistFields val`

1) and 2) are easy. `entityFields :: EntityDef -> [FieldDef]` and it would be trivial to `filter isGenerated . entityFields`. This fixes both of those spots.

3) is trickier. `toPersistFields` is a class method, and it *destroys* information - `PersistEntity e => e -> [SomePersistField]`. `SomePersistField` is a GADT:

```haskell
data SomePersistField = forall a. PersistField a => SomePersistField a
```

`SomePersistField` is only used in the TH code generation, so touching it won't be a big deal. But if we extend it to:

```haskell
data SomePersistField where
  SomePersistField :: (PersistEntity rec, PersistField a) => EntityField rec a -> a -> SomePersistField
```

That gives us:

```haskell
toInsertVals :: (PersistEntity rec) => rec -> [PersistValue]
toInsertVals = map toPersistValue . filter p . toPersistFields
  where
    p (SomePersistField efield val) = not (isGenerated (persistFieldDef efield))
```

Anyway I'm just thinking out loud, I'll write this out myself since it touches the code-gen :)

Thanks so much for all the great work!

friedbrice:
Nice explanation! I had written `valuesToInsert` essentially how you wrote `toInsertVals`, but the problem i was running up against was That by the time we get to `insert` we've already lost `rec`, as you keenly noticed.

Thanks :-)

friedbrice:
How did this compile?

persistent-zookeeper/Database/Persist/Zookeeper/Binary.hs

It has underscores left in it...

parsonsmatt:
Thanks for reminding me, I forgot about #875  :sweat_smile:

PR #1123, opened 2020-09-09
fix PersistDbSpecific escaping in persistent-postgresql

ldub:
Currently, the implementation of `PersistDbSpecific`:
- (in persistent-mysql) ***does not*** escape the given ByteString: https://github.com/yesodweb/persistent/blob/master/persistent-mysql/Database/Persist/MySQL.hs#L239
- (in persistent-sqlite) ***does not*** escape the given ByteString: https://github.com/yesodweb/persistent/blob/master/persistent-sqlite/Database/Sqlite.hs#L470
- (in persistent-postgresql) ***does*** escape the given ByteString (with single quotes): https://github.com/yesodweb/persistent/blob/master/persistent-postgresql/Database/Persist/Postgresql.hs#L553

This inconsistency led me to believe that the persistent-postgresql behavior is a bug. And indeed, looking at Persistent's own documentation:

https://github.com/yesodweb/persistent/blob/7f4abe94941c2ccea0f27d3e7d9036e9d100a79a/persistent/Database/Persist/Types/Base.hs#L381-L405

We can see that `toPoint` escapes the `POINT(44 44)` entry to become `'POINT(44 44)'`, and when inserting, persistent-postgresql will once again escape this to become `''POINT(44 44)''` which will lead to a runtime error. This further indicates that the escaping in persistent-postgresql is a bug.

I wrote a test to show this. It fails with the currently implementation of `instance PGTF.toField Unknown`:

```
GeographyTest
  works FAILED [1]

Failures:

  test/GeographyTest.hs:30:5: 
  1) GeographyTest works
       uncaught exception: SqlError
       SqlError {sqlState = "XX000", sqlExecStatus = FatalError, sqlErrorMsg = "parse error - invalid geometry", sqlErrorDetail = "", sqlErrorHint = "\"'P\" <-- parse error at position 2 within geometry"}

  To rerun use: --match "/GeographyTest/works/"

Randomized with seed 146528878

Finished in 0.0513 seconds
1 example, 1 failure
```

My change to stop escaping `PersistDbSpecific` fixes the test.

This is related to my investigation/implementation in #1122. I chose to make this a standalone PR to more clearly demonstrate the issue.

cc/ @MaxGabriel @friedbrice @parsonsmatt

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -52,6 +52,7 @@ import Data.Aeson.Types (modifyFailure)
 import qualified Data.Attoparsec.ByteString.Char8 as P
 import Data.Bits ((.&.))
 import Data.ByteString (ByteString)
+import qualified Data.ByteString.Builder as BB
 import qualified Data.ByteString.Char8 as B8
 import Data.Char (ord)
 import Data.Conduit
@@ -550,7 +551,7 @@ instance PGFF.FromField Unknown where
         Just dat -> return (Unknown dat)
 
 instance PGTF.ToField Unknown where
-    toField (Unknown a) = PGTF.Escape a
+    toField (Unknown a) = PGTF.Plain $ BB.byteString a
 
 type Getter a = PGFF.FieldParser a
 
--- a/persistent-postgresql/test/GeographyType.hs
+++ b/persistent-postgresql/test/GeographyType.hs
@@ -1,0 +1,24 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- | This type is taken directly from the example given here: https://hackage.haskell.org/package/persistent-2.10.5/docs/Database-Persist-Types.html#t:PersistValue
+module GeographyType where
+
+import PgInit
+import qualified Data.Text as T
+import qualified Data.ByteString as B
+import qualified Data.ByteString.Char8 as B8
+
+data Geo = Geo ByteString deriving (Show, Eq)
+
+instance PersistField Geo where
+  toPersistValue (Geo t) = PersistDbSpecific t
+
+  fromPersistValue (PersistDbSpecific t) = Right $ Geo $ B.concat ["'", t, "'"]
+  fromPersistValue _ = Left "Geo values must be converted from PersistDbSpecific"
+
+instance PersistFieldSql Geo where
+  sqlType _ = SqlOther "GEOGRAPHY(POINT,4326)"
+
+toPoint :: Double -> Double -> Geo
+toPoint lat lon = Geo $ B.concat ["'POINT(", ps $ lon, " ", ps $ lat, ")'"]
+  where ps = B8.pack . show
--- a/persistent-postgresql/persistent-postgresql.cabal
+++ b/persistent-postgresql/persistent-postgresql.cabal
@@ -52,6 +52,8 @@ test-suite test
                      JSONTest
                      CustomConstraintTest
                      PgIntervalTest
+                     GeographyType
+                     GeographyTest
     ghc-options:     -Wall
 
     build-depends:   base                 >= 4.9 && < 5
--- a/persistent-postgresql/test/GeographyTest.hs
+++ b/persistent-postgresql/test/GeographyTest.hs
@@ -1,0 +1,42 @@
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE StandaloneDeriving #-}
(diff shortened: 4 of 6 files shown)


ldub:
Note that while I believe fixing this bug is important and correct, it will cause issues for people who already use PersistDbSpecific with persistent-postgres, requiring them to manually add quotes around in existing toPersistValue implementations. 

I need advice on how we could proceed with fixing this.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -550,7 +551,7 @@ instance PGFF.FromField Unknown where
         Just dat -> return (Unknown dat)
 
 instance PGTF.ToField Unknown where
-    toField (Unknown a) = PGTF.Escape a
+    toField (Unknown a) = PGTF.Plain $ BB.byteString a
```
friedbrice, line 554: @ldub Can you link to the corresponding lines in `persist-mysql` and `persist-sqlite` for comparison?

friedbrice: e.g. whatever handles serializing `PersistDbSpecific` in those packages.

ldub: Yeah, they are linked at the top of my PR description, but here they are again:

Currently, the implementation of `PersistDbSpecific`:
- (in persistent-mysql) ***does not*** escape the given ByteString: https://github.com/yesodweb/persistent/blob/master/persistent-mysql/Database/Persist/MySQL.hs#L239
- (in persistent-sqlite) ***does not*** escape the given ByteString: https://github.com/yesodweb/persistent/blob/master/persistent-sqlite/Database/Sqlite.hs#L470
- (in persistent-postgresql) ***does*** escape the given ByteString (with single quotes): https://github.com/yesodweb/persistent/blob/master/persistent-postgresql/Database/Persist/Postgresql.hs#L553

ldub: Or did you mean linking to them directly in the code as a comment?

persistent-postgresql/test/GeographyType.hs
```diff
@@ -0,0 +1,24 @@
...
+
+-- | This type is taken directly from the example given here: https://hackage.haskell.org/package/persistent-2.10.5/docs/Database-Persist-Types.html#t:PersistValue
+module GeographyType where
+
+import PgInit
+import qualified Data.Text as T
+import qualified Data.ByteString as B
+import qualified Data.ByteString.Char8 as B8
+
+data Geo = Geo ByteString deriving (Show, Eq)
```
friedbrice, line 11: @ldub You said you concluded that `persist-postgres`'s handling of `PersistDbSpecific` was a bug because of differences in how other SQL backends handle `PersistDbSpecific` for this example. Can we move these test fixtures to a more-central location so that we can run this same test example against all the SQL backends.

ldub: We can't run this test against any other backend because PostGIS is a postgres-specific example.

However, I can port the `NullableGenerated` test from my other PR. Would that be good?

persistent-postgresql/Database/Persist/Postgresql.hs
friedbrice: No, I just jumped straight into the code changes w/o reading the PR description. Sorry for making extra work for you 😓

persistent-postgresql/test/GeographyType.hs
friedbrice: So, what examples, specifically, led you to believe `persist-postgres` was in error? It's something along those lines that we would need to automate.

ldub:
For context, here's the PR that introduced this issue: https://github.com/yesodweb/persistent/pull/160

MaxGabriel:
So, this seems potentially crazy dangerous to me, if we're saying that Persistent's Postgresql users will be opened up to SQL injection attacks from lack of escaping. I think the opposite direction, having MySQL and SQLite escape, is probably preferable. Any sort of SQL injection is a HUGE vulnerability that I think should come with bright red warning flags

ldub:
After discussion, @MaxGabriel @friedbrice and I have decided that the correct course of action is to improve #1122 instead of doing what I did in this PR. That will require a bit more work in the coming days/weeks. I still welcome any thoughts on this topic (figuring out how to support unescaped sql keywords, how to release a breaking change to escape mysql/sqlite PersistDbSpecific values, etc).

parsonsmatt:
@ldub We can make a PR that escapes `mysql` and `sqlite` `PersistDbSpecific` values and roll it up with 2.11 along with the literal constructor in #1122.

ldub:
@parsonsmatt I am worried that this would be a big breaking change for people currently using PersistDbSpecific with MySQL and SQLite. We are planning to implement PersistLiteral as is and have a separate PR for the MySQL and SQLite escaping changes. What do you think?

parsonsmatt:
I think it's important to get it right. Fixing it now in a major version bump is a good idea IMO.

2.11 is going to be a major version bump, and so it's Legal to throw that change in there. The changelog should say what's going on and point to the solution - `PersistDbLiteral` should restore the old behavior, if they need it.

Vlix:
This is now considered "done", because of #1122, right? (Even though #1161 is still an issue)

friedbrice:
Yeah, this is done. On `master` branch, new data created using `PersistLiteral` and `PersistLiteralEscaped` will round trip correctly.

parsonsmatt:
Closing this out since the work is handled in other PRs.

PR #1070, opened 2020-04-02
Exists query

limick:
This addresses #1069 by adding an `exists` function to the `PersistQueryRead` type class, which already provides the `count` function. The `count` function will be used for a fallback implementation for backends that don't provide a separate existence check.

Before submitting your PR, check that you've:

- [X] Bumped the version number (can we get this into 2.11?)
- [X] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [X] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

- [x] Update the Changelog.md file with a link to your PR
- [x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

Issue #1069, opened 2020-04-02
Add EXISTS query


limick:
Some backends allow queries to check if there is a record fulfilling certain criteria.

In SQL, for example, you can write `SELECT EXISTS(SELECT 1 FROM tbl WHERE criterion)`. The query will return `True` if and only if there is at least one row such that `criterion` is fulfilled.

The query is logically equivalent to doing `SELECT COUNT(1) FROM tbl WHERE criterion` and then checking if the result is non-zero, but the existence check can in some circumstances be much faster, especially on large amounts of data.

I'd like to add this type of queries. The type signature is the same as for the `count` function, except it returns `Bool`:
```
    -- | Check if there is at least one record fulfilling the given criterion.
    --
    -- @since 2.11
    exists :: (MonadIO m, PersistRecordBackend record backend)
           => [Filter record] -> ReaderT backend m Bool
```
For backends not providing optimized existence queries we can use a default implementation using `count`.

parsonsmatt:
Closed by #1131

PR #1131, opened 2020-10-02
Exists query

limick:
**Note**: Since I no longer have access to the previous repository, this replaces PR #1070 . I've applied the suggestions from that PR and I'm updating the changelog files to reflect the new PR number.

This addresses #1069 by adding an `exists` function to the `PersistQueryRead` type class, which already provides the `count` function. The `count` function will be used for a fallback implementation for backends that don't provide a separate existence check.

- [X] Bumped the version number (can we get this into 2.11?)

After submitting your PR:

### DIFF
--- a/persistent-mongoDB/ChangeLog.md
+++ b/persistent-mongoDB/ChangeLog.md
@@ -2,7 +2,7 @@
 
 ## 2.11.0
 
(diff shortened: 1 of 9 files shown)


parsonsmatt:
Wonderful, thanks!

PR #1139, opened 2020-10-25
Use explicit ordering in RawSqlTest

MaxGabriel:
I was non-deterministically getting an error on this test, where the pets were misordered. I'm having trouble reproducing, but I believe the issue is that Postgres' order is non-deterministic if not specified (in my experience, it can be Postgres version dependent). I believe the test as-is is incorrect without specifying the pet order, for Postgres.

[Postgres docs warn:](https://www.postgresql.org/docs/12/queries-order.html)

> After a query has produced an output table (after the select list has been processed) it can optionally be sorted. If sorting is not chosen, the rows will be returned in an unspecified order. The actual order in that case will depend on the scan and join plan types and the order on disk, but it must not be relied on. A particular output ordering can only be guaranteed if the sort step is explicitly chosen.

This commit orders the pets explicitly. This ensures when we get both of Mathias' pets, we get "Rodolfo" first and "Zeno" second, as the assertions require.

After submitting your PR:

### DIFF
--- a/persistent-test/ChangeLog.md
+++ b/persistent-test/ChangeLog.md
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 2.0.3.3
+
+* Fix RawSqlTest, which could fail non-deterministically for Postgres [#1139](https://github.com/yesodweb/persistent/pull/1139)
+
 ## 2.0.3.2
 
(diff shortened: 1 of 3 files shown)


parsonsmatt (approved):
awesome thanks!

MaxGabriel:
This is now live https://hackage.haskell.org/package/persistent-test-2.0.3.3/changelog

PR #1140, opened 2020-10-26
Add API for checking uniqueness constraints for updates

ivanbakel:
Entities which are already in the DB will conflict with themselves when using `checkUnique`, unless the value of each unique key is changed. This adds an alternative `checkUniqueUpdateable` which ignores conflicts from an entity with itself, so that uniqueness constraints can be checked without an update.

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Class/PersistUnique.hs
+++ b/persistent/Database/Persist/Class/PersistUnique.hs
@@ -17,6 +17,7 @@ module Database.Persist.Class.PersistUnique
   , insertUniqueEntity
   , replaceUnique
   , checkUnique
+  , checkUniqueUpdateable
   , onlyUnique
   , defaultUpsertBy
   , defaultPutMany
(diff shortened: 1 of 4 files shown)


persistent/Database/Persist/Class/PersistUnique.hs
```diff
@@ -571,6 +572,49 @@ checkUniqueKeys (x:xs) = do
...
+       , PersistRecordBackend record backend)
+    => Key record -> [Unique record] -> ReaderT backend m (Maybe (Unique record))
+checkUniqueKeysUpdateable _ [] = return Nothing
+checkUniqueKeysUpdateable key (x:xs) = do
+    y <- getBy x
+    case y of
+        Nothing -> checkUniqueKeysUpdateable key xs
+        Just (Entity k _)
+          | key == k -> checkUniqueKeysUpdateable key xs
+        Just _ ->  return (Just x)
```
parsonsmatt, line 617: This is fantastic, thanks!

persistent/Database/Persist/Class/PersistUnique.hs
```diff
@@ -571,6 +572,49 @@ checkUniqueKeys (x:xs) = do
...
+--
+-- We use <#schema-persist-unique-1 schema-1> and <#dataset-persist-unique-1 dataset-1> here.
+--
+-- This would be 'Nothing':
+--
+-- > mAlanConst <- checkUnique $ User "Alan" 70
+--
+-- While this would be 'Just' because SPJ already exists:
+--
+-- > mSpjConst <- checkUnique $ User "SPJ" 60
```
parsonsmatt, line 595: Awesome docs, thanks :smile:

parsonsmatt (approved):
Would you mind writing a quick test for this?

persistent/Database/Persist/Class/PersistUnique.hs
ivanbakel: I can't take the credit - these are largely copied from `checkUnique`, plus a paragraph on the difference in behaviour.

ivanbakel:
Alright, I've added a test for the basic behaviour - an entity not clashing with itself, but still clashing with other entities.

parsonsmatt:
Looks great, thanks so much!

PR #1145, opened 2020-10-30
Foreign key improvements

### LINKED ISSUE #957: persistent's quasi quoter wrongly rejects foreign keys that are not primary keys.

parsonsmatt:

After submitting your PR:

Continuation of #1121 

Fixes #957

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -44,7 +44,7 @@ import Data.Foldable
 import Control.Arrow
 import Control.Exception (Exception, throw, throwIO)
 import Control.Monad
-import Control.Monad.Trans.Except
+import Control.Monad.Except
 import Control.Monad.IO.Unlift (MonadIO (..), MonadUnliftIO)
 import Control.Monad.Logger (MonadLogger, runNoLoggingT)
 import Control.Monad.Trans.Reader (runReaderT)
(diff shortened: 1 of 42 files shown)


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -942,7 +943,8 @@ getColumn getter tableName' [PersistText columnName, PersistText isNullable, Per
                           , cDefault = fmap stripSuffixes d''
                           , cDefaultConstraintName = Nothing
                           , cMaxLen = Nothing
-                          , cReference = ref
+                          -- TODO: Fix cascade reference is ignored
```
parsonsmatt, line 946: looks like this is available [here](https://www.postgresql.org/docs/9.0/infoschema-referential-constraints.html)

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1066,14 +1068,15 @@ findAlters defs edef col@(Column name isNull sqltype def _defConstraintName _max
                                     a
                                     [name]
                                     (Util.dbIdColumnsEsc escape refdef)
+                                    -- TODO: Fix cascade reference is ignored
```
parsonsmatt, line 1071: seems like we can just copy it on down

persistent-template/Database/Persist/TH.hs
```diff
@@ -379,32 +406,39 @@ mkEntityDefSqlTypeExp emEntities entityMap ent =
...
-                        -- SqlTypeExp. However, there's one corner case: if there's
-                        -- a list of entity IDs, the datatype for the ID has not
-                        -- yet been created, so the compiler will fail. This extra
-                        -- clause works around this limitation.
-                        FTList _ -> SqlType' SqlString
-                        _ -> SqlTypeExp ftype
+            Right _ ->
+                SqlType' SqlString
+            Left (Just FTKeyCon) ->
+                SqlType' SqlString
```
parsonsmatt, line 412: this seems extremely dubious to me

parsonsmatt:
@erikd y'all wanna take a look at this?

PR #1152, opened 2020-11-04
support sqlite generated columns

parsonsmatt:

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Sql/Orphan/PersistStore.hs
+++ b/persistent/Database/Persist/Sql/Orphan/PersistStore.hs
@@ -46,22 +46,14 @@ import Database.Persist.Sql.Raw
 import Database.Persist.Sql.Types
 import Database.Persist.Sql.Util (
     dbIdColumns, keyAndEntityColumnNames, parseEntityValues, entityColumnNames
-  , updatePersistValue, mkUpdateText, commaSeparated)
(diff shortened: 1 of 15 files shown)


persistent/Database/Persist/Sql/Orphan/PersistStore.hs
```diff
@@ -41,27 +41,11 @@ import Web.HttpApiData (ToHttpApiData, FromHttpApiData)
...
 import Database.Persist.Sql.Types
 import Database.Persist.Sql.Util (
     dbIdColumns, keyAndEntityColumnNames, parseEntityValues, entityColumnNames
-  , updatePersistValue, mkUpdateText, commaSeparated)
-
--- Contributor note: any time you execute a prepared statement to
--- insert one or more rows, you need to use 'valuesToInsert entity' to
--- generate the list of values that you feed into your prepared
--- statement so that generated columns are properly scrubbed.
-valuesToInsert :: PersistEntity entity => entity -> [PersistValue]
```
friedbrice, line 55: I'm glad you changed this, because I just realized it introduces a bug in the Sqlite backend where adding a `generated=` attribute to a field (correctly) doesn't make the field generated in the schema but (incorrectly) redacts the value when binding to a prepared statement <.<;

PR #1162, opened 2020-11-22
Replace askLogFunc with askLoggerIO

mitchellvitez:
Just trying to do a small FIXME here

no haddock changes

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -49,7 +49,7 @@ import Control.Exception (Exception, throw, throwIO)
 import Control.Monad
 import Control.Monad.Except
 import Control.Monad.IO.Unlift (MonadIO (..), MonadUnliftIO)
-import Control.Monad.Logger (MonadLogger, runNoLoggingT)
+import Control.Monad.Logger (MonadLoggerIO, runNoLoggingT)
 import Control.Monad.Trans.Reader (runReaderT)
 import Control.Monad.Trans.Writer (WriterT(..), runWriterT)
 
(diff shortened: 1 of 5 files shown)


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -186,7 +186,7 @@ createPostgresqlPool = createPostgresqlPoolModified (const $ return ())
 --
 -- @since 2.1.3
 createPostgresqlPoolModified
-    :: (MonadUnliftIO m, MonadLogger m)
+    :: (MonadUnliftIO m, MonadLoggerIO m)
```
parsonsmatt, line 189: Changing the signature of these functions is a breaking change and would require a major version bump.

persistent/ChangeLog.md
```diff
@@ -2,6 +2,8 @@
 
 ## 2.11.0.1
 
+* [#1162](https://github.com/yesodweb/persistent/pull/1162)
+  * Replace `askLogFunc` with `askLoggerIO`
```
parsonsmatt, line 6: 2.11.0.1 has already been released on Hackage - since this is a breaking change, it'd need to go under 2.12 or 3.0.

parsonsmatt (commented):
Thanks for the PR! I think it'd be best if we made an issue for this and tagged it in for the next major version bump. It doesn't seem like a large enough change to release all on it's own right now.

mitchellvitez:
>best if we made an issue for this and tagged it in for the next major version bump. It doesn't seem like a large enough change to release all on it's own right now.

Agreed. To do that, do I just need to put this change under 2.12 in the changelog?

Also wasn't sure how to get CI building, looks like a github actions thing

parsonsmatt:
I fixed CI on  master, try merging that in.

I'm hesitant to release 2.12 with only this change, and I'd prefer to batch it up in the next breaking release.

PR #1163, opened 2020-11-23
Use resourcet-pool

brandonchinn178:

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Sql/Run.hs
+++ b/persistent/Database/Persist/Sql/Run.hs
@@ -14,6 +14,7 @@ import Data.Acquire (Acquire, ReleaseType(..), mkAcquireType, with)
 import Data.IORef (readIORef)
 import Data.Pool (Pool, LocalPool)
 import Data.Pool as P
+import Data.Pool.Acquire (poolToAcquire)
 import qualified Data.Map as Map
 import qualified Data.Text as T
 import System.Timeout (timeout)
@@ -34,15 +35,7 @@ unsafeAcquireSqlConnFromPool
     :: forall backend m
      . (MonadReader (Pool backend) m, BackendCompatible SqlBackend backend)
     => m (Acquire backend)
-unsafeAcquireSqlConnFromPool = do
-    pool <- MonadReader.ask
-
-    let freeConn :: (backend, LocalPool backend) -> ReleaseType -> IO ()
-        freeConn (res, localPool) relType = case relType of
-            ReleaseException -> P.destroyResource pool localPool res
-            _ -> P.putResource localPool res
-
-    return $ fst <$> mkAcquireType (P.takeResource pool) freeConn
+unsafeAcquireSqlConnFromPool = MonadReader.asks poolToAcquire
 
 
 -- | The returned 'Acquire' gets a connection from the pool, starts a new
--- a/stack.yaml
+++ b/stack.yaml
@@ -9,3 +9,6 @@ packages:
   - ./persistent-postgresql
   - ./persistent-redis
   - ./persistent-qq
+
+extra-deps:
+  - resourcet-pool-0.1.0.0
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for persistent
 
+## 2.11.0.2
+
+* Use `resourcet-pool` to break out some `Data.Pool` logic [#1163](https://github.com/yesodweb/persistent/pull/1163)
+
 ## 2.11.0.1
 
 * Docs/Bugs fixes [#1153](https://github.com/yesodweb/persistent/pull/1153)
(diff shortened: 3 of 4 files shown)


stack.yaml
```diff
@@ -9,3 +9,6 @@ packages:
   - ./persistent-postgresql
   - ./persistent-redis
   - ./persistent-qq
+
+extra-deps:
+  - resourcet-pool-0.1.0.0
```
parsonsmatt, line 14: We need this in `stackage` if we're going to depend on it and continuing offering `persistent` as a Stackage library.

persistent/Database/Persist/Sql/Run.hs
```diff
@@ -34,15 +35,7 @@ unsafeAcquireSqlConnFromPool
...
-unsafeAcquireSqlConnFromPool = do
-    pool <- MonadReader.ask
-
-    let freeConn :: (backend, LocalPool backend) -> ReleaseType -> IO ()
-        freeConn (res, localPool) relType = case relType of
-            ReleaseException -> P.destroyResource pool localPool res
-            _ -> P.putResource localPool res
-
-    return $ fst <$> mkAcquireType (P.takeResource pool) freeConn
+unsafeAcquireSqlConnFromPool = MonadReader.asks poolToAcquire
```
parsonsmatt, line 38: The implementation of [`poolToAcquire`](https://hackage.haskell.org/package/resourcet-pool-0.1.0.0/docs/src/Data.Pool.Acquire.html#poolToAcquire) appears to be copied from this function. This function has an `unsafe` prefix, but the function you're copying doesn't have any mention of unsafety in any way.

Looking at the docs, I think the "unsafety" might be coming from the fact that this `Acquire` doesn't start/commit a transaction. So it may be no real concern.

parsonsmatt (approved):
This MR isn't entirely compatible - `resourcet-pool` depends on newer version of `resource-pool` and `resourcet` than `persistent` does. I'm curious if `persistent` is wrong here, or if you can relax the bounds in `resourcet-pool`.

Overall I like it - factoring out functionality like this is generally a good thing.

stack.yaml
brandonchinn178: I just pushed it to Hackage. I'll put in on Stackage later today. I won't merge this PR until that's done

persistent/Database/Persist/Sql/Run.hs
brandonchinn178: Yeah, I mostly factored out this function. I agree, it seems like the unsafety is persistent-wise, not resource-pool-wise.

brandonchinn178:
> This MR isn't entirely compatible - `resourcet-pool` depends on newer version of `resource-pool` and `resourcet` than `persistent` does. I'm curious if `persistent` is wrong here, or if you can relax the bounds in `resourcet-pool`.

What do you mean? It seems like persistent has
```cabal
                   , resource-pool            >= 0.2.3
                   , resourcet                >= 1.1.10
```
whereas `resourcet-pool` has
```cabal
    , resource-pool >=0.2.1.0 && <0.3
    , resourcet >=1.1.2 && <2
```

parsonsmatt:
> > This MR isn't entirely compatible - `resourcet-pool` depends on newer version of `resource-pool` and `resourcet` than `persistent` does. I'm curious if `persistent` is wrong here, or if you can relax the bounds in `resourcet-pool`.
> 
> What do you mean? It seems like persistent has
> 
> ```cabal
>                    , resource-pool            >= 0.2.3
>                    , resourcet                >= 1.1.10
> ```
> 
> whereas `resourcet-pool` has
> 
> ```cabal
>     , resource-pool >=0.2.1.0 && <0.3
>     , resourcet >=1.1.2 && <2
> ```

Oh man. I think I was confusing which library I was looking at :joy: Good catch!

stack.yaml
brandonchinn178: @parsonsmatt Ok I'm adding it here: https://github.com/commercialhaskell/stackage/pull/5758

When it gets merged and the library is added to nightly, what should I change in this PR? If I just remove `extra-deps` from `stack.yaml` here, `stack build` will no longer work because `stack.yaml` uses `lts-14.27`, right?

parsonsmatt: We'll need to keep it in `extra-deps`, but any `nightly` checks shouldn't need it starting soon.

brandonchinn178:
@parsonsmatt This was added to nightly: https://github.com/commercialhaskell/stackage/pull/5758. Is there anything else blocking this PR?

parsonsmatt:
I fixed CI on  master, try merging that in

brandonchinn178:
@parsonsmatt seems to still be failing

parsonsmatt:
I don't think you merged `master` in to this branch - please give that a shot, it should fix CI

brandon-leapyear:
@parsonsmatt sorry about that! should be good to go now

parsonsmatt:
Yeah, looks good! Just needs the CHANGELOG and version bumps for a release. Should be a patch release since there's no change to the behavior of the code.

brandonchinn178:
Sounds good! How's that?

PR #1167, opened 2020-11-30
Fix references Migrations

kderme:
I noticed that when the cascade behavior of a field foreign key changes (ie by adding OnDeleteCascade), no migrations is generated. Is there any reason not to? There is a comment `An unspecified 'CascadeAction' is defaulted to 'Restrict' when doing migrations.`, so maybe there is already a way to make migrations work that I missed?

After submitting your PR:

### DIFF
--- a/persistent-mysql/ChangeLog.md
+++ b/persistent-mysql/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for persistent-mysql
 
+## 2.10.3.1
+* Fix foreign key migrations [#1167] https://github.com/yesodweb/persistent/pull/1167
+  * Fix a bug where a foreign key of a field to its table was ignored.
+
 ## 2.10.3
 
 * Compatibility with latest persistent
--- a/persistent-postgresql/ChangeLog.md
+++ b/persistent-postgresql/ChangeLog.md
@@ -1,5 +1,10 @@
 # Changelog for persistent-postgresql
 
+##  2.11.0.1
+* Fix foreign key migrations [#1167] https://github.com/yesodweb/persistent/pull/1167
+  * Fix a bug where a foreign key of a field to its table was ignored.
+  * Fix a bug where a altering details of a foreign key didn't trigger a migration
+
 ##  2.11.0.0
 
 * Foreign Key improvements [#1121] https://github.com/yesodweb/persistent/pull/1121
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -361,7 +361,7 @@ migrate' connectInfo allDefs getter val = do
                     let refTarget =
                           addReference allDefs refConstraintName refTblName cname (crFieldCascade cRef)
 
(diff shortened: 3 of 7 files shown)


parsonsmatt:
I fixed CI on  master, try merging that in

kderme:
Actually after thinking a bit more, this will trigger many unecessary migrations like
```
    EXECUTE 'ALTER TABLE "table" DROP CONSTRAINT "table_p_id_fkey"' ;
    EXECUTE 'ALTER TABLE "table" ADD CONSTRAINT "table_p_id_fkey" FOREIGN KEY("p_id") REFERENCES "parent"("id") ON DELETE RESTRICT  ON UPDATE RESTRICT' ;
```
which is just noise. We can make the equality more lenient, considering `Nothing == Just Restrict` to avoid it. Opinions?

kderme:
I rebased this and extended it with a fix for https://github.com/yesodweb/persistent/issues/1169

parsonsmatt:
> We can make the equality more lenient, considering `Nothing == Just Restrict` to avoid it. Opinions?

Yeah, i"m in favor of that! `Nothing` just means "unspecified", which does default to `Just Restrict`, so semantically that's fine.

kderme:
> > We can make the equality more lenient, considering `Nothing == Just Restrict` to avoid it. Opinions?
> 
> Yeah, i"m in favor of that! `Nothing` just means "unspecified", which does default to `Just Restrict`, so semantically that's fine.

Cool this is handled now at `equivalentRef`

parsonsmatt:
Fantastic, thanks! Just needs the CHangelog and a patch version bump :)

persistent-mysql/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for persistent-mysql
 
+## (Unreleased) 2.10.4
```
parsonsmatt, line 3:
```suggestion
## 2.10.4.1
```

parsonsmatt: It's a bugfix release, no dditions to the api, so a patch version is appropriate :)

persistent-postgresql/ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog for persistent-postgresql
 
+##  (Unreleased) 2.11.0.1
```
parsonsmatt, line 3:
```suggestion
## 2.11.0.1
```

I'll get this out asap.

parsonsmatt (approved):
a few version quibbles, thanks!

persistent-mysql/ChangeLog.md
kderme: should that be 2.10.3.1 or I'm wrong?

parsonsmatt: :joy: No, you're right - 2.10.3.1.

kderme: Cool this should be ready then :)

parsonsmatt (approved):

parsonsmatt:
Released and tagged

PR #1179, opened 2020-12-21
Add newtype, TH functions for deriving instances from backend compat

ivanbakel:
If a backend is compatible with another, it has a lot of obvious instances for various Persistent classes in terms of the backend it is compatible with - `RawSqlite` in the codebase is one example.

This adds a newtype, `Compatible`, which has these compatible instances declared on it. `Compatible` is designed to be used with `DerivingVia` to automatically yield these compatible instances when given a witness of backend compatibility.

To help use `Compatible` and write all the standalone derivations necessary to get a fully-featured compatible backend, this also adds some TH invocations to generate the standalone derivations for a given pair of compatible backend types.

`RawSqlite` is rewritten to dogfood on thes invocations, to prove their effectiveness and simplify the code somewhat.

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Compatible.hs
+++ b/persistent/Database/Persist/Compatible.hs
@@ -1,8 +1,8 @@
 module Database.Persist.Compatible
-  ( Compatible(..)
-  , makeCompatibleInstances
-  , makeCompatibleKeyInstances
-  ) where
+    ( Compatible(..)
+    , makeCompatibleInstances
+    , makeCompatibleKeyInstances
+    ) where
 
 import Database.Persist.Compatible.Types
 import Database.Persist.Compatible.TH
--- a/persistent/Database/Persist/Compatible.hs
+++ b/persistent/Database/Persist/Compatible.hs
@@ -1,0 +1,9 @@
+module Database.Persist.Compatible
+  ( Compatible(..)
+  , makeCompatibleInstances
+  , makeCompatibleKeyInstances
+  ) where
+
+import Database.Persist.Compatible.Types
+import Database.Persist.Compatible.TH
+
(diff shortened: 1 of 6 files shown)


persistent/Database/Persist/Compatible.hs
```diff
@@ -0,0 +1,9 @@
+module Database.Persist.Compatible
+  ( Compatible(..)
+  , makeCompatibleInstances
+  , makeCompatibleKeyInstances
+  ) where
```
parsonsmatt, line 5: Can we stick with 4 space indentation here? I know the codebase isn't consistent, but I want to get it moving in that direction. Thanks!

parsonsmatt (approved):
This is awesome! Just one style question.

I'd like to batch this up with 2.12, but I can also release as 2.12.1 if you can't get to it soon.

Thanks!

persistent/Database/Persist/Compatible.hs
ivanbakel: Sure! Is it just the first indent that's 4-space, and should I convert the other files to it as well?

parsonsmatt: Don't worry about converting anything old, but anything new should be 4 space. Thanks!

ivanbakel: Alright, I hope that's done correctly.

ivanbakel:
This build failure is interesting - the new code doesn't really make sense to have on old GHC versions, since if you don't have `DerivingVia`, you're not going to need the `newtype`. However, the `RawSqlite` code still needs its instances - so I could keep the current instances as a conditional compilation for if the GHC version is low enough.

ivanbakel:
Alright, sorry for the delay in fixing this, but it should now work across all the supported GHC versions. The instances get polyfilled for older GHCs which don't have `DerivingVia` and the new `Compatible` modules aren't exposed for old-enough GHC versions.

PR #1192, opened 2021-02-18
Fix XML conversion

vapaj:
Converting `sqltype=xml` values into Haskell types seem to fail:

```
*** Exception: Incompatible {errSQLType = "xml", errSQLTableOid = column number 1 is out of range 0..-1
Nothing, errSQLField = "column number 1 is out of range 0..-1
", errHaskellType = "Text", errMessage = "types incompatible"}
```

This is because Persistent tries to [parse it as `PersistText`](https://github.com/yesodweb/persistent/blob/f69716dcfeca01896ec42dec874393fbe60d3939/persistent-postgresql/Database/Persist/Postgresql.hs#L679) (using the [`Database.PostgreSQL.Simple.FromField` instance of `Text`](https://github.com/haskellari/postgresql-simple/blob/09a5bab9d4aa83ce38f2f35845fe742838d50f4c/src/Database/PostgreSQL/Simple/FromField.hs#L419-L422)), however the oid of `xml`, 142, does not match the oid's of `name`, `text`, `char`, `bpchar` or `varchar`, resulting in an `Incompatible` error.

To fix this, I simply changed the type from `Text` to (unknown) `ByteString`.

After submitting your PR:


persistent-postgresql/ChangeLog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog for persistent-postgresql
 
+##  2.12.0.1
+* Fix XML conversion [#1192](https://github.com/yesodweb/persistent/pull/1192)
```
parsonsmatt, line 4: 2.12 is not yet released - can you put this back to 2.12?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -705,7 +705,7 @@ builtinGetters = I.fromList
     , (1005,             listOf PersistInt64)
     , (1007,             listOf PersistInt64)
     , (1009,             listOf PersistText)
-    , (143,              listOf PersistText)
+    , (143,              listOf (PersistByteString . unUnknown))
```
parsonsmatt, line 708: Interesting.

So, based on what you're saying - the `xml` type shouldn't *ever* have worked, right? So no existing working code should be affected by this change?

vapaj: That's right.

parsonsmatt (approved):

Issue #1061, opened 2020-03-30
Refactor `[]` to `NonEmpty` where appropriate


parsonsmatt:
There are a number of places where we use lists instead of non-empty lists. We should refactor to use `NonEmpty` instead of `[]` here to make the implementations safer.

duobei:
Could someone assign the issue to me?

parsonsmatt:
You got it! Have fun!

PR #1193, opened 2021-02-24
Refactor [] to NonEmpty in Quasi module

danbroooks:
A refactoring contribution towards https://github.com/yesodweb/persistent/issues/1061

### DIFF
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -650,16 +650,9 @@ removeSpaces =
 
 -- | Divide lines into blocks and make entity definitions.
 parseLines :: PersistSettings -> NonEmpty Line -> [EntityDef]
-parseLines ps lines =
-    fixForeignKeysAll $ toEnts lines
+parseLines ps =
+    fixForeignKeysAll . map mk . associateLines . skipEmpty
   where
-    toEnts :: NonEmpty Line -> [UnboundEntityDef]
-    toEnts =
-        map mk
-        . associateLines
-        . skipEmpty
-        . NEL.toList
-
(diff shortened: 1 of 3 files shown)


parsonsmatt:
I've fixed CI, mind rebasing and trying again?

persistent/Database/Persist/Quasi.hs
```diff
@@ -538,11 +538,12 @@ lowerCaseSettings = defaultPersistSettings
 
 -- | Parses a quasi-quoted syntax into a list of entity definitions.
 parse :: PersistSettings -> Text -> [EntityDef]
-parse ps = parseLines ps . preparse
+parse ps = maybe [] (parseLines ps) . preparse
 
-preparse :: Text -> [Line]
+preparse :: Text -> Maybe (NonEmpty Line)
```
parsonsmatt, line 543: `Maybe (NonEmpty a)` is equivalent to `[a]`. What's the purpose of this change?

parsonsmatt: I think I'm just not sure about the benefit of *this* particular `[] -> NonEmpty` change. We don't appear to have any behavior that assumes or depends on `[]` being non-empty somewhere with this set of code.

But, it's also not a breaking change, and it seems like it makes sense semantically in the parsing section here.

parsonsmatt (approved):
Thanks! 😄

parsonsmatt:
Since this is a refactoring-only change, isn't exposed to the public interface, and is covered entirely by the tests, I'm going to merge without requiring a changelong entry.

If you want to put a changelog entry in for the sweet sweet credit, feel free to push it up. I'll merge this sometime in the next few days.

Thanks!

PR #1195, opened 2021-03-02
Increase upper bound on `mysql`

paul-rouse:
`mysql-0.2` is now released

Before submitting your PR, check that you've:

- [ ] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [ ] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

- [ ] Update the Changelog.md file with a link to your PR
- [ ] Bumped the version number if there isn't an `(unreleased)` on the Changelog
- [ ] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

PR #1196, opened 2021-03-04
Relaxes mysql upper bound

jkachmar:
Relates to https://github.com/commercialhaskell/stackage/issues/5891.

I haven't tested this out locally, but the changelog for `mysql-0.2` makes it seem like the only breaking change introduced doesn't affect `persistent-mysql` at first glance (cf. https://hackage.haskell.org/package/mysql-0.2/changelog).

Hopefully CI is enough of a sanity check, but let me know if there's anything I should try out locally that would help get this merged.

### DIFF
--- a/persistent-mysql/persistent-mysql.cabal
+++ b/persistent-mysql/persistent-mysql.cabal
@@ -35,7 +35,7 @@ library
                    , conduit          >= 1.2.12
                    , containers       >= 0.5
                    , monad-logger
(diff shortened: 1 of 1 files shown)


paul-rouse:
This is actually a duplicate of #1195 which I submitted when I uploaded the new version of `mysql`.  I have successfully run the persistent tests locally with `mysql-0.2`.  The failures in CI tests seem to be unrelated.

Issue #1161, opened 2020-11-20
Deserialization of data formerly using `PersistDbSpecific` and now using `PersistLiteralEscaped`


friedbrice:
Persistent 2.11 deprecated `PersistDbSpecific` due to inconsistent handling of escaping among SQL backends (which went so far as to even cause the examples provided in the documentation to fail), replacing it with `PersistLiteral` and `PersistLiteralEscaped`. Users are directed to replace occurrences of `PersistDbSpecific` with a constructor appropriate for their intended purpose. The advice is misleading, however, because doing so will cause data created with prior versions of persistent to fail to deserialize in some cases.

I have constructed [an example](https://github.com/friedbrice/persist_ser). While it's a bit complicated, it does demonstrate data that successfully makes a round trip in Persistent 2.10 but that fails to deserialize once Persistent has been upgraded to 2.11. My attempts to find a simpler example have so far failed. Users can sidestep the problem if they are willing to keep `PersistDbSpecific` in their code and put up with the deprecation warnings, but this confounds our attempts to migrate users away from it. The alternative is to ask users to do a costly and manual database migration, which is less than ideal for its own reasons.

The de/serialization in question is defined by the `FromJSON`/`ToJSON` instances for `PersistValue` in [Database.Persist.Types.Base](https://github.com/yesodweb/persistent/blob/master/persistent/Database/Persist/Types/Base.hs). I believe that we can provide a backwards-compatible format, allowing users to migrate away from `PersistDbSpecific` without needing a costly migration, by liberalizing the deserializer so that `p`-prefixed byte arrays are accepted as valid `PersistLiteral` and `PersistLiteralEscaped` values, though the details are necessarily backend specific, as Postgres will need to deserialize such values to `PersistLiteralEscaped` while MySQL and Sqlite will need to deserialize such values to `PersistLiteral`. If this sounds like an acceptable approach to the maintainers, I can begin work on a patch this weekend.

friedbrice:
^ @MaxGabriel and @parsonsmatt

parsonsmatt:
Sounds good to me! Thanks for finding and proposing a fix for this.

parsonsmatt:
[Potential fix for postgres, but potential break for mysql and sqlite](https://github.com/friedbrice/persistent/commit/f2fb5475d26487c0ec017004149fcd11a7e742cc).

As mentioned, the problem is due to `ToJSON` and `FromJSON` instances for `PersistValue`. These are 

[...]

PR #1205, opened 2021-03-15
PersistLiteral pattern synonym

parsonsmatt:

After submitting your PR:

--- 

This PR provides a fix for #1161. Briefly, we replace the various `PeristLiteral` constructors with pattern synonyms that back the `PersistLiteral_` constructor, which accepts a `LiteralKind` value. The pattern synonyms can be used to construct a value as expected, but in pattern matching, it actually ignores the `LiteralKind` value.

This approach has few downsides, documented in the #1161 discussion. To summarize, a pattern match on `PersistDbSpecific bs` expands into a match on `PersistLiteral_ _kind bs`, which means that it will *always* succeed. Patterns that attempted to disambiguate and provide difference behavior based on `PersistDbSpecific` vs `PersistLiteral` will always hit the first case that matches instead of differentiating. Specifically, if it hits a `PersistDbSpecific bs` and delegates to `PersistLiteral bs`, then it will loop.


parsonsmatt (commented):
This should occur in two steps.

First, I want to get a patch release out ASAP for 2.11 that modifies the deprecation message. Users should know to continue pattern matching on `PersistDbSpecific` even if they've added a case for `PersistLiteral`. I'll write an issue with the relevant information, since it's a pretty specific code path it has to go down. Users that upgrade, see the patch, and include both patterns will work fine.

Second, we can roll this change out. Since it's a breaking change, it'll go out with 2.12. But users that upgrade directly to 2.12 *should not* observe any problematic behavior.

> Why even do this change?

Uhhhh that's a good question. We have deprecated the `PersistDbSpecific` constructor, so folks want to get rid of it to clean up warnings. But we're now requiring that you keep it around for backwards compatibility on the data loading side! THat's unfortunate.

So this PR can allow you to have a backwards-compatible deserialization logic while avoiding any deprecated terms.

parsonsmatt:
I've tried this out on the Mercury codebase, and we were able to load all of our models successfully with this branch. I'm going to move forward with the plan.

PR #1209, opened 2021-03-17
Backport the 1207 fix to `persistent-2.11.0`

parsonsmatt:

After submitting your PR:

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -10,6 +10,19 @@ on:
 jobs:
   build:
     services:
+      postgres:
+        image: postgres:12-alpine
+        env:
+          POSTGRES_USER: perstest
+          POSTGRES_PASSWORD: perstest
+          POSTGRES_DB: persistent
+        ports:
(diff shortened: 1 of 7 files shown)


codygman:
Alright using:

```
  - git: https://github.com/yesodweb/persistent
    commit: f496bbdd74f8b26be1d09dd0bfcc697bd7d8d48b
    subdirs:
     - persistent
     - persistent-mongoDB
     - persistent-postgresql
     - persistent-template
```

It passed 10 times in ghcid, though I'm suspicious of this type of problem there. Before I tried that I pushed to our CI and it's also passing there.

I'll keep running it in CI which starts from a clean slate and see what happens.

parsonsmatt:
The problem occurred when running against the single threaded runtime as well as the threaded runtime, and only was alleviated by having more and more threads available (eg `-N2` had more errors than `-N4` etc). Not sure if `ghcid` would exacerbate or ameliorate the issue. 

Here's hoping for those good CI passes 😄

PR #1222, opened 2021-03-31
Mer 676

dmarticus:

-  [x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)

After submitting your PR:

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -1428,7 +1428,7 @@ copyField = CopyField
 --   []
 -- @
 --
--- Once we run that code on the datahase, the new data set looks like this:
+-- Once we run that code on the database, the new data set looks like this:
 --
 -- > items:
 -- > +------+-------------+-------+----------+
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -1,5 +1,9 @@
 {-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -23,6 +27,7 @@ module Database.Persist.Postgresql
     , ConnectionString
     , PostgresConf (..)
     , PgInterval (..)
+    , upsertWhere
     , openSimpleConn
     , openSimpleConnWithVersion
     , tableName
@@ -50,7 +55,7 @@ import Control.Monad
 import Control.Monad.Except
 import Control.Monad.IO.Unlift (MonadIO (..), MonadUnliftIO)
 import Control.Monad.Logger (MonadLoggerIO, runNoLoggingT)
-import Control.Monad.Trans.Reader (runReaderT)
+import Control.Monad.Trans.Reader (ReaderT(..), runReaderT)
 import Control.Monad.Trans.Writer (WriterT(..), runWriterT)
 
 import qualified Blaze.ByteString.Builder.Char8 as BBB
@@ -73,7 +78,7 @@ import Data.Function (on)
 import Data.Int (Int64)
 import qualified Data.IntMap as I
 import Data.IORef
-import Data.List (find, sort, groupBy, foldl')
+import Data.List (find, sort, groupBy, foldl', transpose, inits)
 import Data.List.NonEmpty (NonEmpty)
 import qualified Data.List as List
 import qualified Data.List.NonEmpty as NEL
@@ -82,7 +87,7 @@ import Data.Maybe
 import Data.Monoid ((<>))
 import Data.Pool (Pool)
 import Data.String.Conversions.Monomorphic (toStrictByteString)
-import Data.Text (Text)
+import Data.Text (Text, pack)
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Data.Text.IO as T
@@ -91,6 +96,7 @@ import Data.Time (utc, NominalDiffTime, localTimeToUTC)
 import System.Environment (getEnvironment)
 
 import Database.Persist.Sql
+import Database.Persist.Sql.Util (isIdField)
 import qualified Database.Persist.Sql.Util as Util
 
 -- | A @libpq@ connection string.  A simple example of connection
@@ -396,7 +402,6 @@ insertSql' ent vals =
                 , ")"
                 ]
         ]
-
 
 upsertSql' :: EntityDef -> NonEmpty (FieldNameHS, FieldNameDB) -> Text -> Text
 upsertSql' ent uniqs updateVal =
@@ -1487,14 +1492,23 @@ tableName = escapeE . tableDBName
 fieldName :: (PersistEntity record) => EntityField record typ -> Text
 fieldName = escapeF . fieldDBName
 
+fieldName' ::  forall record typ. (PersistEntity record, PersistEntityBackend record ~ SqlBackend) => EntityField record typ -> FieldNameDB
+fieldName' f = fieldDB $ persistFieldDef f
+
 escapeC :: ConstraintNameDB -> Text
 escapeC = escapeWith escape
 
 escapeE :: EntityNameDB -> Text
 escapeE = escapeWith escape
 
+escapeES :: EntityNameDB -> String
+escapeES = escapeWith (escapeDBName . T.unpack)
+
 escapeF :: FieldNameDB -> Text
 escapeF = escapeWith escape
+
+escapeFS :: FieldNameDB -> String
+escapeFS = escapeWith (escapeDBName . T.unpack)
 
 escape :: Text -> Text
 escape s =
@@ -1503,6 +1517,14 @@ escape s =
     go "" = ""
     go ('"':xs) = "\"\"" ++ go xs
     go (x:xs) = x : go xs
+
+-- | Escape a database name to be included on a query.
+escapeDBName :: String -> String
+escapeDBName str = '`' : go str
+    where
+      go ('`':xs) = '`' : '`' : go xs
+      go ( x :xs) =     x     : go xs
+      go ""       = "`"
 
 -- | Information required to connect to a PostgreSQL database
 -- using @persistent@'s generic facilities.  These values are the
@@ -1737,6 +1759,292 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
   where
     fields = keyAndEntityFields ent
     conflictColumns = escapeF . fieldDB <$> entityKeyFields ent
+
+-- | This type is used to determine how to update rows using MySQL's
+-- @INSERT ... ON DUPLICATE KEY UPDATE@ functionality, exposed via
+-- 'insertManyOnDuplicateKeyUpdate' in this library.
+--
+-- @since 2.8.0
+data HandleUpdateCollision record where
+  -- | Copy the field directly from the record.
+  CopyField :: EntityField record typ -> HandleUpdateCollision record
+  -- | Only copy the field if it is not equal to the provided value.
+  CopyUnlessEq :: PersistField typ => EntityField record typ -> typ -> HandleUpdateCollision record
+
+-- | Postgres specific 'upsertWhere'. This will prevent multiple queries, when one will
+-- do. The record will be inserted into the database. In the event that the
+-- record already exists in the database, the record will have the
+-- relevant updates performed.
+upsertWhere
+  :: ( backend ~ PersistEntityBackend record
+     , PersistEntity record
+     , PersistEntityBackend record ~ SqlBackend
+     , MonadIO m
+     , PersistStore backend
+     , BackendCompatible SqlBackend backend
+     )
+  => record
+  -> SqlBackend
+  -> [Update record]
+  -> [Filter record]
+  -> ReaderT backend m ()
+upsertWhere record conn updates filts =
+  upsertManyWhere [record] conn [] updates filts
+
+upsertManyWhere ::
+  forall record backend m.
+  ( backend ~ PersistEntityBackend record,
+    BackendCompatible SqlBackend backend,
+    PersistEntityBackend record ~ SqlBackend,
+    PersistEntity record,
+    MonadIO m
+  ) =>
+  -- | A list of the records you want to insert, or update
+  [record] ->
+  SqlBackend ->
+  -- | A list of the fields you want to copy over.
+  [HandleUpdateCollision record] ->
+  -- | A list of the updates to apply that aren't dependent on the record being inserted.
+  [Update record] ->
+  [Filter record] ->
+  ReaderT backend m ()
+upsertManyWhere [] _ _ _ _ = return ()
+upsertManyWhere records conn fieldValues updates conditions =
+  uncurry rawExecute $
+    mkBulkUpsertQuery records conn fieldValues updates conditions
+
+-- | This creates the query for 'bulkInsertOnDuplicateKeyUpdate'. If you
+-- provide an empty list of updates to perform, then it will generate
+-- a dummy/no-op update using the first field of the record. This avoids
+-- duplicate key exceptions.
+mkBulkUpsertQuery
+    :: (PersistEntity record, PersistEntityBackend record ~ SqlBackend)
+    => [record] -- ^ A list of the records you want to insert, or update
+    -> SqlBackend
+    -> [HandleUpdateCollision record] -- ^ A list of the fields you want to copy over.
+    -> [Update record] -- ^ A list of the updates to apply that aren't dependent on the record being inserted.
+    -> [Filter record]
+    -> (Text, [PersistValue])
+mkBulkUpsertQuery records conn fieldValues updates filts =
+    (q, recordValues <> updsValues <> copyUnlessValues)
+  where
+    mfieldDef x = case x of
+        CopyField rec -> Right (fieldDbToText (persistFieldDef rec))
+        CopyUnlessEq rec val -> Left (fieldDbToText (persistFieldDef rec), toPersistValue val)
+    (fieldsToMaybeCopy, updateFieldNames) = partitionEithers $ map mfieldDef fieldValues
+    fieldDbToText = T.pack . escapeFS . fieldDB
+    entityDef' = entityDef records
+    firstField = case entityFieldNames of
+        [] -> error "The entity you're trying to insert does not have any fields."
+        (field:_) -> field
+    entityFieldNames = map fieldDbToText (entityFields entityDef')
+    nameOfTable = T.pack . escapeES . entityDB $ entityDef'
+    copyUnlessValues = map snd fieldsToMaybeCopy
+    recordValues = concatMap (map toPersistValue . toPersistFields) records
+    recordPlaceholders = Util.commaSeparated $ map (Util.parenWrapped . Util.commaSeparated . map (const "?") . toPersistFields) records
+    mkCondFieldSet n _ = T.concat
+        [ n
+        , "=COALESCE("
+        ,   "NULLIF("
+        ,     "VALUES(", n, "),"
+        ,     "?"
+        ,   "),"
+        ,   n
+        , ")"
+        ]
+    condFieldSets = map (uncurry mkCondFieldSet) fieldsToMaybeCopy
+    fieldSets = map (\n -> T.concat [n, "=VALUES(", n, ")"]) updateFieldNames
+    upds = map (Util.mkUpdateText' (pack . escapeFS) id) updates
+    updsValues = map (\(Update _ val _) -> toPersistValue val) updates
+    wher = if null filts then "" else filterClause False conn filts
+    updateText = case fieldSets <> upds <> condFieldSets of
+        [] -> T.concat [firstField, "=", firstField]
+        xs -> Util.commaSeparated xs
+    q = T.concat
+        [ "INSERT INTO "
+        , nameOfTable
+        , " ("
+        , Util.commaSeparated entityFieldNames
+        , ") "
+        , " VALUES "
+        , recordPlaceholders
+        , " ON CONFLICT DO UPDATE SET "
+        , updateText
+        , " WHERE "
+        , wher
+        ]
+
+filterClause :: (PersistEntity val, PersistEntityBackend val ~ SqlBackend)
+             => Bool -- ^ include table name?
+             -> SqlBackend
+             -> [Filter val]
+             -> Text
+filterClause b c = fst . filterClauseHelper b True c OrNullNo
+
+filterClauseHelper :: (PersistEntity val, PersistEntityBackend val ~ SqlBackend)
+             => Bool -- ^ include table name?
+             -> Bool -- ^ include WHERE?
+             -> SqlBackend
+             -> OrNull
+             -> [Filter val]
+             -> (Text, [PersistValue])
+filterClauseHelper includeTable includeWhere conn orNull filters =
+    (if not (T.null sql) && includeWhere
+        then " WHERE " <> sql
+        else sql, vals)
+  where
+    (sql, vals) = combineAND filters
+    combineAND = combine " AND "
+
+    combine s fs =
+        (T.intercalate s $ map wrapP a, mconcat b)
+      where
+        (a, b) = unzip $ map go fs
+        wrapP x = T.concat ["(", x, ")"]
+
+    go (BackendFilter _) = error "BackendFilter not expected"
+    go (FilterAnd []) = ("1=1", [])
+    go (FilterAnd fs) = combineAND fs
+    go (FilterOr []) = ("1=0", [])
+    go (FilterOr fs)  = combine " OR " fs
+    go (Filter field value pfilter) =
+        let t = entityDef $ dummyFromFilts [Filter field value pfilter]
+        in case (isIdField field, entityPrimary t, allVals) of
+                 (True, Just pdef, PersistList ys:_) ->
+                    if length (compositeFields pdef) /= length ys
+                       then error $ "wrong number of entries in compositeFields vs PersistList allVals=" ++ show allVals
+                    else
+                      case (allVals, pfilter, isCompFilter pfilter) of
+                        ([PersistList xs], Eq, _) ->
+                           let sqlcl=T.intercalate " and " (map (\a -> connEscapeFieldName conn (fieldDB a) <> showSqlFilter pfilter <> "? ")  (compositeFields pdef))
+                           in (wrapSql sqlcl,xs)
+                        ([PersistList xs], Ne, _) ->
+                           let sqlcl=T.intercalate " or " (map (\a -> connEscapeFieldName conn (fieldDB a) <> showSqlFilter pfilter <> "? ")  (compositeFields pdef))
+                           in (wrapSql sqlcl,xs)
+                        (_, In, _) ->
+                           let xxs = transpose (map fromPersistList allVals)
+                               sqls=map (\(a,xs) -> connEscapeFieldName conn (fieldDB a) <> showSqlFilter pfilter <> "(" <> T.intercalate "," (replicate (length xs) " ?") <> ") ") (zip (compositeFields pdef) xxs)
+                           in (wrapSql (T.intercalate " and " (map wrapSql sqls)), concat xxs)
+                        (_, NotIn, _) ->
+                           let xxs = transpose (map fromPersistList allVals)
+                               sqls=map (\(a,xs) -> connEscapeFieldName conn (fieldDB a) <> showSqlFilter pfilter <> "(" <> T.intercalate "," (replicate (length xs) " ?") <> ") ") (zip (compositeFields pdef) xxs)
+                           in (wrapSql (T.intercalate " or " (map wrapSql sqls)), concat xxs)
+                        ([PersistList xs], _, True) ->
+                           let zs = tail (inits (compositeFields pdef))
+                               sql1 = map (\b -> wrapSql (T.intercalate " and " (map (\(i,a) -> sql2 (i==length b) a) (zip [1..] b)))) zs
+                               sql2 islast a = connEscapeFieldName conn (fieldDB a) <> (if islast then showSqlFilter pfilter else showSqlFilter Eq) <> "? "
+                               sqlcl = T.intercalate " or " sql1
+                           in (wrapSql sqlcl, concat (tail (inits xs)))
+                        (_, BackendSpecificFilter _, _) -> error "unhandled type BackendSpecificFilter for composite/non id primary keys"
+                        _ -> error $ "unhandled type/filter for composite/non id primary keys pfilter=" ++ show pfilter ++ " persistList="++show allVals
+                 (True, Just pdef, []) ->
+                     error $ "empty list given as filter value filter=" ++ show pfilter ++ " persistList=" ++ show allVals ++ " pdef=" ++ show pdef
+                 (True, Just pdef, _) ->
+                     error $ "unhandled error for composite/non id primary keys filter=" ++ show pfilter ++ " persistList=" ++ show allVals ++ " pdef=" ++ show pdef
+
+                 _ ->   case (isNull, pfilter, length notNullVals) of
+                            (True, Eq, _) -> (name <> " IS NULL", [])
+                            (True, Ne, _) -> (name <> " IS NOT NULL", [])
+                            (False, Ne, _) -> (T.concat
+                                [ "("
+                                , name
+                                , " IS NULL OR "
+                                , name
+                                , " <> "
+                                , qmarks
+                                , ")"
+                                ], notNullVals)
+                            -- We use 1=2 (and below 1=1) to avoid using TRUE and FALSE, since
+                            -- not all databases support those words directly.
+                            (_, In, 0) -> ("1=2" <> orNullSuffix, [])
+                            (False, In, _) -> (name <> " IN " <> qmarks <> orNullSuffix, allVals)
+                            (True, In, _) -> (T.concat
+                                [ "("
+                                , name
+                                , " IS NULL OR "
+                                , name
+                                , " IN "
+                                , qmarks
+                                , ")"
+                                ], notNullVals)
+                            (False, NotIn, 0) -> ("1=1", [])
+                            (True, NotIn, 0) -> (name <> " IS NOT NULL", [])
+                            (False, NotIn, _) -> (T.concat
+                                [ "("
+                                , name
+                                , " IS NULL OR "
+                                , name
+                                , " NOT IN "
+                                , qmarks
+                                , ")"
+                                ], notNullVals)
+                            (True, NotIn, _) -> (T.concat
+                                [ "("
+                                , name
+                                , " IS NOT NULL AND "
+                                , name
+                                , " NOT IN "
+                                , qmarks
+                                , ")"
+                                ], notNullVals)
+                            _ -> (name <> showSqlFilter pfilter <> "?" <> orNullSuffix, allVals)
+
+      where
+        isCompFilter Lt = True
+        isCompFilter Le = True
+        isCompFilter Gt = True
+        isCompFilter Ge = True
+        isCompFilter _ =  False
+
+        wrapSql sqlcl = "(" <> sqlcl <> ")"
+        fromPersistList (PersistList xs) = xs
+        fromPersistList other = error $ "expected PersistList but found " ++ show other
+
+        filterValueToPersistValues :: forall a.  PersistField a => FilterValue a -> [PersistValue]
+        filterValueToPersistValues = \case
+            FilterValue a -> [toPersistValue a]
+            FilterValues as -> toPersistValue <$> as
+            UnsafeValue x -> [toPersistValue x]
+
+        orNullSuffix =
+            case orNull of
+                OrNullYes -> mconcat [" OR ", name, " IS NULL"]
+                OrNullNo -> ""
+
+        isNull = PersistNull `elem` allVals
+        notNullVals = filter (/= PersistNull) allVals
+        allVals = filterValueToPersistValues value
+        tn = connEscapeTableName conn $ entityDef $ dummyFromFilts [Filter field value pfilter]
+        name =
+            (if includeTable
+                then ((tn <> ".") <>)
+                else id)
+            $ connEscapeFieldName conn (fieldName' field)
+        qmarks = case value of
+                    FilterValue{} -> "(?)"
+                    UnsafeValue{} -> "(?)"
+                    FilterValues xs ->
+                        let parens a = "(" <> a <> ")"
+                            commas = T.intercalate ","
+                            toQs = fmap $ const "?"
+                            nonNulls = filter (/= PersistNull) $ map toPersistValue xs
+                         in parens . commas . toQs $ nonNulls
+        showSqlFilter Eq = "="
+        showSqlFilter Ne = "<>"
+        showSqlFilter Gt = ">"
+        showSqlFilter Lt = "<"
+        showSqlFilter Ge = ">="
+        showSqlFilter Le = "<="
+        showSqlFilter In = " IN "
+        showSqlFilter NotIn = " NOT IN "
+        showSqlFilter (BackendSpecificFilter s) = s
+
+data OrNull = OrNullYes | OrNullNo
+
+dummyFromFilts :: [Filter v] -> Maybe v
+dummyFromFilts _ = Nothing
+
+
 
 putManySql' :: [Text] -> [FieldDef] -> EntityDef -> Int -> Text
 putManySql' conflictColumns (filter isFieldNotGenerated -> fields) ent n = q
--- a/persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
+++ b/persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
@@ -7,7 +7,6 @@ module Database.Persist.Sql.Orphan.PersistQuery
     ( deleteWhereCount
     , updateWhereCount
     , decorateSQLWithLimitOffset
-    , filterClauseHelper
     ) where
 
 import Control.Exception (throwIO)
--- a/persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
+++ b/persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
@@ -7,6 +7,7 @@ module Database.Persist.Sql.Orphan.PersistQuery
     ( deleteWhereCount
     , updateWhereCount
     , decorateSQLWithLimitOffset
+    , filterClauseHelper
     ) where
 
 import Control.Exception (throwIO)
--- a/persistent-postgresql/Database/Persist/query.sql
+++ b/persistent-postgresql/Database/Persist/query.sql
@@ -1,8 +1,0 @@
-INSERT INTO (
-  currencycloud_balances
-) VALUES (
-  externalAccountID, 
-  externalAmount
-) ON CONFLICT DO UPDATE SET (
-  amount = externalAmount
-) WHERE amount != externalAmount
--- a/persistent-postgresql/Database/Persist/query.sql
+++ b/persistent-postgresql/Database/Persist/query.sql
@@ -1,0 +1,8 @@
+INSERT INTO (
+  currencycloud_balances
+) VALUES (
+  externalAccountID, 
+  externalAmount
+) ON CONFLICT DO UPDATE SET (
+  amount = externalAmount
+) WHERE amount != externalAmount


persistent-postgresql/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for persistent-postgresql
 
+## 2.12.0.1
```
dmarticus, line 3: does this merit a more significant bump than like, a patch?  I'm not sure what the versioning convention for this project is.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1428,7 +1428,7 @@ copyField = CopyField
 --   []
 -- @
 --
--- Once we run that code on the datahase, the new data set looks like this:
+-- Once we run that code on the database, the new data set looks like this:
```
dmarticus, line 1431: couldn't help it ;)

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1,5 +1,9 @@
 {-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
```
dmarticus, line 6: I copied over a method from `MySQL.hs` to work with `postgresql`, and all of these imports were needed.  Should I alphabetize them manually?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -23,6 +27,8 @@ module Database.Persist.Postgresql
     , ConnectionString
     , PostgresConf (..)
     , PgInterval (..)
+    , upsertWhere
```
dmarticus, line 30: these are the two new methods I want to expose to any user of this library

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1487,15 +1493,24 @@ tableName = escapeE . tableDBName
 fieldName :: (PersistEntity record) => EntityField record typ -> Text
 fieldName = escapeF . fieldDBName
 
+fieldName' ::  forall record typ. (PersistEntity record, PersistEntityBackend record ~ SqlBackend) => EntityField record typ -> FieldNameDB
```
dmarticus: I copied over this method from `MySQL.hs` and it's only used in my new methods; is it worth keeping in?  Or should I give it a new name?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1487,15 +1493,24 @@ tableName = escapeE . tableDBName
...
+fieldName' f = fieldDB $ persistFieldDef f
+
 escapeC :: ConstraintNameDB -> Text
 escapeC = escapeWith escape
 
 escapeE :: EntityNameDB -> Text
 escapeE = escapeWith escape
 
+escapeES :: EntityNameDB -> String
```
dmarticus: same here; the implementation that I copied from `MySQL.hs` converted `EntityNameDB`s to `String`s instead of `Text`, so I updated the methods in this file to convert to `Strings` for my use case.  Should I just refactor the method to work with `Text` and then remove these (this, and then the `escapeFS` below) methods?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1504,6 +1519,14 @@ escape s =
     go ('"':xs) = "\"\"" ++ go xs
     go (x:xs) = x : go xs
 
+-- | Escape a database name to be included on a query.
+escapeDBName :: String -> String
```
dmarticus: basically does the same thing as the above method, but with `String`s instead of `Text`.  In reviewing this, it makes me want to just refactor the DB method to use `Text` instead of `String`s.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1761,310 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
     fields = keyAndEntityFields ent
     conflictColumns = escapeF . fieldDB <$> entityKeyFields ent
 
+-- | This type is used to determine how to update rows using Postgres'
+-- @INSERT ... ON CONFLICT KEY UPDATE@ functionality, exposed via
+-- 'upsertWhere' and 'upsertManyWhere' in this library.
+--
+-- @since 2.12.0.1
+data HandleUpdateCollision record where
```
dmarticus, line 1753: copied from `MySQL.hs`, should I just import it instead?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1761,310 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+     , PersistStore backend
+     , BackendCompatible SqlBackend backend
+     )
+  => record
+  -> SqlBackend
+  -> [Update record]
+  -> [Filter record]
+  -> ReaderT backend m ()
+upsertWhere record conn updates filts =
+  upsertManyWhere [record] conn [] updates filts
```
dmarticus, line 1785: I don't know if I love this, since it implies that the user is passing in the `SqlBackend` at call time and I don't like that.  Can I just have this method always use a `postgresql` backend?  What's the pattern for that?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1745,310 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+-- 'upsertWhere' and 'upsertManyWhere' in this library.
+--
+-- @since 2.12.0.1
+data HandleUpdateCollision record where
+  -- | Copy the field directly from the record.
+  CopyField :: EntityField record typ -> HandleUpdateCollision record
+  -- | Only copy the field if it is not equal to the provided value.
+  CopyUnlessEq :: PersistField typ => EntityField record typ -> typ -> HandleUpdateCollision record
+
+-- | Postgres specific 'upsertWhere'. This method does the following:
```
dmarticus, line 1758: is this haddock sufficient?

persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
```diff
@@ -6,6 +6,8 @@
 module Database.Persist.Sql.Orphan.PersistQuery
     ( deleteWhereCount
     , updateWhereCount
+    , filterClause
```
dmarticus, line 9: I'm exporting these because I want to use the filter helpers in `Postgresql.hs`

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1744,140 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+        ,     "?"
+        ,   "),"
+        ,   n
+        , ")"
+        ]
+    condFieldSets = map (uncurry mkCondFieldSet) fieldsToMaybeCopy
+    fieldSets = map (\n -> T.concat [n, "=VALUES(", n, ")"]) updateFieldNames
+    upds = map (Util.mkUpdateText' (escapeF) id) updates
+    updsValues = map (\(Update _ val _) -> toPersistValue val) updates
+    wher = if null filts then "" else filterClause False conn filts
```
dmarticus, line 1863: I don't like that have to use `conn` here, but it comes from an imported method.  Can I just instantiate a Postgres connection here and pass it in at execution time so the user doesn't have to provide it?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1744,140 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+upsertManyWhere [] _ _ _ _ = return ()
+upsertManyWhere records conn fieldValues updates conditions =
+  uncurry rawExecute $
+    mkBulkUpsertQuery records conn fieldValues updates conditions
+
+-- | This creates the query for 'upsertManyWhere'. If you
+-- provide an empty list of updates to perform, then it will generate
+-- a dummy/no-op update using the first field of the record. This avoids
+-- duplicate key exceptions.
+mkBulkUpsertQuery
```
dmarticus, line 1824: this method was copied from `MySQL.hs` as well.  I just had to change some of the type constraints.

persistent-postgresql/ChangeLog.md
parsonsmatt: An addition to the API is a minor version bump - so since we're adding a function, we'd go to 2.12.1.0

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: You can run `stylish-haskell` and it should work to make it right.

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: We don't want to import it, since this package is only for postgres stuff, and we don't want to depend on `persistent-mysql`. But we may want to upstream the definition to `persistent` somewhere so it can be shared. I'm happy to leave that as a later refactoring task - it's entirely possible that the two datatypes will need to diverge to support platform-specific features.

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: So `ReaderT backend` means that we have a `backend` implcity that we can get with `ask :: ReaderT r m r`. So instead of asking for the `conn` as a parameter, we can write:

```haskell
upsertManyWhere records updates filts = do
    conn <- asks projectBackend
    ...
```

(where `projectBackend` uses the `BackendCompatible SqlBackend backend` instance to get the `SqlBackend` out of the `backend`).

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1745,310 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+     )
+  => record
+  -> SqlBackend
+  -> [Update record]
+  -> [Filter record]
+  -> ReaderT backend m ()
+upsertWhere record conn updates filts =
+  upsertManyWhere [record] conn [] updates filts
+
+-- | Postgres specific 'upsertManyWhere'. This method does the following:
```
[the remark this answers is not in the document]
parsonsmatt: I'm a fan of duplicate docs. Can also link to the other function, like `See 'upsertManyWhere' for more information.`

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1745,310 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+        , ") "
+        , " VALUES "
+        , recordPlaceholders
+        , " ON CONFLICT DO UPDATE SET "
+        , updateText
+        , " WHERE "
+        , wher
+        ]
+
+filterClause :: (PersistEntity val, PersistEntityBackend val ~ SqlBackend)
```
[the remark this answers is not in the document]
parsonsmatt: Yeah, that makes sense! If it's not exported, then it really should be.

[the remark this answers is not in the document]
dmarticus: ooops yeah this comment was from when I didn't export it ;)

persistent-postgresql/Database/Persist/Postgresql.hs
dmarticus: yeah that's exactly the answer I was looking for, thanks!

.github/workflows/haskell.yml
```diff
@@ -57,6 +56,7 @@ jobs:
     env:
       CONFIG: "--enable-tests"
     steps:
+      - continue-on-error: true
```
parsonsmatt, line 59: i think we may want to put this on the haddock step

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -21,8 +25,11 @@ module Database.Persist.Postgresql
     , createPostgresqlPoolWithConf
     , module Database.Persist.Sql
     , ConnectionString
+    , HandleUpdateCollision
```
parsonsmatt, line 28: We're not currently exporting any of the constructors for this, nor are we exporting any way to create it.

Let's copy the functions from MySQL so we can do this.

We also need the `EntityField rec typ -> Filter rec` that'll return `False` when the `excluded` column is equal to the table's current column.

dmarticus: I was WONDERING why HLS was yelling at me!  Couldn't get the compiler to break locally so thank you for confirming that something was wrong.

.github/workflows/haskell.yml
```diff
@@ -82,4 +82,5 @@ jobs:
       - run: cabal v2-build all --disable-optimization $CONFIG
       - run: cabal v2-test all --disable-optimization $CONFIG
       - run: cabal v2-haddock all $CONFIG
+        continue-on-error: true
```
dmarticus, line 85: I _think_ this'll make it so that it only fails CI on doc errors, but I'm not sure.  Can someone else confirm it?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1754,176 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+  CopyField :: EntityField record typ -> HandleUpdateCollision record
+  -- | Only copy the field if it is not equal to the provided value.
+  CopyUnlessEq :: PersistField typ => EntityField record typ -> typ -> HandleUpdateCollision record
+
+-- | An alias for 'HandleUpdateCollision'. The type previously was only
+-- used to copy a single value, but was expanded to be handle more complex
+-- queries.
+--
+-- @since 2.12.1
+type SomeField = HandleUpdateCollision
```
dmarticus, line 1773: the rest of these types (L1773-1812) I copied over from `MySQL.hs`.  Do they make sense here?  It made everything compile but I'm not really sure what the difference between copying vs importing was.

persistent-postgresql/README.md
```diff
@@ -21,5 +21,5 @@ $ createdb test
 
 The tests do not pass a test and expect to connect with the `postgres` user.
 Ensure that peer authentication is allowed for this.
-An easy/insecure way to do this is to set the `METHOD` to `trust` for all the login methods in `/etc/postgresql/XX/main/pg_hba.coinf`.
+An easy/insecure way to do this is to set the `METHOD` to `trust` for all the login methods in `/etc/postgresql/XX/main/pg_hba.conf`.
```
dmarticus, line 24: had to ;)

persistent-postgresql/test/PgInit.hs
```diff
@@ -56,7 +56,7 @@ import Test.HUnit ((@?=),(@=?), Assertion, assertFailure, assertBool)
 import Test.QuickCheck
 
 import Control.Monad (unless, (>=>))
-import Control.Monad.IO.Class
+
```
dmarticus, line 59: this was a superfluous import, per my IDE

persistent/Database/Persist/Sql/Util.hs
```diff
@@ -220,6 +221,19 @@ mkUpdateText' escapeName refColumn x =
   where
     n = escapeName . fieldDB . updateFieldDef $ x
 
+mkPostgresUpdateText :: PersistEntity record => (FieldNameDB -> Text) -> (Text -> Text) -> Update record -> Text
```
dmarticus, line 224: `postgresql` has different update methods that use the `EXCLUDED` field on the SQL values, so I wrote a method to do operator updates that specific to `postgresql`.  Is this appropriate to use here?

persistent-postgresql/test/UpsertWhere.hs
```diff
@@ -0,0 +1,111 @@
...
+    it "updates existing records" $ runConnAssert $ do
+      deleteWhere ([] :: [Filter Item])
+      insertMany_ items
+      upsertManyWhere
+        items
+        []
+        [ItemQuantity +=. Just 1]
+        []
+    -- TODO: this test doesn't pass
+    it "only copies passing values" $ runConnAssert $ do
```
dmarticus, line 74: this test fails due to an error like this
```
UpsertWhere.upsertManyWhere only copies passing values
       uncaught exception: FormatError
       FormatError {fmtMessage = "8 single '?' characters, but 9 parameters", fmtQuery = "INSERT INTO \"item\"(\"name\", \"description\", \"price\", \"quantity\") VALUES (?, ?, ?, ?), (?, ?, ?, ?) ON CONFLICT (\"name\") DO UPDATE SET \"price\"=EXCLUDED.\"price\", \"quantity\"=EXCLUDED.\"quantity\"", fmtParams = ["item1","","6.0","0","item2","hello world","null","0","0"]}
```

Where in my formatting am I missing the value that I need?

persistent-postgresql/test/UpsertWhere.hs
```diff
@@ -0,0 +1,111 @@
...
+      insertMany_ items
+      upsertManyWhere
+        (newItem : items)
+        []
+        []
+        []
+      dbItems <- sort . fmap entityVal <$> selectList [] []
+      dbItems @== sort (newItem : items)
+    -- TODO: this test doesn't pass
+    it "inserts without modifying existing records if no updates specified and there's a filter" $
```
dmarticus, line 100: similar to the test above, this test also fails with a format error:
```
uncaught exception: FormatError
       FormatError {fmtMessage = "13 single '?' characters, but 12 parameters", fmtQuery = "INSERT INTO \"item\"(\"name\", \"description\", \"price\", \"quantity\") VALUES (?, ?, ?, ?), (?, ?, ?, ?), (?, ?, ?, ?) ON CONFLICT (\"name\") DO UPDATE SET \"name\"=EXCLUDED.\"name\" WHERE (\"description\"=?)", fmtParams = ["item3","hi friends!","null","null","item1","","3.0","null","item2","hello world","null","2"]}
```

I imagine this issue is similar to the one before; is it obvious what I'm doing wrong here?  I spent all afternoon on Friday looking at this and couldn't grok what I was doing wrong.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1754,176 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+        [] -> T.concat [firstField, "=EXCLUDED.", firstField]
+        xs -> Util.commaSeparated xs
+    q = T.concat
+        [ "INSERT INTO "
+        , nameOfTable
+        , Util.parenWrapped . Util.commaSeparated $ entityFieldNames
+        , " VALUES "
+        , recordPlaceholders
+        , " ON CONFLICT "
+        , Util.parenWrapped $ firstField
```
dmarticus, line 1921: I _think_ this is wrong, shouldn't the conflict field be conditional to what the user wants to pass in, and not guaranteed to be the first field?

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1754,176 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+        , " DO UPDATE SET "
+        , updateText
+        , wher
```
dmarticus, line 1924: this where clause is giving me the correct SQL but it's still failing its test

persistent/Database/Persist/Sql/Util.hs
parsonsmatt: If it's specific to postgres, it's probably best for it to live in the `persistent-potgresql` package - `Database.Persist.Postgresql`, perhaps in an `.Internal` module

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: Copying means that they are *distinct* - the type `Database.Persist.Postgresql.HandleUpdateCollision` is not the same as the one defined in `Database.Persist.MySQL`, and you'll get a type error if you try to conflate the two.

I'd probably *not* copy this type synonym - it's a backwards compatibility thing. Since we're starting fresh, we can omit it.

.github/workflows/haskell.yml
parsonsmatt: That's my understanding too!

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: So, the way `upsert` works is that it requires a *single* `Unique` key for the record, and that `Unique` is what is used to trigger the `update` part instead of the `insert` part. If the entity has many `Unique` keys, you're required to use `upsertBy`, which accepts the `Unique` in question.

MySQL doesn't explicitly list the conflict target - it triggers if *any* uniqueness key is duplicated. 

I wonder what happens if there's a uniqueness constraint that is violated by an `INSERT` that isnt' listed here. I'd imagine it just errors. Worth investigating!

I think we can just list *all* of the uniqueness keys here - the primary key + all `Unique`s that persistent knows about.

persistent-postgresql/test/UpsertWhere.hs
parsonsmatt: Let's write some tests for the `mkBulkUpsertQuery` to see what's getting generated here.

I think we'd need to see the generated SQL to get an easy sense of what's going on here.

.github/workflows/haskell.yml
dmarticus: I confirmed it because this CI failed :joy:

persistent-postgresql/Database/Persist/Postgresql.hs
dmarticus: good call, and thanks for the explanation!

persistent-postgresql/test/UpsertWhere.hs
dmarticus: yeah I should've included it; here's the generated SQL
```
INSERT INTO "item"("name", "description", "price", "quantity") VALUES (?, ?, ?, ?), (?, ?, ?, ?) ON CONFLICT ("name") DO UPDATE SET "price"=EXCLUDED."price", "quantity"=EXCLUDED."quantity"
```

persistent-postgresql/test/UpsertWhere.hs
dmarticus: here's the generated SQL for this failure
```
INSERT INTO "item"("name", "description", "price", "quantity") VALUES (?, ?, ?, ?), (?, ?, ?, ?), (?, ?, ?, ?) ON CONFLICT ("name") DO UPDATE SET "name"=EXCLUDED."name" WHERE ("description"=?)
```

persistent-postgresql/test/UpsertWhere.hs
parsonsmatt: Can you also dump the `[PersistValue]` that we're generating? The error says that we're giving it `9` parameters.

dmarticus:
```
[PersistText "item1",PersistText "",PersistDouble 6.0,PersistInt64 0,PersistText "item2",PersistText "hello world",PersistNull,PersistInt64 0,PersistInt64 0]
```

persistent-postgresql/test/UpsertWhere.hs
dmarticus: and the `PersistValue`
```
[PersistText "item3",PersistText "hi friends!",PersistNull,PersistNull,PersistText "item1",PersistText "",PersistDouble 3.0,PersistNull,PersistText "item2",PersistText "hello world",PersistNull,PersistInt64 2]
```

persistent-postgresql/test/UpsertWhere.hs
parsonsmatt: So we have `PersistInt64 0` in there twice. We should have the first one (since that't the `Just 0` from the `quantity`), but the second one is spurious.

The `DO UPDATE` clause mentions price and quantity, and we've got `copyField ItemPrice` which is fine and `copyUnlessEq ItmeQuantity (Just 0)`. But the generated SQL has `DO UPDATE SET "price"=EXCLUDED."price", "quantity"=EXCLUDED."quantity"` - these are both `CopyField`-looking to me. Looking at the implementation...

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1754,176 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+    entityDef' = entityDef records
+    firstField = case entityFieldNames of
+        [] -> error "The entity you're trying to insert does not have any fields."
+        (field:_) -> field
+    entityFieldNames = map fieldDbToText (entityFields entityDef')
+    nameOfTable = escapeE . entityDB $ entityDef'
+    copyUnlessValues = map snd fieldsToMaybeCopy
+    recordValues = concatMap (map toPersistValue . toPersistFields) records
+    recordPlaceholders = Util.commaSeparated $ map (Util.parenWrapped . Util.commaSeparated . map (const "?") . toPersistFields) records
+    mkCondFieldSet n _ = T.concat [n, "=EXCLUDED.", n]
```
parsonsmatt, line 1905: Continuing from the test suite comment...

Here, we're throwing away the `_ :: PersistValue`. That's probably what the problem is - we'd need to actually make this a conditional update.

dmarticus: Aha!  Good find!  

> we'd need to actually make this a conditional update.

What do you mean by _conditional update_, like would I need to do something where I map an operator between the text and persistent value passed in that varies depending on what the type of update is?

persistent-postgresql/Database/Persist/Postgresql.hs
dmarticus: donezo

persistent-postgresql/test/UpsertWhere.hs
```diff
@@ -0,0 +1,110 @@
...
+      sort dbItems @== sort (newItem : items)
+    it "updates existing records" $ runConnAssert $ do
+      deleteWhere ([] :: [Filter Item])
+      insertMany_ items
+      upsertManyWhere
+        items
+        []
+        [ItemQuantity +=. Just 1]
+        []
+    -- TODO: this test doesn't pass
```
dmarticus, line 73: this test is still failing in the following way
```
  1) UpsertWhere.upsertManyWhere only copies passing values
       expected: [Item {itemName = "item1", itemDescription = "", itemPrice = Just 6.0, itemQuantity = Nothing},Item {itemName = "item2", itemDescription = "hello world", itemPrice = Nothing, itemQuantity = Just 2}]
        but got: [Item {itemName = "item1", itemDescription = "", itemPrice = Just 6.0, itemQuantity = Nothing},Item {itemName = "item2", itemDescription = "hello world", itemPrice = Nothing, itemQuantity = Nothing}]
```

and here's the values + query
```
INSERT INTO "item"("name", "description", "price", "quantity") VALUES (?, ?, ?, ?), (?, ?, ?, ?) ON CONFLICT ("name") DO UPDATE SET "price"=EXCLUDED."price", "quantity"=COALESCE(NULLIF(EXCLUDED."quantity",?))
[PersistText "item1",PersistText "",PersistDouble 6.0,PersistInt64 0,PersistText "item2",PersistText "hello world",PersistNull,PersistInt64 0,PersistInt64 0]
```

any ideas as to what I could be missing?

persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
```diff
@@ -225,18 +228,19 @@ dummyFromFilts _ = Nothing
...
 filterClauseHelper :: (PersistEntity val, PersistEntityBackend val ~ SqlBackend)
              => Bool -- ^ include table name?
              -> Bool -- ^ include WHERE?
+             -> Bool -- ^ use postgresl EXCLUDE
              -> SqlBackend
              -> OrNull
              -> [Filter val]
              -> (Text, [PersistValue])
-filterClauseHelper includeTable includeWhere conn orNull filters =
+filterClauseHelper includeTable includeWhere includeExcluded conn orNull filters =
```
dmarticus, line 243: I added this kinda ratchet param because the filterClause was breaking on the `EXCLUDED` value for PostgreSQL again.  See the next comment for where I use it...

persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
```diff
@@ -335,7 +339,12 @@ filterClauseHelper includeTable includeWhere conn orNull filters =
                                 , qmarks
                                 , ")"
                                 ], notNullVals)
-                            _ -> (name <> showSqlFilter pfilter <> "?" <> orNullSuffix, allVals)
+                            _ -> (T.concat 
+                                [
+                                  if includeExcluded then "EXCLUDED." else ""
```
dmarticus, line 344: for here, we add the `EXCLUDED` keyword if the `filterClauseHelper` is being called in the context of PostgreSQL.  Is there a more elegant change I can make here?  I don't love how this is an argument that is coming from a boolean that is passed in a call time.

persistent-postgresql/test/UpsertWhere.hs
parsonsmatt: So, we expect the `item2 { itemQuantiuty = Just 2 }`, but it's actually `Nothing`. 

We've inserted `Just 2` as the value for the field in the database. The command we call is `upsertManyWhere newItems [ copyUnlessEq ItemQuantity (Just 0), .. ] [] []`, and for `newItems`, we set `itemQuantity = Just 0`. So, what we should expect here is that, since `Excluded.ItemQuantity == Just 0`, we shouldn't copy it over.

But, instead, we've got `NULL` in the database! So our `upsert` SQL is causing it to be `NULL` for some reason.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1751,178 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+      T.concat
+        [ n
+            ,"=COALESCE("
+              ,"NULLIF("
+                ,"EXCLUDED."
+                  ,n
+                ,","
+              ,"?"
+            ,")"
+          ,")"
```
parsonsmatt, line 1902: I think the problem is here. This text will expand out to:

```
n = COALESCE( NULLIF( EXCLUDED.n, ?))
```

But what we want is:

```
n = COALESCE( NULLIF( EXCLUDED.n, ?), n)
```

```suggestion
          , ", "
          , n
          ,")"
```

parsonsmatt (commented):
almost there!

persistent-postgresql/Database/Persist/Postgresql.hs
dmarticus: hmmm I tested this and it still didn't completely work; now I have this:
```
INSERT INTO "item"("name", "description", "price", "quantity") VALUES (?, ?, ?, ?), (?, ?, ?, ?) ON CONFLICT ("name") DO UPDATE SET "price"=EXCLUDED."price", "quantity"=COALESCE(NULLIF(EXCLUDED."quantity",?), EXCLUDED."quantity")
[PersistText "item1",PersistText "",PersistDouble 6.0,PersistInt64 0,PersistText "item2",PersistText "hello world",PersistNull,PersistInt64 0,PersistInt64 0]
```

and the test still fails with
```
expected: [Item {itemName = "item1", itemDescription = "", itemPrice = Just 6.0, itemQuantity = Nothing},Item {itemName = "item2", itemDescription = "hello world", itemPrice = Nothing, itemQuantity = Just 2}]
        but got: [Item {itemName = "item1", itemDescription = "", itemPrice = Just 6.0, itemQuantity = Just 0},Item {itemName = "item2", itemDescription = "hello world", itemPrice = Nothing, itemQuantity = Just 0}]
```

dmarticus: It's so odd to me that it expects `Nothing` but it returns a `Just 0` instead and somehow that's a different thing.  When I change that parameter to `Nothing` it still fails, just in a different way

```
  1) UpsertWhere.upsertManyWhere only copies passing values
       expected: [Item {itemName = "item1", itemDescription = "", itemPrice = Just 6.0, itemQuantity = Nothing},Item {itemName = "item2", itemDescription = "hello world", itemPrice = Nothing, itemQuantity = Just 2}]
        but got: [Item {itemName = "item1", itemDescription = "", itemPrice = Just 6.0, itemQuantity = Nothing},Item {itemName = "item2", itemDescription = "hello world", itemPrice = Nothing, itemQuantity = Nothing}]
```

persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
```diff
@@ -366,6 +373,8 @@ filterClauseHelper includeTable includeWhere conn orNull filters =
         name =
             (if includeTable
                 then ((tn <> ".") <>)
+                else if includeExcluded -- need this for PostgreSQL queries
```
dmarticus, line 376: here's where I pass in the `includeExcluded` param; I tried to follow the same pattern as `includeTable`; does this seem to make sense?  All the tests that use this code still pass since the param is only used in `filterClauseWithVals`.

persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
```diff
@@ -291,51 +295,52 @@ filterClauseHelper includeTable includeWhere conn orNull filters =
...
                                 , name
                                 , " NOT IN "
                                 , qmarks
-                                , ")"
                                 ], notNullVals)
-                            _ -> (name <> showSqlFilter pfilter <> "?" <> orNullSuffix, allVals)
+                            _ -> (T.concat 
+                                [name
+                                   <> showSqlFilter pfilter <> "?" <> orNullSuffix
+                                ], allVals)
```
parsonsmatt, line 343: these formatting changes make it a bit difficult to understand the signal v noise, can you back them out?

persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
parsonsmatt: Yeah, I agree this is a bit dicey.

`Bool`ean blindness is a problem here, and since this is the first time we're 'exposing' this function, we can fix that.

Instead of `includeExcluded :: Bool` and `includeTable :: Bool` (note that `includeTable = True` means that `includeExcluded` *doesn't matter*!), let's make a new type:

```haskell
data FilterTablePrefix
    = PrefixTableName -- ^ Include the table name
    | PrefixExcluded -- ^ Valid only on Postgresql
```

Then we pass a `Maybe FilterTablePrefix` instead of the two `Bool` arguments - `Nothing` corresponds to `False, False`.

persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
dmarticus: yes!

persistent/Database/Persist/Sql/Orphan/PersistQuery.hs
dmarticus: this is a great way to data model

persistent-postgresql/Database/Persist/Postgresql/Util.hs
```diff
@@ -0,0 +1,30 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Database.Persist.Postgresql.Util (
+  mkPostgresUpdateText
```
parsonsmatt, line 4: Since this module only has one function, let's just put it directly in `Database.Persist.Postgresql`

persistent-postgresql/Database/Persist/Postgresql/Util.hs
```diff
@@ -0,0 +1,30 @@
...
+    Divide -> T.concat [n, "=EXCLUDED.", refColumn n, "/?"]
+    BackendSpecificUpdate up ->
+      error . T.unpack $ "mkUpdateText: BackendSpecificUpdate " <> up <> " not supported"
+  where
+    n = escapeName . fieldDB . updateFieldDef $ x
+
+-- | Gets the 'FieldDef' for an 'Update'.
+updateFieldDef :: PersistEntity v => Update v -> FieldDef
+updateFieldDef (Update f _ _) = persistFieldDef f
+updateFieldDef BackendUpdate {} = error "updateFieldDef: did not expect BackendUpdate"
```
parsonsmatt, line 30: For a general purpose library function, we shouldn't ever call `error` when we're handling inputs. Instead I'd write this with a `Maybe FieldDef` return type.

Also, since this is only used in `mkPostgresUpdateText`, we can inline it directly and perform the pattern match there.

persistent-postgresql/Database/Persist/Postgresql/Util.hs
```diff
@@ -0,0 +1,30 @@
...
+  case updateUpdate x of
+    Assign -> n <> "=?"
+    Add -> T.concat [n, "=EXCLUDED.", refColumn n, "+?"]
+    Subtract -> T.concat [n, "=EXCLUDED.", refColumn n, "-?"]
+    Multiply -> T.concat [n, "=EXCLUDED.", refColumn n, "*?"]
+    Divide -> T.concat [n, "=EXCLUDED.", refColumn n, "/?"]
+    BackendSpecificUpdate up ->
+      error . T.unpack $ "mkUpdateText: BackendSpecificUpdate " <> up <> " not supported"
+  where
+    n = escapeName . fieldDB . updateFieldDef $ x
```
parsonsmatt, line 25: Following on with the above - I think it'd be better to write this whole function like:

```haskell
mkPostgresUpdateText escapeName refColumn upd = 
    case upd of
        Update f op _ ->
            let 
                n = escapeName . fieldDB . persistField $ f
            in 
                case op of
                    Assign -> Right $  n <> "=?"
            -- ...
        BackendSpecificUpdate up ->
            Left $ errorMessage
```

That's more a style thing, though.

persistent-postgresql/Database/Persist/Postgresql/Util.hs
```diff
@@ -0,0 +1,30 @@
...
+
+import Data.Monoid ((<>))
+import Data.Text (Text)
+import qualified Data.Text as T
+
+import Database.Persist (FieldNameDB, PersistEntity (..), PersistUpdate(..), Update(..), updateUpdate, fieldDB, FieldDef)
+
+
+mkPostgresUpdateText :: PersistEntity record => (FieldNameDB -> Text) -> (Text -> Text) -> Update record -> Text
+mkPostgresUpdateText escapeName refColumn x =
```
parsonsmatt, line 15: So I'm not sure about this function! At least, not as a general purpose function. We only call it in the `mkBulkUpsertQuery`, which suggests to me that it should be local to that function. I'm also not sure that the name makes sense to me - this function is turning an `Update record` into the actual assignment statement, so something like `RecordField +=. 1` would become `"table_name"."field" = EXCLUDED."field" + 1`.

So, this is *only* valid during the `ON CONFLICT DO UPDATE` part of a query - and it does also mean that we're using the `EXCLUDED` for the source, rather than the pre-existing database table. We're calling this on the `[Update rec]`, which is *not* what we want! The `[Update rec]` is supposed to be a list of `Update`s that are *not* dependent on the inserted record.

See the comment on the parameter:

```haskell
  -- | A list of the updates to apply that aren't dependent on the record being inserted.
  [Update record] ->
```

I think we acn just use `Util.mkUpdateText`, like the MySQL version does.

parsonsmatt (changes requested):

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1751,181 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+-- For example, here's how this method could be called like such:
+--
+-- upsertManyWhere [record] [recordField =. newValue] [recordField /= newValue]
+--
+-- Called thusly, this method will insert a new record (if none exists) OR update a recordField with a new value
+-- assuming the condition in the last block is met.
+--
+-- -- @since 2.12.1.0
+upsertManyWhere ::
+  forall record backend m.
```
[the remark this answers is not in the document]
dmarticus: oh my bad, I tried running `stylish-haskell` before pushing; I'll try again.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1751,181 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+  ( backend ~ PersistEntityBackend record,
+    BackendCompatible SqlBackend backend,
+    PersistEntityBackend record ~ SqlBackend,
```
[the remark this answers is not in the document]
dmarticus: I needed to add it to get it compiling tbh; when I get rid of it I get this error 
```
/Users/dmarticus/Personal/github.com/dmarticus/persistent/persistent-postgresql/Database/Persist/Postgresql.hs:1861:5: error:
    • Could not deduce: backend ~ SqlBackend
        arising from a use of ‘mkBulkUpsertQuery’
      from the context: (backend ~ PersistEntityBackend record,
                         BackendCompatible SqlBackend backend, PersistEntity record,
                         MonadIO m)
        bound by the type signature for:
                   upsertManyWhere :: forall record backend (m :: * -> *).
                                      (backend ~ PersistEntityBackend record,
                                       BackendCompatible SqlBackend backend, PersistEntity record,
                                       MonadIO m) =>
                                      [record]
                                      -> [HandleUpdateCollision record]
                                      -> [Update record]
                                      -> [Filter record]
                                      -> ReaderT backend m ()
        at Database/Persist/Postgresql.hs:(1841,1)-(1856,22)
      ‘backend’ is a rigid type variable bound by
        the type signature for:    
          upsertManyWhere :: forall record backend (m :: * -> *).
                             (backend ~ PersistEntityBackend record,
                              BackendCompatible SqlBackend backend, PersistEntity record,
                              MonadIO m) =>
                             [record]
                             -> [HandleUpdateCollision record]
                             -> [Update record]
                             -> [Filter record]
                             -> ReaderT backend m ()
        at Database/Persist/Postgresql.hs:(1841,1)-(1856,22)
    • In the second argument of ‘($)’, namely
        ‘mkBulkUpsertQuery records conn fieldValues updates filters’
      In a stmt of a 'do' block:   
        uncurry rawExecute         
          $ mkBulkUpsertQuery records conn fieldValues updates filters
      In the expression:           
        do conn <- asks projectBackend
           uncurry rawExecute      
             $ mkBulkUpsertQuery records conn fieldValues updates filters
    • Relevant bindings include    
        upsertManyWhere :: [record]
                           -> [HandleUpdateCollision record]
                           -> [Update record]
                           -> [Filter record]
                           -> ReaderT backend m ()
          (bound at Database/Persist/Postgresql.hs:1857:1)
```

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1751,181 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+  conn <- asks projectBackend
+  uncurry rawExecute $
+    mkBulkUpsertQuery records conn fieldValues updates filters
+
+-- | This creates the query for 'upsertManyWhere'. If you
+-- provide an empty list of updates to perform, then it will generate
+-- a dummy/no-op update using the first field of the record. This avoids
+-- duplicate key exceptions.
+mkBulkUpsertQuery
+    :: (PersistEntity record, PersistEntityBackend record ~ SqlBackend)
```
parsonsmatt, line 1868: I don't think we need the `PersistEntityBackend record ~ SqlBackend` constraint here, either.
```suggestion
    :: (PersistEntity record)
```

persistent-postgresql/Database/Persist/Postgresql.hs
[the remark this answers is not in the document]
parsonsmatt: OK, I see. It's because we're requiring it in `mkBulkUpsertQuery`. But that's also unnecessary. If we delete it from that function, then we can delete it here, too.

persistent-postgresql/Database/Persist/Postgresql.hs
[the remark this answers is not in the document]
parsonsmatt: `stylish-haskell` won't do formatting changes like that :\ It's fine to leave as-is, it can ge trefactored later.

persistent-postgresql/Database/Persist/Postgresql/Util.hs
dmarticus: This is good feedback!  I think I must be doing something else wrong to work around this, then, because isn't a query like this:

```
INSERT INTO "item"("name", "description", "price", "quantity") 
VALUES (?, ?, ?, ?), (?, ?, ?, ?), (?, ?, ?, ?) 
ON CONFLICT ("name") 
DO UPDATE SET "quantity"=EXCLUDED."quantity"+? 
WHERE (EXCLUDED."description"=?)
```

the type of query we want?  When I run the query without referencing the `EXCLUDED` table, the SQL bugs out.

dmarticus: Like am I just working around something i don't understand?

parsonsmatt: Yeah. It's giving an error because the `"field_name"` is going to be ambiguous, and in a `ON CONFLICT DO UPDATE` there's going to *also* be the `EXCLUDED` table available with the same field names, so now we *have* to prefix the update with either the table name or the `EXCLUDED` name.

So I think we can do the same thing as we did with the `Prefix_` type - accept that value, and if it is `PrefixExcluded`, then we'll make it `EXCLUDED`, otherwise we prefix with the table name.

persistent-postgresql/Database/Persist/Postgresql.hs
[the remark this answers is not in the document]
dmarticus: OH good to know

persistent-postgresql/Database/Persist/Postgresql.hs
dmarticus: oh maybe THAT was it

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1738,6 +1751,214 @@ repsertManySql ent n = putManySql' conflictColumns fields ent n
...
+  -> [Filter record]
+  -> ReaderT backend m ()
+upsertWhere record updates filts =
+  upsertManyWhere [record] [] updates filts
+
+-- | Exclude any record field if it doesn't match the filter record.  Used only in `upsertWhere` and
+-- `upsertManyWhere`
+--
+-- @since 2.12.1.0
+-- TODO: we could probably make a sum type for the `Filter` record that's passed into the `upserWhere` and
```
dmarticus, line 1833: @parsonsmatt does this TODO say enough/make sense?

persistent-postgresql/test/UpsertWhere.hs
```diff
@@ -0,0 +1,146 @@
...
+        deleteWhere ([] :: [Filter Item])
+        insertMany_ items
+        upsertManyWhere
+          (newItem : items)
+          []
+          [ItemQuantity +=. Just 1]
+          [ItemDescription ==. "bye friends!"]
+        dbItems <- sort . fmap entityVal <$> selectList [] []
+        dbItems @== sort (newItem : items)
+    it "inserts an item and excludes a field if it matches the filter" $
```
dmarticus, line 135: here's the new test; I _think_ it does what I'm attempting to test for?

persistent-postgresql/test/UpsertWhere.hs
```diff
@@ -0,0 +1,146 @@
...
+    it "inserts and modifies existing records if there are updates specified and the filter doesn't apply" $
+      runConnAssert $ do
+        let newItem = Item "item3" "hi friends!" Nothing Nothing
+        deleteWhere ([] :: [Filter Item])
+        insertMany_ items
+        upsertManyWhere
+          (newItem : items)
+          []
+          [ItemQuantity +=. Just 1]
+          [ItemDescription ==. "bye friends!"]
```
parsonsmatt, line 132: So, with this in `[Filter Item]`, we won't ever apply the update, because none of the `ItemDescription`s are equal to `"bye friends!"`.

We don't actually have any tests that show that the `[Update Item]` is actually being performed, which I think is a bit of a problem. We do test that `[HandleUpdateCollision Item]` is being performed.

I'd like to see one more test - one that has an item in each of the lists - and then an assertion that each item was successfully used. I can merge this into a temp branch and then finish that out

persistent-postgresql/test/UpsertWhere.hs
```diff
@@ -0,0 +1,146 @@
...
+      dbItems <- map entityVal <$> selectList [] []
+      sort dbItems @== sort (newItem : items)
+    it "updates existing records" $ runConnAssert $ do
+      deleteWhere ([] :: [Filter Item])
+      insertMany_ items
+      upsertManyWhere
+        items
+        []
+        [ItemQuantity +=. Just 1]
+        []
```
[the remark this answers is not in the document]
dmarticus: lmao whoops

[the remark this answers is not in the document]
dmarticus: the original test doesn't have one either I'm gonna dd it

persistent-postgresql/test/UpsertWhere.hs
parsonsmatt: Sort of ya - let's ensure that it a) does not update a record that should not be updated, and b) does update a record that should be updated

dmarticus:
@parsonsmatt did I address all the changes needed?

persistent-mysql/test/InsertDuplicateUpdate.hs
```diff
@@ -61,12 +61,15 @@ specs = describe "DuplicateKeyUpdate" $ do
...
     it "updates existing records" $ db $ do
+      let postUpdate = map (\i -> i { itemQuantity = fmap (+1) (itemQuantity i) }) items
       deleteWhere ([] :: [Filter Item])
       insertMany_ items
       insertManyOnDuplicateKeyUpdate
         items
         []
         [ItemQuantity +=. Just 1]
+      dbItems <- sort . fmap entityVal <$> selectList [] []
+      dbItems @== sort postUpdate
```
parsonsmatt, line 72: thank you for fixing this 😄

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: Yeah!

parsonsmatt (approved):

PR #1223, opened 2021-04-01
Fix ToJSON for PersistValue

parsonsmatt:

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Types/Base.hs
+++ b/persistent/Database/Persist/Types/Base.hs
@@ -654,9 +654,15 @@ instance A.ToJSON PersistValue where
     toJSON PersistNull = A.Null
     toJSON (PersistList l) = A.Array $ V.fromList $ map A.toJSON l
     toJSON (PersistMap m) = A.object $ map (second A.toJSON) m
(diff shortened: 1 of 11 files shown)


persistent/Database/Persist/Types/Base.hs
```diff
@@ -654,9 +652,15 @@ instance A.ToJSON PersistValue where
...
-    toJSON (PersistLiteralEscaped b) = A.String $ T.cons 'e' $ TE.decodeUtf8 $ B64.encode b
+    toJSON (PersistLiteral_ litTy b) =
+        let encoded = TE.decodeUtf8 $ B64.encode b
+            prefix =
+                case litTy of
+                    DbSpecific -> 'p'
+                    Unescaped -> 'l'
+                    Escaped -> 'e'
+         in
+            A.String $ T.cons prefix encoded
```
parsonsmatt, line 663: This is *exactly* the problem I was trying to solve, and *exactly* the problem that my patch introduced. Sigh. I'm disappointed that I didn't catch it.

So - the patch makes `PersistDbSpecific` a catch-all match, expanding to `PersistLiteral_ _ bs`. Which means that the `PersistDbSpecific` woudl encode *all* values with the `p` prefix! That's bad. This fixes that problem.

PR #1225, opened 2021-04-07
Internalize SqlBackend

parsonsmatt:
This PR makes the `SqlBackend` type abstract (sort of). Breaking changes to the `SqlBackend` API will no longer be represented in the major version number of the code, which should make it much easier to keep `persistent` versions forwards compatible.

API consumers are recommended to use the functions and helpers defined in `Database.Persist.SqlBackend`.

After submitting your PR:


persistent/Database/Persist/Quasi.hs
```diff
@@ -419,838 +419,6 @@ module Database.Persist.Quasi
     , upperCaseSettings
     , lowerCaseSettings
     , nullable
-#if TEST
```
parsonsmatt, line 422: All these changes are mostly to support removing the need for this bit of CPP, and also so the test-sutie depends on the library instead of the source files directly.

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -0,0 +1,851 @@
...
+{-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
+
+-- | This @Internal@ module may have breaking changes that will not be reflected
+-- in major version bumps. Please use "Database.Persist.Quasi" instead.
+--
+-- @since 2.13.0.0
+module Database.Persist.Quasi.Internal
```
parsonsmatt, line 13: All this is copied directly from the old module.

persistent/Database/Persist/SqlBackend.hs
```diff
@@ -0,0 +1,150 @@
+-- | This module contains types and information necessary for a SQL database.
+-- Database support libraries, like @persistent-postgresql@, will be responsible
+-- for constructing these values.
+module Database.Persist.SqlBackend
+    ( -- * The type and construction
+      SqlBackend
+    , mkSqlBackend
+    , MkSqlBackendArgs(..)
```
parsonsmatt, line 8: Breaking changes to the `MkSqlBackendArgs` type (representing *required* parameters) will be a breaking change to the library.

persistent/persistent.cabal
```diff
@@ -66,6 +73,8 @@ library
                      Database.Persist.Class.PersistField
                      Database.Persist.Class.PersistStore
 
+    other-modules:   Database.Persist.Types.Base
```
parsonsmatt, line 76: probably going to further reorganize the module hierarchy here

PR #1231, opened 2021-04-20
Simplify Line type in Quasi module, always use NonEmpty

danbroooks:

After submitting your PR:

----

This is a followon from https://github.com/yesodweb/persistent/pull/1206 that addresses the TODO comment here:

https://github.com/yesodweb/persistent/blob/53359c834f649442c2f17379e24d78fb6e8719d4/persistent/Database/Persist/Quasi.hs#L546-L555

Following recent refactoring there is not actually any instance of `Line` where the contents of the list are _not_ `NonEmpty`, so this PR changes that type to always use `NonEmpty`, rather than being polymorphic, which greatly simplifies a lot of the code around `Line`, and means we can also get rid of the `skipEmpty` function, as `NonEmpty` guarantees us nonempty-ness :+1:

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,9 +1,12 @@
 # Changelog for persistent
 
-## 2.12.1.0
+## 2.12.1.1
 
 * [#1231](https://github.com/yesodweb/persistent/pull/1231)
     * Simplify Line type in Quasi module, always use NonEmpty 
+
+## 2.12.1.0
+
 * [#1218](https://github.com/yesodweb/persistent/pull/1218)
     * Refactoring name generating functions in TH 
 * [#1226](https://github.com/yesodweb/persistent/pull/1226)
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -2,6 +2,8 @@
 
 ## 2.12.1.0
 
+* [#1231](https://github.com/yesodweb/persistent/pull/1231)
+    * Simplify Line type in Quasi module, always use NonEmpty 
 * [#1218](https://github.com/yesodweb/persistent/pull/1218)
     * Refactoring name generating functions in TH 
 * [#1226](https://github.com/yesodweb/persistent/pull/1226)
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -421,12 +421,11 @@ module Database.Persist.Quasi
     , nullable
 #if TEST
     , Token (..)
-    , Line' (..)
+    , Line (..)
     , preparse
     , parseLine
     , parseFieldType
     , associateLines
-    , skipEmpty
     , LinesWithComments(..)
     , splitExtras
     , takeColsEx
@@ -543,16 +542,10 @@ preparse txt = do
     lns <- NEL.nonEmpty (T.lines txt)
     NEL.nonEmpty $ mapMaybe parseLine (NEL.toList lns)
 
--- TODO: refactor to return (Line' NonEmpty), made possible by
--- https://github.com/yesodweb/persistent/pull/1206 but left out
--- in order to minimize the diff
 parseLine :: Text -> Maybe Line
-parseLine txt =
-    case tokenize txt of
-      [] ->
-          Nothing
-      toks ->
-          pure $ Line (parseIndentationAmount txt) toks
+parseLine txt = do
+    parsedTokens <- NEL.nonEmpty (tokenize txt)
+    pure $ Line (parseIndentationAmount txt) parsedTokens
 
 -- | A token used by the parser.
 data Token = Token Text    -- ^ @Token tok@ is token @tok@ already unquoted.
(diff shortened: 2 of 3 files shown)


persistent/ChangeLog.md
```diff
@@ -2,6 +2,8 @@
 
 ## 2.12.1.0
 
+* [#1231](https://github.com/yesodweb/persistent/pull/1231)
+    * Simplify Line type in Quasi module, always use NonEmpty
```
parsonsmatt, line 6: `2.12.1.0` has already been released, this is going to get bundled in with 2.12.1.1

persistent/Database/Persist/Quasi.hs
```diff
@@ -543,16 +542,10 @@ preparse txt = do
...
 parseLine :: Text -> Maybe Line
-parseLine txt =
-    case tokenize txt of
-      [] ->
-          Nothing
-      toks ->
-          pure $ Line (parseIndentationAmount txt) toks
+parseLine txt = do
+    parsedTokens <- NEL.nonEmpty (tokenize txt)
+    pure $ Line (parseIndentationAmount txt) parsedTokens
```
parsonsmatt, line 548: Nice 😄

persistent/Database/Persist/Quasi.hs
```diff
@@ -622,46 +615,32 @@ tokenize t
...
-type Line = Line' []
-
 lowestIndent
     :: Functor f
     => Foldable f
-    => Functor g
-    => f (Line' g)
+    => f Line
     -> Int
 lowestIndent = minimum . fmap lineIndent
```
parsonsmatt, line 632: Can we specialize this function, too? I feel like we can get away with specializing this, possibly even to `NonEmpty` which would make it even safer (`mimumum [] = undefined`).

persistent/test/main.hs
```diff
@@ -212,7 +214,7 @@ main = hspec $ do
         it "mid-token quote in later token" $
             parseLine "foo bar baz=(bin\")" `shouldBe`
                 Just
-                    ( Line 0
+                    ( Line 0 $ nonEmptyOrFail
```
parsonsmatt, line 217: For the test suite, we could enable `OverloadedLists`, and then we can use list literal syntax for `NonEmpty a` - it's a gross kludge to write `[] :: NonEmpty Int`, but, lmao, it does make the test syntax nicer.

persistent/test/main.hs
```diff
@@ -900,8 +896,11 @@ takePrefix :: Value -> Value
 takePrefix (String a) = String (T.take 1 a)
 takePrefix a = a
 
-asTokens :: [T.Text] -> [Token]
-asTokens = fmap Token
+nonEmptyOrFail :: [a] -> NonEmpty a
+nonEmptyOrFail = maybe failure id . NEL.nonEmpty
```
parsonsmatt, line 900: there's the horribly named`NEL.fromList :: [a] -> NonEmpty a` unless by some act of grace it was removed

parsonsmatt: buuut if we just use list literal syyntax I think this can go away

parsonsmatt (approved):
looks great!

made a note that the version is going to be 2.11.1.1, which i'm planning on releasing today

persistent/ChangeLog.md
danbroooks: No probs I will shift this up :+1:

persistent/test/main.hs
```diff
@@ -223,22 +225,13 @@ main = hspec $ do
...
+                            (DocComment "this is a comment")
                         )
 
-            it "map parseLine" $ do
-                mapM parseLine ["Foo", "-- | Hello"]
-                    `shouldBe`
-                        Just
-                            [ Line 0 [Token "Foo"]
-                            , Line 0 [DocComment "Hello"]
-                            ]
```
danbroooks, line 237: The switch to using `OverloadedLists` caused this to give an ambiguous type error:

```         
    • Couldn't match expected type ‘GHC.Exts.Item (t0 Line)’
                  with actual type ‘Line’
      The type variable ‘t0’ is ambiguous
    • In the expression: Line 0 [Token "Foo"]
      In the first argument of ‘Just’, namely
        ‘[Line 0 [Token "Foo"], Line 0 [DocComment "Hello"]]’
      In the second argument of ‘shouldBe’, namely
        ‘Just [Line 0 [Token "Foo"], Line 0 [DocComment "Hello"]]’
    |       
236 |                             [ Line 0 [Token "Foo"]
    |                               ^^^^^^^^^^^^^^^^^^^^        
```

I've opted to actually just delete this test, this function is being tested already with the other tests here (the one above and below it), and also in the context of multiple lines, in tests like this one:

https://github.com/yesodweb/persistent/blob/813fe422e8144afe76ac83f7fb94f158c496bc22/persistent/test/main.hs#L538

parsonsmatt: Cool works for me!

parsonsmatt (approved):

danbroooks:
Hey I wasn't finished! :joy: sorry, maybe I shouldn't have pushed that stuff up. I will have another PR shortly, it should still fit in fine as a follow on I think :)

parsonsmatt:
Oh, shoot!

I've already released as 2.11.1.1 😅 

Try to mark a PR as a draft if it's not ready yet?

danbroooks:
Yeah its no big deal, what I meant was I wasn't done with the review updates, I've still yet to remove the `nonEmptyOrFail` function, its totally fine to be a part of that release :+1:

PR #1237, opened 2021-04-20
Remove nonEmptyOrFail function from recent tests

danbroooks:

After submitting your PR:

---

Removes the `nonEmptyOrFail` function in favour of making use of `OverloadedLists` after changes in https://github.com/yesodweb/persistent/pull/1231

### DIFF
--- a/persistent/test/main.hs
+++ b/persistent/test/main.hs
@@ -229,13 +229,14 @@ main = hspec $ do
             it "recognizes one line" $ do
                 parseLine "-- | this is a comment" `shouldBe`
                     Just
-                        ( Line 0 $ pure
-                            (DocComment "this is a comment")
+                        ( Line 0
+                            [ DocComment "this is a comment"
+                            ]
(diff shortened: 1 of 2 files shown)


persistent/test/main.hs
```diff
@@ -54,32 +54,36 @@ main = hspec $ do
                 `shouldBe`
                     ( [NEL.toList helloWorldTokens, NEL.toList foobarbazTokens], mempty )
         it "works4" $ do
-            let foobarbarz = ["foo", "Bar", "baz"]
```
danbroooks, line 57: These were also causing some type ambiguity errors, where with `OverloadedLists` we want to use this value here both as a `[Text]` and `NonEmpty Text`, so I have opted to re-writing the test with literals rather than trying to re-use the values. I've tried to include some more realworld looking values too while changing things here.

parsonsmatt (approved):
looks great thanks!

PR #1244, opened 2021-04-22
Implement config for customising the FK name

### LINKED ISSUE #1124: Foreign key constraint name is weird
Hello,

I am unsure if this is intentional - in which case this is a request for improvement - or not - in which case this is a bug report.

The foreign key constraint name is weird: it seems that the (mandatory) constraint name passed in the quasi quoter is appended with no separator (e.g. `_`) to the FK table name.

FYI I use:
* Stack with resolver `lts-16.5`, so Persistent `2.10.5.2` and Persistent Template `2.8.2.3`
* PostgreSQL `11.8`

To reproduce:

```haskell
share
  [mkPersist sqlSettings, mkMigrate "migrateAll"]
  [persistLowerCase|
  FooBar
      foo String
      bar Bool
      Primary foo bar
      deriving Show
  FooBarBaz
      foo String
      bar Bool
      baz Int
      Primary foo bar baz
      Foreign FooBar fk_something foo bar
      deriving Show
|]
```

The foreign key constraint is named `foo_bar_bazfk_something`, i.e. it appends the table name `foo_bar_baz` to the provided QQ name `fk_something` with no separator.

While I don't mind being forced to prefix with the table name `foo_bar_baz`, I don't understand why there isn't at least a separator `_` between the table name and the given suffix (so it would be `foo_bar_baz_fk_something`). And it's illegal (doesn't compile) to write `Foreign FooBar _fk_something foo bar`.

What do you think?

Cheers!

danbroooks:

After submitting your PR:

---

Closes #1124

Initially, I went with an approach that mirrors what is already done with `psToDBName` (a function `Text -> Text` rather than an optional `Maybe Text` value), though I thought this might not be preferred, and that it may make more sense to provide an optional infix as described on the issue.

I then thought it might make sense to create a newtype wrapper that is private to the module, which you can either go with the default (that just appends the two strings without any infix), or an infixing configuration, as illustrated in the test:

https://github.com/yesodweb/persistent/pull/1244/files#diff-bb8c841390595945f4d325e011ab6d4a0dc800bccaae6ea1ed348164dda7e7bfR362

This might be a bit overkill, so I am happy to change to how it is described in the issue if that is preferred.


danbroooks:
I've converted this back to draft for now, I think I will have a second go at it and investigate some other approaches. This might be better as a breaking change.

parsonsmatt:
Can you base new changes off of the `persistent-2.13` branch? I'm going to be working pretty heavily on refactoring and reorganizing code for that release and it'd be a lot easier if these changes were on top of that.

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -98,8 +99,16 @@ parseFieldType t0 =
...
+newtype ToFKName = ToFKName
+  { unToFKName :: Text -> Text -> Text
+  }
+
 data PersistSettings = PersistSettings
     { psToDBName :: !(Text -> Text)
+    , psToFKName :: !ToFKName
+    -- ^ Function used to generate the FK name. Default value: @mappend@
+    --
+    -- @since 2.12.1.2
```
parsonsmatt, line 111: This PR is currently against `persistent-2.13` - are you planning on backporting it?

Note that the `PersistSettings` is currently hidden, so we'll want to export a setter `setPsToFKName :: ToFKName -> PersistSettings -> PersistSettings`.

parsonsmatt (changes requested):
Looks great!

We need to expose a setter in the `Database.Persist.Quasi` module since the type and fields are only `Internal` details.

Also the `@since` declarations are for 2.12.1.2 - adding an export item is a minor bump, so backporting would be `persistent-2.12.2.0`. And this PR is against `persistent-2.13`, so it would come out with `persistent-2.13` as the initial release currently without a backport.

persistent/Database/Persist/Quasi/Internal.hs
danbroooks: This is just a hangover from my original change request, I will update these `@since` declarations :+1: looks like I need to rethink how this is switched on as well as you say, as in `2.13` `PersistSettings` is now in an `Internal` module, so I will add something like `setPsToFKName`

danbroooks: I've just pushed some more stuff up, so thats ready for a second look :+1: I will add some annotations to what I've done as there are some additional things in there to consider

persistent/Database/Persist/Quasi.hs
```diff
@@ -413,11 +413,14 @@ Unfortunately, we can't use this to create Haddocks for you, because <https://gi
 -}
 module Database.Persist.Quasi
     ( parse
-    , PersistSettings (..)
+    , PersistSettings
```
danbroooks, line 416: I noticed that the fields for `PersistSettings` were not actually being hidden, though from your comment I assume this is the desired behaviour in this version of persistent going forward, so I have snipped the fields off the export here.

persistent/test/main.hs
```diff
@@ -22,7 +22,18 @@ import Data.Time
...
+       ( Line(..)
+       , LinesWithComments(..)
+       , Token(..)
+       , associateLines
+       , parseFieldType
+       , parseLine
+       , preparse
+       , splitExtras
+       , takeColsEx
+       )
```
danbroooks, line 36: Likewise with this test, we are importing `Quasi.Internal`, but I assume, long term, we want to test the main module really without touching the `Internal` module (except other than perhaps the internal types, `Line`/`Token` and so on). I can revert this if you want though.

Having worked a little now on these modules/tests, I think it might be nice (especially after this `Internal` split on the quasi module), to have a test file for `Quasi` and a test file for `Quasi.Internal`, though I feel like a lot of these tests in here were added initially to cover as much of the module as possible because of the initial lack of coverage, it might be that a test file for `Quasi` to test the expected parsing behaviour would be sufficient. 

Just a thought I had while making this change :+1:

persistent/Database/Persist/Quasi.hs
```diff
@@ -413,11 +413,14 @@ Unfortunately, we can't use this to create Haddocks for you, because <https://gi
...
     , lowerCaseSettings
-    , toFKNameInfixed
+    , setPsUseSnakeCaseForiegnKeys
     , nullable
     ) where
 
 import Database.Persist.Quasi.Internal
+
+setPsUseSnakeCaseForiegnKeys :: PersistSettings -> PersistSettings
+setPsUseSnakeCaseForiegnKeys ps = ps { psToFKName = toFKNameInfixed "_" }
```
danbroooks, line 426: Bit of a long winded function name, but I went with this approach because I think people will want to either have this switched on or off. There is room for further flexibility though if this is wanted, but maybe that would be overkill.

I also thought with this change, that `persistent` should have this behaviour as default, potentially allowing people to configure this if they wanted to (though why they would I'm not sure), though unfortunately if we did that off the bat then that would break existing persistent codebases, right? Not sure what the long term fix is here, or it might have to be something that came along with a major version bump. Even along with a major version bump we are making people deal with renaming their foreign keys when there is actually technically wrong with them (they're just not aesthetically pleasing to the human reader).

Another approach with a major bump (or minor, I'm not sure which of the two this would fit in), would be to swap this functionality, so rather than _opt-in_ it would become _opt-out_, if for some reason people would rather just change the value in `PersistentSettings` than changing things in the DB.

Again, just thinking out loud here :)

persistent/Database/Persist/Quasi.hs
parsonsmatt: oh, dang, good catch 👍🏻

persistent/Database/Persist/Quasi.hs
parsonsmatt: IMO the default behavior should be consistent with whatever is currently happening - so end users wouldn't notice a change from this new option.

I'd prefer to see a more general setter exposed, even if this is the most likely variant that folks will want. People *can* just refer to the `.Internal` module but that's clunky.

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -98,8 +99,16 @@ parseFieldType t0 =
             PSDone -> PSSuccess (front []) t
             -- _ ->
 
+newtype ToFKName = ToFKName
```
parsonsmatt, line 102: As an exposed type, let's put a doc comment on it:

```suggestion
-- | A function for converting a table and column name into a constraint name.
-- 
-- @since 2.13.0.0
newtype ToFKName = ToFKName
```

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -98,8 +99,16 @@ parseFieldType t0 =
...
+  { unToFKName :: Text -> Text -> Text
+  }
```
parsonsmatt, line 104:
```suggestion
    { unToFKName :: Text -> Text -> Text
    }
```

persistent/Database/Persist/Quasi.hs
```diff
@@ -415,10 +413,14 @@ Unfortunately, we can't use this to create Haddocks for you, because <https://gi
 -}
 module Database.Persist.Quasi
     ( parse
-    , PersistSettings (..)
+    , PersistSettings
     , upperCaseSettings
     , lowerCaseSettings
+    , setPsUseSnakeCaseForiegnKeys
```
parsonsmatt, line 419: We should export the `FKName` type and constructor here too, for use with the setter.

persistent/test/main.hs
parsonsmatt: Yeah I'd eventually like to hollow this out into a more 'typical' `hspec` test suite, so we'd have `Database.Persist.QuasiSpec` etc that contains all this stuff.

parsonsmatt (changes requested):
Let's export a setter and the types for constructing the foreign key conversion functions.

It'd be nice to document the `Text` parameters to the function - `TableName` and `ColumnName` - somehow. But I'm not sure if you can attach Haddock to function arguments in a record field 🤔

persistent/Database/Persist/Quasi.hs
```diff
@@ -439,6 +440,21 @@ getPsToDBName = psToDBName
 setPsToDBName :: (Text -> Text) -> PersistSettings -> PersistSettings
 setPsToDBName f ps = ps { psToDBName = f }
 
+-- | Set a custom function used to create the constraint name
+-- for a foreign key.
+--
+-- @since 2.13.0.0
+setPsToFKName :: (EntityNameHS -> ConstraintNameHS -> Text) -> PersistSettings -> PersistSettings
+setPsToFKName setter ps = ps { psToFKName = setter }
```
parsonsmatt, line 448: Beautiful 👍🏻

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -102,6 +103,11 @@ parseFieldType t0 =
 data PersistSettings = PersistSettings
     { psToDBName :: !(Text -> Text)
     -- ^ Modify the Haskell-style name into a database-style name.
+    , psToFKName :: !(EntityNameHS -> ConstraintNameHS -> Text)
```
parsonsmatt, line 106: blessed, thanks! I thought it was a *table* name :joy: types FTW

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -226,14 +237,50 @@ lowestIndent = minimum . fmap lineIndent
...
+    fixForeignKeysAll . fmap (mkEntityDef ps . toParsedEntityDef) . associateLines
+
+data ParsedEntityDef = ParsedEntityDef
+    { parsedEntityDefComments :: [Text]
+    , parsedEntityDefEntityName :: EntityNameHS
+    , parsedEntityDefIsSum :: Bool
+    , parsedEntityDefEntityAttributes :: [Attr]
+    , parsedEntityDefFieldAttributes :: [[Token]]
+    , parsedEntityDefExtras :: M.Map Text [ExtraLine]
+    }
```
parsonsmatt, line 250: Oh this is *fantastic*

In #1256 I'm probably going to be only forwarding this type along, instead of the `EntityDef`. Maybe the `UnboundEntityDef`.

parsonsmatt (approved):
heck yeah! Thank you so much 👐🏻

PR #1252, opened 2021-04-26
Fix migrations

parsonsmatt:

After submitting your PR:

Migrations now do not do anything fancy. They call `mkEntityDefList` and then do an entirely ordinary `traverse_ (filter shouldmIgrate defs) (migrate defs)`

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -500,7 +500,7 @@ findTypeOfColumn allDefs name col =
         ((,) col)
         $ do
             entDef   <- find ((== name) . getEntityDBName) allDefs
-            fieldDef <- find ((== col)  . fieldDB) (getEntityFields entDef)
+            fieldDef <- find ((== col)  . fieldDB) (getEntityFieldsDatabase entDef)
             return (fieldType fieldDef)
 
 -- | Find out the maxlen of a column (default to 200)
@@ -509,7 +509,7 @@ findMaxLenOfColumn allDefs name col =
    maybe (col, 200)
          ((,) col) $ do
            entDef     <- find ((== name) . getEntityDBName) allDefs
-           fieldDef   <- find ((== col) . fieldDB) (getEntityFields entDef)
+           fieldDef   <- find ((== col) . fieldDB) (getEntityFieldsDatabase entDef)
            findMaxLenOfField fieldDef
 
 -- | Find out the maxlen of a field
@@ -1484,7 +1484,7 @@ mkBulkInsertQuery records fieldValues updates =
     firstField = case entityFieldNames of
         [] -> error "The entity you're trying to insert does not have any fields."
         (field:_) -> field
-    entityFieldNames = map fieldDbToText (getEntityFields entityDef')
+    entityFieldNames = map fieldDbToText (getEntityFieldsDatabase entityDef')
     tableName = T.pack . escapeE . getEntityDBName $ entityDef'
     copyUnlessValues = map snd fieldsToMaybeCopy
     recordValues = concatMap (map toPersistValue . toPersistFields) records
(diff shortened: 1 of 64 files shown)


persistent/Database/Persist/TH.hs
```diff
@@ -1185,19 +1200,21 @@ fieldError tableName fieldName err = mconcat
     ]
 
 mkEntity :: EntityMap -> MkPersistSettings -> EntityDef -> Q [Dec]
-mkEntity entityMap mps entDef = do
-    entityDefExp <-
-        if mpsGeneric mps
-           then liftAndFixKeys entityMap entDef
-           else makePersistEntityDefExp mps entityMap entDef
```
parsonsmatt, line 1192: I had to remove this optimization, since it wasn't handling `FieldDef` for `MigrationOnly` stuff.

persistent/Database/Persist/TH.hs
```diff
@@ -1185,19 +1200,21 @@ fieldError tableName fieldName err = mconcat
...
+        name = mkEntityDefName entDef
+        clazz = ConT ''PersistEntity `AppT` genDataType
+
     tpf <- mkToPersistFields mps entDef
     fpv <- mkFromPersistValues mps entDef
     utv <- mkUniqueToValues $ entityUniques entDef
     puk <- mkUniqueKeys entDef
     let primaryField = entityId entDef
-    fields <- mapM (mkField mps entDef) $ primaryField : getEntityFields entDef
+    fields <- mapM (mkField mps entDef) $ primaryField : getEntityFieldsDatabase entDef
```
parsonsmatt, line 1219: Since migrations are now handled by `mkEntityDefList`, we need to preserve *all* the field information.

persistent/Database/Persist/TH.hs
```diff
@@ -1743,67 +1762,56 @@ derivePersistFieldJSON s = do
...
-    return
-        [ SigD (mkName fun) typ
-        , FunD (mkName fun) [normalClause [] body']
+mkMigrate fun eds = do
+    let entityDefListName = ("entityDefListFor" <> fun)
+    body <- [| migrateModels $(varE (mkName entityDefListName)) |]
+    edList <- mkEntityDefList entityDefListName eds
+    pure $ edList <>
+        [ SigD (mkName fun) (ConT ''Migration)
+        , FunD (mkName fun) [normalClause [] body]
```
parsonsmatt, line 1815: This is the main change, as far as users are concerend

Instead of fixing the keys and faffing about with all that, we just use the result of `mkEntityDefList`, which contains the canonical `EntityDef` defined in `PersistEntity`.

persistent/Database/Persist/Types/Base.hs
```diff
@@ -297,24 +297,40 @@ data EmbedFieldDef = EmbedFieldDef
...
 toEmbedEntityDef ent = embDef
   where
     embDef = EmbedEntityDef
-      { embeddedHaskell = entityHaskell ent
-      , embeddedFields = map toEmbedFieldDef $ entityFields ent
-      }
+        { embeddedHaskell = entityHaskell ent
+        , embeddedFields =
+            map toEmbedFieldDef
+            $ filter isHaskellField
```
parsonsmatt, line 316: If we're embedding the entity, then we only care about the Haskell variants. We can't have a normal SQL schema, so it's inherently flexible/untyped.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -500,7 +500,7 @@ findTypeOfColumn allDefs name col =
         ((,) col)
         $ do
             entDef   <- find ((== name) . getEntityDBName) allDefs
-            fieldDef <- find ((== col)  . fieldDB) (getEntityFields entDef)
+            fieldDef <- find ((== col)  . fieldDB) (getEntityFieldsDatabase entDef)
```
parsonsmatt, line 503: Lots of this is in the backend modules. We need *all* the columns, not just the Haskell fields.

persistent/Database/Persist/TH.hs
```diff
@@ -475,7 +491,7 @@ mkPersist mps ents' = do
         , symbolToFieldInstances
         ]
   where
-    ents = map (fixEntityDef . setDefaultIdFields mps) ents'
+    ents = embedEntityDefs $ map (setDefaultIdFields mps) ents'
```
parsonsmatt, line 494: This fixes the embed definitions.

parsonsmatt (commented):
lorde help i Despise laziness right now

PR #1256, opened 2021-04-27
QuasiQuoter Improvements

### LINKED ISSUE #786: Runtime exception when foreign key is specified using `Key Table` instead of using TableId
I have a schema which looks something like this:

```
share
    [ mkPersist sqlSettings ]
    [persistLowerCase|
RoleTable
    name  RoleName
    deriving Generic Show

UserTable
    roleId        (Key RoleTable)
```

This results in a runtime exception:

```
PersistMarshalError "field roleId: int64 Expected Integer, received: PersistText \"1\""
```

If I use `RoleTableId` instead of `(Key RoleTable)` in `UserTable` everything works fine. This is perhaps not an expected usage of the DSL but the error was not caught. Perhaps something went wrong in the TH code generation from the schema and it compiled fine but resulted in a runtime error. If that's the case then the schema DSL can perhaps be made a bit more robust against unexpected usage.

### LINKED ISSUE #994: Not in scope error when nesting table per class inheritance
To encode sum types / subtyping I am attempting to use the table per class strategy, similar to the [shared primary keys strategy](https://www.parsonsmatt.org/2019/03/19/sum_types_in_sql.html).

However I am running into a very strange not in scope error when modeling nested inheritance:

The following works and seems to give me what I want (if a bit hacky):

```
Vehicle

Car
    vehicle VehicleId sql=id
    Primary vehicle

Bike
    vehicle VehicleId sql=id
    Primary vehicle
```

However it breaks when I append the following to the bottom:

```
Coupe
    car CarId sql=id
    Primary car

Sedan
    car CarId sql=id
    Primary car
```

With the following error:

```
    • Not in scope: type constructor or class ‘VehicleId’
    • In the quasi-quotation:
        [persistLowerCase|
```

This is with `persistent-2.9.2` and `persistent-template-2.6.0`.

### LINKED ISSUE #1042: Issues resolving the type of entity IDs
`persistent` tries *real hard* to make foreign keys work out, but it's not easy!

### LINKED ISSUE #1073: Foreign syntax doesn't recognize tables in other modules
Suppose you have the following modules:

```haskell
-- Schema/Foo.hs
module Schema.Foo where

mkPersist $(persistFileWith "foo.persistentmodels")

-- foo.persistentmodels
Foo
    name Text
    Primary name

-- Schema/Bar.hs

module Schema.Bar where

import Schema.Foo

mkPersist $(persistFileWith "bar.persistentmodels")

-- bar.persistentmodels
Bar
    foo Text
    Foreign Foo fk_bar_to_foo foo
```

This code will currently fail, because the `HaskellName "Foo"` is not present in the `EmbedEntityMap`, because it was not parsed from the same block of text.

I think there needs to be some way to lookup the entity *without* requiring it be present in the map. Attempting to call `entityDef (Nothing :: Maybe $(Foo))` instead of relying on what's in the `Map` might work.

### LINKED ISSUE #1149: Can't refer to `UserId` in an `Id` field
This QQ breaks with an error that `UserId` is not in scope.

```haskell
share [ mkPersist sqlSettings ] [persistLowerCase|

User
    name    String

Profile
    Id      UserId
    email   String
|]
```

However, using `UserId` is *usually* fine in other contexts. So why isn't this working?

### LINKED ISSUE #1241: Pass existing `[EntityDef]` into `mkPersist`
```haskell
mkPersist :: MkPersistSettings -> [EntityDef] -> Q [Dec]
```

A problem with this is that the generated declarations don't know about any of the `EntityDef`s that are not provided as part of this invocation. So if you have two blocks, one of which defines `User` and another defining an `Organization` with a `user UserId` field, the foreign key won't get properly invoked, since the `liftAndFIxKeys` machinery won't find the `User` in the existing map of tables.

This would fix #1073 as well. That issue has some good exploration on why this particular problem impacts foreign key generation. `persistent` figures that a `BlahBlahId` type is a reference to a `BlahBlah` model, and it tries to find the `BlahBlah` model in a `Map HaskellName EntityDef` that it constructs... from the `[EntityDef]` that is passed in.

A work-around is to... well, pass them in!

```haskell
module Foo where

    share [mkPersist sqlSettings, mkEntityDefList "fooDefs"] [persistLowerCase|

        Foo
            name Text
    |]

module Bar where

    import Foo

    share
        [ mkPersist sqlSettings . mappend fooDefs
        , mkEntityDefList "barAndFooDefs"
        ] [persistLowerCase|
         
            Bar
                foo FooId
                name Text
    |]
```

This requires that you do that `. mappend dependentDefs` for each one. But it should fix the issue.

It's a bit annoying to tease these relationships together manually. And it's *very* easy to forget, especially since, well, everyone pretty much has. Can we do better?

parsonsmatt:
This PR improves the QuasiQuoter. It no longer returns an `EntityDef`, instead it returns `UnboundEntityDef`. All of the compile-time type splicing responsibilities have been shifted over to `mkPersist`.

You can now refer to foreign keys defined outside of the module.

You can refer to entity IDs in an `Id` column (eg `Id UserId` now works).

- [x] Still need to fix the setting of foreign key references.

---

This PR also disables the MongoDB build. The reason is that breaking the cycle in the `EmbedEntityDef` is necessary, and is also quite difficult! It'd be nice to support `MongoDB`, but I'd rather just pause support for Mongo for now.  Given that Persistent gets 4470 downlaods in the last 30 days and Mongo only has 115, I'm OK with temporarily dropping support.

I've added upper bounds constraints so folks won't be able to upgrade when this is released.  

Consider this a call to action for additional maintainers for `persistent-mongoDB`!

---

After submitting your PR:

Fixes #1241 

Fixes #1073 

Fixes #994 

Fixes #1042 

FIxes #786

Fixes #1149


parsonsmatt:
Currently we're getting a test failure where all `foo   SomeTableId` columns are determined to be `SqlInt64` by the QuasiQuoter, and this isn't getting fixed in post.

I think a fix is to have an `UnboundFieldDef` on `UnboundEntityDef`, and these are spliced up nice and right in post. That should make the threading behavior more clear.

parsonsmatt (commented):
Massive change. Wish this were a smaller PR but I just kept smashing into problems and fixing things.

Very excited to get this in. Will make it a lot easier to work on the QQ and add features now that we're being more direct about what we can provide.

I need to write a blog post on this whole technique. It's pretty amazing. A two phase Template Haskell splicing setup is really quite brilliant. But I think it may be possible to get it to a single phase...

pbrisbin (commented):
I wish I had the domain knowledge to add a meaningful review here. The idea sounds great, and thanks for poking me for review. Maybe if you keep doing that, I will be able to add something meaningful some day :)

On the _not_ meaningful side, I did just notice "Ran stylish-haskell on any changed files." in the checklist. I have a side-project that actually automates that, if you're interested: https://restyled.io. If you want any more details or help setting it up, let me know. It's all open source and in Haskell too.

parsonsmatt:
@pbrisbin If you're still at Freckle, it'd be nice to know how these changes might affect y'all's codebase. I'll make an issue to get restyled working, that'd be great 😄

pbrisbin:
> If you're still at Freckle, it'd be nice to know how these changes might affect y'all's codebase

Oh yeah, totally appreciate that. Can you maybe link me to a specific part that shows the "user"-facing changes? Will we have to modify our existing entity definitions? I got the sense this would open up some new things, but not require any actual migration to them yet, but that could be wrong.

parsonsmatt:
Yeah, that *should* be the case. The biggest differences are:

1. If you're writing your own thing that uses the QuasiQuoter, then the changes to that will be breaking. eg `[peristLowerCase|...|]` no longer spliced in a `[EntityDef]`, it splices in a `[UnboundEntityDef]`.
2. If you're using embedded entities extensively, or in ways that aren't covered by the test suite, then this may break your code. I'm not terribly confident that the tests there cover all possible uses.

pbrisbin:
Cool, (2) won't impact us. For (1), all our definitions are of the form:

```hs
Teacher sql=teachers
  -- ...
|]
```

So as long as `mkPersist` accepts whatever `persistLowerCase` will now splice in, we are unaffected.

Issue #1166, opened 2020-11-27
model TH support of kinds (promoted types)


tim2CF:
Kinds (promoted types) are the "level-up" for phantom type parameters. For example instead of blind `a` and `b` phantom types

```haskell
newtype MoneyAmount a b = MoneyAmount Rational
```

We can build awesome compile-time safe and clear code like this

```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE KindSignatures #-}

data Owner = Merchant | Customer
data AccountKind = Debit | Credit

newtype MoneyAmount (a :: Owner) (b :: AccountKind) = MoneyAmount Rational
```

And then I want to use this newtype in my persistent models, for example

```
CustomerTransfer
  customerId CustomerId
  moneyAmount (MoneyAmount 'Customer 'Debit)
  currencyCode CurrencyCode
  uuid TransferUuid
```

But persistent TH code is failing with error like

```
/app/src/CoinsAgent/Data/Model.hs:25:3: error:
    * Exception when trying to run compile-time code:
        Invalid field type "MoneyAmount 'Customer" PSFail ('\'',"Customer")
CallStack (from HasCallStack):
  error, called at ./Database/Persist/Quasi.hs:382:36 in persistent-2.9.2-3XKdI42in9fIZrbGu
vFI1Z:Database.Persist.Quasi
      Code: persistFileWith lowerCaseSettings "config/model"
    * In the untyped splice:
        $(persistFileWith lowerCaseSettings "config/model")
   |
25 |   $(persistFileWith lowerCaseSettings "config/model")
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

parsonsmatt:
If you omit the ticks, it should work.

```
    moneyAmount (MonyAmount Customer Debit)
```

Can you try that and report back?

tim2CF:
Well, by default compiler don't like that. Of course it can be done by suppressing warnings or introducing type alias with `type` keyword. But promoted type syntax is part of the Haskell, why not support that? As I mentioned, there are possible workarounds, so this is not critical issue, just nice to have. 

```
error: [-Wunticked-promoted-constructors, -Werror=u
nticked-promoted-constructors]
    Unticked promoted constructor: `Kind.User'.
    Use 'Kind.User instead of `Kind.User'.
   |
32 | share
   | ^^^^^...
```

[...]

PR #1258, opened 2021-04-28
Support promoted types in Quasi Quoter

danbroooks:

After submitting your PR:

--------

Closes https://github.com/yesodweb/persistent/issues/1166

I've opened this as a draft PR with a view towards getting some feedback on it still. I think everything here should be fine for addressing that ticket, but there might be something I have overlooked here. I will annotate what I've got so far.

I have also opened this PR against `master` as it is not a super invasive change, and I would be happy to deal with porting it forward into the `persistent-2.13` branch after it is merged in. Alternatively if it is preferred that this PR goes against that branch, and is backported to `master`, I'd be happy to do that as well.

It may also make sense for this to just go into `2.13` like my other recent change :+1:

### DIFF
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -487,20 +487,28 @@ parseFieldType t0 =
                 | c == '(' -> parseEnclosed ')' id t'
                 | c == '[' -> parseEnclosed ']' FTList t'
                 | isUpper c || c == '\'' ->
-                    let (a, b) = T.break (\x -> isSpace x || x `elem` ("()[]"::String)) t
-                     in PSSuccess (getCon a) b
+                    let (a, b) = T.break (\x -> isSpace x || x `elem` ("()[]"::String)) t'
+                     in PSSuccess (parseFieldTypePiece c a) b
                 | otherwise -> PSFail $ show (c, t')
-    getCon t =
-        case T.breakOnEnd "." t of
-            (_, "") -> FTTypeCon Nothing t
-            ("", _) -> FTTypeCon Nothing t
-            (a, b) -> FTTypeCon (Just $ T.init a) b
+
     goMany front t =
         case parse1 t of
             PSSuccess x t' -> goMany (front . (x:)) t'
             PSFail err -> PSFail err
             PSDone -> PSSuccess (front []) t
             -- _ ->
+
+parseFieldTypePiece :: Char -> Text -> FieldType
+parseFieldTypePiece fstChar rest =
+    case fstChar of
+        '\'' ->
+            FTTypePromoted rest
+        _ ->
+            let t = T.cons fstChar rest
+             in case T.breakOnEnd "." t of
+                (_, "") -> FTTypeCon Nothing t
+                ("", _) -> FTTypeCon Nothing t
+                (a, b) -> FTTypeCon (Just $ T.init a) b
 
 data PersistSettings = PersistSettings
     { psToDBName :: !(Text -> Text)
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -486,7 +486,7 @@ parseFieldType t0 =
                 | isSpace c -> parse1 $ T.dropWhile isSpace t'
                 | c == '(' -> parseEnclosed ')' id t'
                 | c == '[' -> parseEnclosed ']' FTList t'
-                | isUpper c ->
+                | isUpper c || c == '\'' ->
                     let (a, b) = T.break (\x -> isSpace x || x `elem` ("()[]"::String)) t
                      in PSSuccess (getCon a) b
                 | otherwise -> PSFail $ show (c, t')
--- a/persistent/Database/Persist/TH.hs
+++ b/persistent/Database/Persist/TH.hs
@@ -350,7 +350,7 @@ mEmbedded _ (FTTypeCon Just{} _) =
 mEmbedded ents (FTTypeCon Nothing (EntityNameHS -> name)) =
     maybe (Left Nothing) Right $ M.lookup name ents
 mEmbedded ents (FTTypePromoted (EntityNameHS -> name)) =
-    maybe (Left Nothing) Right $ M.lookup name ents
+    Left Nothing
 mEmbedded ents (FTList x) =
     mEmbedded ents x
 mEmbedded ents (FTApp x y) =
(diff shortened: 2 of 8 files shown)


persistent/test/Database/Persist/TH/KindEntitiesSpec.hs
```diff
@@ -0,0 +1,43 @@
...
+|]
+
+spec :: Spec
+spec = describe "KindEntities" $ do
+    it "works" $ do
+        {-
+        let UserName = #name
+            OrganizationName = #name
+            DogName = #name
+-}
```
danbroooks, line 39: There might be some more stuff to do here for this test maybe ... for now I have something that just asserts that it typechecks :+1:

persistent/Database/Persist/TH.hs
```diff
@@ -349,6 +349,8 @@ mEmbedded _ (FTTypeCon Just{} _) =
     Left Nothing
 mEmbedded ents (FTTypeCon Nothing (EntityNameHS -> name)) =
     maybe (Left Nothing) Right $ M.lookup name ents
+mEmbedded ents (FTTypePromoted (EntityNameHS -> name)) =
+    maybe (Left Nothing) Right $ M.lookup name ents
```
danbroooks, line 353: Not sure about this. I basically just mirrored what is done for `FTTypeCon` when there is no module (ie the case above).

persistent/Database/Persist/Quasi.hs
```diff
@@ -486,22 +486,30 @@ parseFieldType t0 =
...
+parseFieldTypePiece fstChar rest =
+    case fstChar of
+        '\'' ->
+            FTTypePromoted rest
+        _ ->
+            let t = T.cons fstChar rest
+             in case T.breakOnEnd "." t of
+                (_, "") -> FTTypeCon Nothing t
+                ("", _) -> FTTypeCon Nothing t
+                (a, b) -> FTTypeCon (Just $ T.init a) b
```
parsonsmatt, line 511: Looks great 👍🏻

persistent/Database/Persist/TH.hs
parsonsmatt: Hmm - I'd say let's go with a `Left Nothing` on all cases here. If we hit this, it means we're trying to determine whether we need an embedded entity reference, which we shouldn't - the only place this could *possibly* be is if we've got an `FTApp` or similar, and we want the SQL type construction to proceed unchanged.

```suggestion
    Left Nothing
```

Like, we'll have `foo   (Proxy 'User)`. That'll get turned into `FTApp (FTypeCon Nothing "Proxy") (FTTypePromoted "User")`. This code recurses and then finds that the corresponding `EmbedEntityDef` is for a `User`, even though that's totally wrong.

persistent/Database/Persist/TH.hs
```diff
@@ -1758,13 +1760,21 @@ maybeNullable :: FieldDef -> Bool
...
+    FTTypeCon (Just "Data.Int") "Int64" ->
+        ConT ''Int64
+    FTTypeCon (Just m) t ->
+        ConT $ mkName $ unpack $ concat [m, ".", t]
+    FTTypePromoted t ->
+        PromotedT $ mkName $ T.unpack t
+    FTApp x y ->
+        ftToType x `AppT` ftToType y
+    FTList x ->
+        ListT `AppT` ftToType x
```
parsonsmatt, line 1777: 👨🏻‍🍳

persistent/test/Database/Persist/TH/KindEntitiesSpec.hs
```diff
@@ -0,0 +1,43 @@
...
+-- {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedLabels #-}
+-- {-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TemplateHaskell #-}
+-- {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Database.Persist.TH.KindEntitiesSpec where
```
parsonsmatt, line 16: beautiful tests <3

persistent/test/Database/Persist/TH/KindEntitiesSpec.hs
parsonsmatt: Yeah - let's check the type of the field accessors and the `EntityField`s.

Still should jsut be a compilation check, but something like:

```haskell
let mkTransfer :: CustomerId -> MoneyAmount 'CustomerOwned 'Debt -> CustomerTransfer
    mkTransfer = CustomerTransfer
    getAmount :: CustomerTransfer -> MoneyAmount 'CustomerOwned 'Debt
    getAmount = customerTransferMoneyAmount
```

persistent/test/Database/Persist/TH/KindEntitiesSpecImports.hs
```diff
@@ -0,0 +1,21 @@
...
+data Owner = MerchantOwned | CustomerOwned
+data AccountKind = Debit | Credit
+
+newtype MoneyAmount (a :: Owner) (b :: AccountKind) = MoneyAmount Rational
+
+instance PersistFieldSql (MoneyAmount a b) where
+    sqlType _ = SqlInt64
+
+instance PersistField (MoneyAmount a b) where
+    toPersistValue (MoneyAmount n) = PersistRational n
```
parsonsmatt, line 18: I don't think this `PeristRational` lines up with the `SqlInt64` type you specified

parsonsmatt (changes requested):
There are a few changes I'd like on this before accepting it, but it is fantastic work - thank you so much!

I'm afraid the changes I put into #1256  are going to be a massive conflict here. So let's get this merged in to `master`, we can release it as a patch bump since it's just a change to the QQ and purely additive. Then I'll work on getting it merged into `persistent-2.13`.

parsonsmatt (approved):
🎉

parsonsmatt:
All good to merge?

danbroooks:
Yeah all good :+1: I'm happy to do the merge/conflict resolution into `persistent-2.13` if you want (I can do it straight after this PR is merged if it is going in now), I'm also happy to leave it to you :smile:

parsonsmatt:
That would be really helpful, thank you 🙌🏻

Issue #277, opened 2014-08-30
support Maybe before the type


gregwebs:
Rather than

```
  Person
      name    Text Maybe
```

prefer

```
  Person
      name    (Maybe Text)
```

snoyberg:
Closing out old issues, please reopen if still relevant

gregwebs:
I would like to add this as an enhancement since it seems to be a point of a confusion for every persistent user.

parsonsmatt:
This one is actually pretty easy.

We need an instance of `PersistFieldSql (Maybe a)` and then we need to change the `mkColumns` function to check if `Maybe` is the first type constructor.

danbroooks:
@parsonsmatt can I assign myself to this? I can pick this up once the dust has settled with your current changes to the TH/Quasi modules. I was pondering how you would go about doing this today funnily enough!

parsonsmatt:
yeah definitely!

In `Database.Persist.Sql.Internal`,. we have:

```haskell
mkColumns = -- snip ...
    go :: FieldDef -> Column
    go fd =
        Column
            { cName = fieldDB fd
            , cNull = nullable (fieldAttrs fd) /= NotNullable || isEntitySum t
            , cSqlType = fieldSqlType fd
            , cDefault = defaultAttribute $ fieldAttrs fd
            , cGenerated = fieldGenerated fd
            , cDefaultConstraintName =  Nothing
            , cMaxLen = maxLen $ fieldAttrs fd
            , cReference = mkColumnReference fd
            }
```

We need to ensure that `cNull` is set for a column 

Right now, it just sees if `Maybe` is present in the `FieldAttr`s. But we could also look at the `fieldType` -

```haskell
isMaybe :: FieldType -> Bool
isMaybe ft =
    case ft of
        FTApp (FTTypeCon _ "Maybe") _ ->
            True
        _ ->
            False
```

PR #1264, opened 2021-05-06
Support declaring Maybe before the type in model definitions

danbroooks:

After submitting your PR:

Closes https://github.com/yesodweb/persistent/issues/277

### DIFF
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -366,6 +366,36 @@ Notification
     Foreign User fk_noti_user sentToFirst sentToSecond References emailFirst emailSecond
 @
 
+= Nullable Fields
+
+As illustrated in the example at the beginning of this page, we are able to represent nullable
+fields by including 'Maybe' at the end of the type declaration:
+
+> TableName
+>     fieldName      FieldType
+>     otherField     String
+>     nullableField  Int       Maybe
+
+Alternatively we can specify the keyword nullable:
+
+> TableName
+>     fieldName      FieldType
+>     otherField     String
+>     nullableField  Int       nullable
+
+However the difference here is in the first instance the Haskell type will be 'Maybe Int',
+but in the second it will be 'Int'.
+
+If you wish to define your Maybe types in a way that is similar to the actual Haskell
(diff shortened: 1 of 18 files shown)


persistent/Database/Persist/Quasi.hs
```diff
@@ -366,6 +366,36 @@ Notification
...
+
+Alternatively we can specify the keyword nullable:
+
+> TableName
+>     fieldName      FieldType
+>     otherField     String
+>     nullableField  Int       nullable
+
+However the difference here is in the first instance the Haskell type will be 'Maybe Int',
+but in the second it will be 'Int'.
```
parsonsmatt, line 387: Should add a note that this will cause runtime errors if the database returns `NULL` and the `PersistField` instance for the type in question doesn't handle `PersistNull`.

persistent/Database/Persist/Quasi.hs
```diff
@@ -366,6 +366,36 @@ Notification
...
+
+If you wish to define your Maybe types in a way that is similar to the actual Haskell
+definition, you can define 'Maybe Int' like so:
+
+> TableName
+>     fieldName      FieldType
+>     otherField     String
+>     nullableField  (Maybe Int)
```
parsonsmatt, line 395: awesome

persistent/test/Database/Persist/THSpec.hs
```diff
@@ -176,6 +178,7 @@ spec = describe "THSpec" $ do
     SharedPrimaryKeySpec.spec
     SharedPrimaryKeyImportedSpec.spec
     ImplicitIdColSpec.spec
+    MaybeFieldDefsSpec.spec
```
parsonsmatt, line 181: okay awesome

This is testing that the TH generation does the right thing for the Haskell side

parsonsmatt (changes requested):
Let's write a test in `persistent-test` that'll be run on the various backends and ensure that migrations are getting handled correctly.

I expect that they are, but it's good to double check :)

parsonsmatt:
CI is fixed on master

parsonsmatt (approved):

PR #1277, opened 2021-05-26
Improve documentation around migration safety

### LINKED ISSUE #1081: The documentation for addMigration is confusing
Let me refer to the code:

```
-- | Add a migration to the migration plan.
--
-- @since 2.9.2
addMigration
    :: Bool
    -- ^ Is the migration safe to run? (eg a non-destructive and idempotent
    -- update on the schema)
    -> Sql
    -- ^ A 'Text' value representing the command to run on the database.
    -> Migration
addMigration isSafe sql = lift (tell [(isSafe, sql)])
```

What I understand from this is that `addMigration True "test"` would add a safe migration.

But if we look at the code for `safeSql`:

```
safeSql :: CautiousMigration -> [Sql]
safeSql = allSql . filter (not . fst)
```

It would filter out that migration because it takes the opposite condition.

ivanbakel:
The Haddocks around CautiousMigration were ambiguous, since it wasn't clear if the flag signified whether each migration was safe or not - and there is at least one instance in the code that gets it wrong.

This corrects that error, opting to make the `addMigration` documentation match the code's behaviour.

Personally, I think the better approach to this mess would be to replace all the use of `(Bool, Text)` with a datatype that enforces better usage through a stricter API - something like
```haskell
data CautiousSql = MkCautiousSql { ... } -- Hide this constructor

-- Expose these smart constructors
mkSafeSql :: Sql -> CautiousSql
mkSafeSql = ...

mkUnsafeSql :: Sql -> CautiousSql
mkUnsafeSql = ...

-- Expose these instead of record fields
isUnsafe :: CautiousSql -> Bool
isUnsafe = ...

rawSql :: CautiousSql -> Sql
rawSql = ...

-- Maybe even a smart destructor?
ifSafe :: (Bool -> Sql -> a) -> CautiousSql -> a
ifSafe = ...
```
That way consumers could not be misled about the nature of the `Bool` value. But I thought that would be quite drastic and breaking, so I didn't want to make it my first suggestion.

Fixes #1081.

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Sql/Migration.hs
+++ b/persistent/Database/Persist/Sql/Migration.hs
@@ -22,20 +22,20 @@ module Database.Persist.Sql.Migration
 import Control.Exception (throwIO)
 import Control.Monad (liftM, unless)
 import Control.Monad.IO.Unlift
-import Control.Monad.Trans.Class (MonadTrans (..))
-import Control.Monad.Trans.Reader (ReaderT (..), ask)
+import Control.Monad.Trans.Class (MonadTrans(..))
+import Control.Monad.Trans.Reader (ReaderT(..), ask)
 import Control.Monad.Trans.Writer
-import Data.Text (Text, unpack, snoc, isPrefixOf, pack)
+import Data.Text (Text, isPrefixOf, pack, snoc, unpack)
 import qualified Data.Text.IO
+import GHC.Stack
 import System.IO
 import System.IO.Silently (hSilence)
-import GHC.Stack
 
+import Database.Persist.Sql.Orphan.PersistStore ()
+import Database.Persist.Sql.Raw
 import Database.Persist.Sql.Types
 import Database.Persist.Sql.Types.Internal
-import Database.Persist.Sql.Raw
 import Database.Persist.Types
-import Database.Persist.Sql.Orphan.PersistStore()
 
 allSql :: CautiousMigration -> [Sql]
 allSql = map snd
--- a/persistent/Database/Persist/Sql/Migration.hs
+++ b/persistent/Database/Persist/Sql/Migration.hs
@@ -195,12 +195,14 @@ reportErrors = tell
 -- @since 2.9.2
 addMigration
     :: Bool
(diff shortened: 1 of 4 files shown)


ivanbakel:
I've discovered this fixes an issue, so I'll add that into the PR message.

persistent/Database/Persist/Sql/Migration.hs
```diff
@@ -200,7 +200,7 @@ addMigration
     -> Sql
     -- ^ A 'Text' value representing the command to run on the database.
     -> Migration
-addMigration isSafe sql = lift (tell [(isSafe, sql)])
+addMigration isSafe sql = lift (tell [(not isSafe, sql)])
```
parsonsmatt, line 203: This would be a breaking change in behavior. Let's instead change the documentation? That way, we won't break any existing code, and folks writing new code will be able to tell from the docs how to do it right.

persistent/Database/Persist/Sql/Types.hs
```diff
@@ -63,7 +63,10 @@ type SqlPersistM = SqlPersistT (NoLoggingT (ResourceT IO))
 
 type Sql = Text
 
--- Bool indicates if the Sql is safe
+-- A list of SQL operations, marked with a safety flag. If the 'Bool' is
```
parsonsmatt, line 66:
```suggestion
-- | A list of SQL operations, marked with a safety flag. If the 'Bool' is
```

This would make it show up in Haddocks. 😄

parsonsmatt (changes requested):
Let's make this a documentation change instead of a behavior change. That way we can release it as a patch fix. If we're changing behavior, that'd need to be in a major version bump.

persistent/Database/Persist/Sql/Migration.hs
ivanbakel: This is done now.

parsonsmatt:
looks great! `persistent-2.13.0.3` was released since, but I'll get this out as `persistent-2.13.0.4`.

parsonsmatt (approved):

PR #1280, opened 2021-06-08
Derive persistent definitions from a data type

lykahb:
A declaration of a model with Persistent combines both the creation of the data type and the Persistent instances. If we could write a regular Haskell data type and separately define a persistent instance for it, they would be loosely coupled. The declarations would have less quasi-quotation magic but as a downside, they may be more verbose.

This is a draft PR. Before the final cleanup and adding the documentation, it would be helpful like to get a high-level review of the API and its integration with Persistent.

Here is an example:

```haskell
data DerivePrimary = DerivePrimary {
  derivePrimaryName :: String,
  derivePrimaryAge :: Int
} deriving (Eq, Show)

derivePersist persistSettings {mpsGeneric=False, mpsRecordFieldToHaskellName = stripEntityNamePrefix} lowerCaseSettings [
  (mkDeriveEntityDef 'DerivePrimary) {
    deriveFields = [(mkDeriveFieldDef 'derivePrimaryName) {
      sqlNameOverride="sql_name"
      }]
  }]
```

### DIFF
--- a/persistent-test/src/DerivePersistTest.hs
+++ b/persistent-test/src/DerivePersistTest.hs
@@ -17,7 +17,7 @@ data DerivePrimary = DerivePrimary {
 derivePersist persistSettings {mpsGeneric=False, mpsRecordFieldToHaskellName = stripEntityNamePrefix} lowerCaseSettings [
   (mkDeriveEntityDef 'DerivePrimary) {
     deriveFields = [(mkDeriveFieldDef 'derivePrimaryName) {
-      sqlNameOverride=Just "sql_name"
+      sqlNameOverride="sql_name"
       }]
   }]
 
--- a/persistent-test/src/DerivePersistTest.hs
+++ b/persistent-test/src/DerivePersistTest.hs
@@ -2,7 +2,8 @@
 
 module DerivePersistTest where
 
-import Database.Persist.TH (mkDeleteCascade, derivePersist, persistLowerCase, stripEntityNamePrefix, DeriveEntityDef(..), DeriveFieldDef(..), DeriveForeignKey(..), DeriveUniqueDef(..), mkDeriveEntityDef, mkDeriveFieldDef, mkDeriveUniqueDef)
+import Database.Persist.TH (mkDeleteCascade, persistLowerCase)
+import Database.Persist.DerivePersist
 import Database.Persist.Quasi (lowerCaseSettings)
 import Data.Maybe (isJust)
 
@@ -129,9 +130,7 @@ specsWith runDb = describe "derivePersist" $ do
     k3 <- insert m3
     m3' <- get k3
     m3' @== Just m3
-    res <- rawSql "SELECT * FROM menu_object" []
-    liftIO $ print (res :: [(Single Text, Single Text)])
-  
+
   it "Supports composite keys" $ runDb $ do
     -- copy from CompositeTest
     let p1 = TestParent2 "a1" "b1" 11 "p1"
--- a/persistent-test/src/DerivePersistTest.hs
+++ b/persistent-test/src/DerivePersistTest.hs
@@ -2,7 +2,7 @@
 
 module DerivePersistTest where
 
-import Database.Persist.TH (mkDeleteCascade, derivePersist, persistLowerCase, stripEntityNamePrefix, DeriveEntityDef(..), DeriveFieldDef(..), DeriveForeignKey(..), mkDeriveEntityDef, mkDeriveFieldDef)
+import Database.Persist.TH (mkDeleteCascade, derivePersist, persistLowerCase, stripEntityNamePrefix, DeriveEntityDef(..), DeriveFieldDef(..), DeriveForeignKey(..), DeriveUniqueDef(..), mkDeriveEntityDef, mkDeriveFieldDef, mkDeriveUniqueDef)
 import Database.Persist.Quasi (lowerCaseSettings)
 import Data.Maybe (isJust)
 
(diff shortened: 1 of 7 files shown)


persistent-test/src/DerivePersistTest.hs
```diff
@@ -0,0 +1,156 @@
...
+
+module DerivePersistTest where
+
+import Database.Persist.TH (mkDeleteCascade, derivePersist, persistLowerCase, stripEntityNamePrefix, DeriveEntityDef(..), DeriveFieldDef(..), DeriveForeignKey(..), DeriveUniqueDef(..), mkDeriveEntityDef, mkDeriveFieldDef, mkDeriveUniqueDef)
+import Database.Persist.Quasi (lowerCaseSettings)
+import Data.Maybe (isJust)
+
+import Init
+
+data DerivePrimary = DerivePrimary {
```
lykahb, line 11: This module has copies of several tests, covering composite keys, uniques, and the overrides for instance definitions. It would be great to run make whole test suite work with both quasi-quotation and derivePersist, but it's not clear how to automate this.

persistent/Database/Persist/TH.hs
```diff
@@ -273,6 +287,211 @@ preprocessUnboundDefs preexistingEntities unboundDefs =
     (embedEntityMap, noCycleEnts) =
         embedEntityDefsMap preexistingEntities unboundDefs
 
+-- | Include entities that are related to each other in one call
+derivePersist :: MkPersistSettings -> PersistSettings -> [DeriveEntityDef] -> Q [Dec]
```
lykahb, line 291: Before the version 2.13 taking a list of entities was mandatory if they were referencing each other. Now it is just a design choice for API. It may be worth to merge this with `mkDeriveEntityDef` but it's not clear whether it's more ergonomic.

persistent/persistent.cabal
```diff
@@ -37,6 +37,7 @@ library
       , template-haskell         >= 2.13 && < 2.18
       , text                     >= 1.2
       , th-lift-instances        >= 0.1.14    && < 0.2
+      , th-abstraction           >= 0.2.8
```
lykahb, line 40: It is difficult to maintain code that introspects `Dec` for multiple versions of template-haskell. The fields often get added or moved. This offers a stable abstraction and an alternative to CPP pragmas.

persistent/Database/Persist/TH.hs
```diff
@@ -929,6 +1153,12 @@ data MkPersistSettings = MkPersistSettings
     -- ^ TODO: document
     --
     -- @since 2.13.0.0
+    , mpsCreateDataType :: !Bool
+    -- ^ Create data type for the entity
+    , mpsFieldSqlType :: FieldDef -> SqlTypeExp
+    , mpsRecordFieldToHaskellName :: Text -> Text -> Text
+    -- ^ Derive the Haskell names using the entity and field name. The result
+    -- should be a valid haskell indentifier (start with an lower cased letter).
```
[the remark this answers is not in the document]
parsonsmatt: I don't think I understand the meaning of this option from the comments.

The default doesn't alter it at all, and the `derivePersist` stuff uses `stripEntityNamePrefix`. The intent is to keep the `FieldNameHS` behavior consistent, so that you don't get the entity name prefix on a derived record.

My guess is that the intent here is to allow folks to use eg `DuplicateRecordFields` or similar and *not* have the type name prefix.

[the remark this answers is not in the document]
parsonsmatt: OK, since this flag is only used by the `derivePersist` functions, I don't think we need it - it can be accepted by `derivePersist` or in a `DerivePersistSettings` record.

parsonsmatt (changes requested):
I think this might be best as another library, and I think I'd rather export terms (possibly from a `Database.Persist.TH.Internal` module) that support that use rather than supporting it directly.

As far as I can tell, the only modification to `persistent` this needs is `mpsCreateDataType`. However, if you're doing `derivePersist` that calls `mkPersist` under the hood, you'd be able to filter out the data declaration from the resulting `[Dec]`.

eg:

```haskell
isDatatypeDeclarationForEntity :: EntityNameHS -> Dec -> Bool
isDatatypeDeclarationForEntity (EntityNameHS txt) dec =
    case dec of
        DataD _ name _ _ _ _ ->
            Text.unpack txt == nameBase name
        _ ->
            False
```

This would allow you (or anyone else) to make these libraries with Hackage versions of `persistent` and offer them up as alternatives.

I don't think I want to incorporate it into `persistent` itself. I imagine that the API for the `derivePersist` stuff will want to change quite a bit while it's being stabilized, and it'd be nice to not need to do a major bump of `persistent` for changes and alterations to alternative syntaxes.

Let's turn this into a separate library, and turn this PR into an "Expose everything we need from `Database.Persist.TH.Internal` to make this work." That way we can land this new means of deriving entities without locking anything in. I'd also be happy to link to it in the documentation.

persistent-test/src/DerivePersistTest.hs
```diff
@@ -0,0 +1,156 @@
...
+
+share [mkPersist sqlSettings { mpsGeneric = True }] [persistLowerCase|
+SubType3
+  object3 [MenuObject3]
+  deriving Show Eq
+
+MenuObject3
+  sub3 SubType3 Maybe
+  deriving Show Eq
+|]
```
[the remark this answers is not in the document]
lykahb: This remained from debugging the difference in implementation.

persistent-test/src/DerivePersistTest.hs
lykahb: Creation of `EntityDef` is difficult to extract from the rest of `mkPersistWith`. Comparing `UnboundEntityDef`s created by `parse` and `datatypeToEntityDef` would work though. That would require two definitions, but that's more robust than the current DerivePersistTest.

lykahb:
Thanks for the review. I also think that the API is going to have many breaking changes. At first I hoped to find a way to reuse the existing QQ definitions in the test suite for testing with diverse entity configurations, but that didn't work out. So a standalone library may be the best choice for now.

My only concern is that a library would have a difficulty with putting bounds for the Persistent dependency. Breaking changes in the internals do not require a major or even minor version bump. Setting dependency to a particular version is too restrictive for a user, and a version range may result in breakage. What is your opinion about `Database.Persist.TH.ExperimentalDerive`? It sets a different expectation for stability than the rest of Persistent.

PR #1281, opened 2021-06-09
Add put function

aschmois:

- [X] Adhered to the code style (see the `.editorconfig` file for details)

After submitting your PR:

Saw a weird behavior when wanting to do a simple `repsert` but also get back the key. I wrote up a draft to see what people think. I tried to copy behavior as much as possible but I don't know the codebase enough to know if this is the right move or not. 

Other possible solutions I had thought of:
- make repsert return the key (may not always be possible)
- make upsert return the key
- make a different upsert/repsert within the non-unique namespace (not sure why they're split up)

Important note: There are multiple different ways to implement this to be performant on the sql/mongo/etc. side, is this something that I should attempt? I'm not a big expert on all of the different database structures.


aschmois:
Don't have the bandwidth to complete this, I think I can come back to this at some other time.

Issue #1204, opened 2021-03-15
Is anyone actually using mpsGeneric ?


parsonsmatt:
The [`mpsGeneric`](https://www.stackage.org/haddock/lts-17.6/persistent-template-2.9.1.0/Database-Persist-TH.html#v:mpsGeneric) flag complicates `persistent-template` considerably and has a dubious value proposition - it allows you to reuse a model in a SQL context and a Redis context and a MongoDB context.

This doesn't seem very useful. Removing it would be great. But I don't want to break people who need this for some reason.

Unfortunately, I can't just, like, collect data on this. What I *can* do, however, is deprecate the field, and link to this issue. Folks that upgrade will see the deprecation notice and be able to common on whether they use it or not.

parsonsmatt:
I'm going to throw a deprecation notice on `mpsGeneric` for 2.13 with a link to this PR and a request for comment. If I don't get any bites then by 2.14 let's remove it.

parsonsmatt:
[Stackage LTS 18.0 has `persistent-2.13.0.2`](https://www.stackage.org/lts-18.0/package/persistent-2.13.0.2), and was released on [June 16th, 2021](https://www.stackage.org/lts-18.0). Six months of lead time means that I can delete `mpsGeneric` in December.

Going to set a calendar reminder to do that...

parsonsmatt:
A problem with the above approach is that `Key backend` becomes impossible to introspect on - you can't do *anything* with it aside from using it as an abstract database operation.

Right now, with `Key (UserGeneric backend)`, I can still call` unUserKey :: Key (UserGeneric backend) -> BackendKey backend` and `UserKey :: BackendKey backend -> Key (UserGeneric backend)`. With `Key rec`, I can't do anything at all beyond use it directly.

In a non-default ID case, like:

```
User
    Id     UUID
    name Text
    age Int
```

then the generic *does* let me introspect on it: 

```
UserKey :: UUID -> Key (UserGeneric backend)
unUserKey :: Key (UserGeneric backend) -> UUID
```

But the type variables *don't* give me the ability to write these. I'd need to include that stuff in the class.

parsonsmatt:
Something that's striking me as a somewhat obvious fix here is to have, like:

```haskell
module Persistent.Test.Models where

    models :: String
    models = [str|
    
    |]

module Persistent.Test.Sql where

    import Persistent.Test.Models

    mkPersist sqlSettings $(quoteExp persistLowerCase models)

[...]

PR #1287, opened 2021-06-17
Fix duplicate entity checks

### LINKED ISSUE #1288: The duplicate entity check doesn't work transitively
So the fix in https://github.com/yesodweb/persistent/pull/1255 isn't quite enough. The tests in #1287 document the issue.

With the following format:

```haskell

module A where
  
    A
      name Text
    |]

module B where

  import A
  mkPersistWith sqlSettings $(discoverEntities) [persistLowerCase|
    B
      a AId
  |]

module C where

  import B

  mkPersistWith sqlSettings $(discoverEntities) [persistLowerCase|
    C
      b BId
    |]
```

You'll get duplicate name errors because the `mkPersistWith` in `C` is trying to make models for `A`. `A` is not available for `lookupTypeName`, but it is still present in the input entity list.

Seems like it should be straightforward to just *not* generate things from the input list??

parsonsmatt:

After submitting your PR:

Closes #1288

### DIFF
--- a/persistent/Database/Persist/TH.hs
+++ b/persistent/Database/Persist/TH.hs
@@ -2758,7 +2758,7 @@ mkSymbolToFieldInstances mps entityMap (fixEntityDef -> ed) = do
             lowerFirstIfId xs = xs
 
             fieldTypeT
-                | fieldHaskellName == FieldNameHS "Id" = do
+                | fieldHaskellName == FieldNameHS "Id" =
                     conT ''Key `appT` recordNameT
                 | otherwise =
                     pure $ maybeIdType mps entityMap fieldDef Nothing Nothing
--- a/persistent/Database/Persist/TH.hs
+++ b/persistent/Database/Persist/TH.hs
@@ -771,10 +771,12 @@ mkPersistWith mps preexistingEntities ents' = do
             $ predefs
         entityMap =
             constructEntityMap allEnts
+        preexistingSet =
+            Set.fromList $ map getEntityHaskellName preexistingEntities
         newEnts =
-            Set.toList $
-                Set.fromList allEnts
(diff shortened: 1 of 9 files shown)


persistent/Database/Persist/TH.hs
```diff
@@ -773,7 +771,11 @@ mkPersistWith mps preexistingEntities ents' = do
             $ predefs
         entityMap =
             constructEntityMap allEnts
-    ents <- filterM shouldGenerateCode allEnts
+        newEnts =
+            Set.toList $
+                Set.fromList allEnts
+                Set.\\ Set.fromList (map unbindEntityDef preexistingEntities)
```
parsonsmatt, line 777: Lol, sigh, alas, oh well, so silly,,,

parsonsmatt:
mm no this is broken still

looks like the code in the `symbolToField` stuff needs to do a qualified lookup

persistent/test/Database/Persist/TH/SharedPrimaryKeyImportedSpec.hs
```diff
@@ -28,7 +28,7 @@ import Database.Persist.TH.SharedPrimaryKeySpec (User, UserId)
 
 mkPersistWith sqlSettings $(discoverEntities) [persistLowerCase|
 
-Profile
+ProfileX
```
parsonsmatt, line 31: `persistent` now fails to generate code for `Profile` because `Profile` is already a table defined in `discoverEntities`.

parsonsmatt (commented):
OK, I've got it fixed. Mostly by *skipping* the transformation from `ModelNameId` to `Key ModelName` if we're not in the `mpsGeneric` mode.

Another bug from #1204 :\

PR #1291, opened 2021-06-27
Use TemplateHaskell for RawSql instances

mitchellvitez:
#1087 mentions possibly generating `RawSql` instances with TemplateHaskell instead of repeated code; this PR does that.

This is largely just to see what it would look like, so I won't be offended if people think the way things are now is better

Some questions, especially since I haven't done much TH before:

- What should we do about `since` declarations? This removes existing `since` declarations because it's generating functions instead of writing them out with comments.
- Should this be doing more quoting? (Especially inside `InstanceD`)
- Should this use a better way to assign names (e.g. using `''RawSql` and `'rawSqlCols`)? Should this move relevant types to the `TH` module just to be able to do that?
- Should this use the lowercase versions of TH-related constructors?

Before submitting:

  - see question above 

After submitting:


mitchellvitez:
CI seems to be failing because `TupE` takes a `Maybe Exp` in some versions and an `Exp` in others. Will take a look at which `MIN_VERSION` this needs later in the week, if we actually want to merge this

persistent/Database/Persist/Sql/TH.hs
```diff
@@ -0,0 +1,119 @@
...
+minTupleSize :: Int
+minTupleSize = 3 -- instance for 2 is already written
+
+maxTupleSize :: Int
+maxTupleSize = 62 -- GHC limitation
+
+pairedList :: [a] -> [[a]]
+pairedList [] = []
+pairedList [x] = [[x]]
+pairedList (x:y:rest) = [x, y] : pairedList rest
```
parsonsmatt, line 32: I feel like this would be better as a new data structure that captures the invariant we want.

```suggestion
data PairedList a 
  = Pair a a (PairedList a)
  | End (Maybe a)

pairedList :: [a] -> PairedList a
pairedList [] = End Nothing
pairedList [x] = End (Just x)
pairedList (x:y:rest) = Pair x y $ pairedList rest
```

parsonsmatt: `PairedList` can be factored out to `([(a, a)], Maybe a)`, but a tuple of list of tuples sounds annoying to work with.

persistent/Database/Persist/Sql/TH.hs
```diff
@@ -0,0 +1,119 @@
...
+flatTupleT names = foldl1 AppT $ (TupleT $ length names) : map VarT names
+
+nestedTupleT :: [Name] -> Type
+nestedTupleT names =
+   foldl1 AppT $ (TupleT $ (length names + 1) `div` 2) : map tupleOrSingleT (pairedList names)
+
+tupleOrSingleT :: [Name] -> Type
+tupleOrSingleT [a, b] = AppT (AppT (TupleT 2) (VarT a)) (VarT b)
+tupleOrSingleT [a] = VarT a
+tupleOrSingleT x = error $ "tupleOrSingleT failed to convert: " <> show x
```
parsonsmatt, line 44:
```suggestion
nestedTupleT :: [Name] -> Type
nestedTupleT names =
   foldl1 AppT $ (TupleT $ (length names + 1) `div` 2) : tupleOrSingleT (pairedList names)

tupleOrSingleT :: PairedList Name -> [Type]
tupleOrSingleT (Pair a b rest) = 
    AppT (AppT (TupleT 2) (VarT a)) (VarT b) : tupleOrSingleT rest
tupleOrSingleT (End (Just a)) = [VarT a]
tupleOrSingleT (End Nothing) =[]
```

No more errors :)

persistent/Database/Persist/Sql/TH.hs
```diff
@@ -0,0 +1,119 @@
...
+        tupleOrSingleP [a, b] = TupP $ map VarP [a, b]
+        tupleOrSingleP [a] = VarP a
+        tupleOrSingleP x = error $ "tupleOrSingleP failed to convert: " <> show x
+
+rawSqlInstanceN :: Int -> Q Dec
+rawSqlInstanceN n = do
+    names <- replicateM n $ newName "x"
+
+    let constraints = map (\name -> AppT (ConT $ mkName "RawSql") (VarT name)) names
+        tuple = AppT (ConT $ mkName "RawSql") . foldl1 AppT $ (TupleT $ length names) : map VarT names
```
parsonsmatt, line 79: `mkName` makes a *local* name, which can be shadowed in the environment and refer to local things. End users will need to ensure these names are in scope, which means importing the modules specifically. This isn't a great UX - generally speaking, folks don't wantt o see "Name not in scope" errors for things they didn't write!

For constant terms like this, prefer the Template Haskell quotes. 

```suggestion
        tuple = AppT (ConT $ 'RawSql) . foldl1 AppT $ (TupleT $ length names) : map VarT names
```

persistent/Database/Persist/Sql/TH.hs
```diff
@@ -0,0 +1,119 @@
...
+
+    pure $ InstanceD Nothing constraints tuple
+        [ FunD (mkName "rawSqlCols") [Clause [] (NormalB rawSqlColsBody) []]
+        , FunD (mkName "rawSqlColCountReason") [Clause [] (NormalB rawSqlColCountReasonBody) []]
+        , FunD (mkName "rawSqlProcessRow") [Clause [] (NormalB rawSqlProcessRowBody) []]
```
parsonsmatt, line 84:
```suggestion
        [ FunD ('rawSqlCols) [Clause [] (NormalB rawSqlColsBody) []]
        , FunD ('rawSqlColCountReason) [Clause [] (NormalB rawSqlColCountReasonBody) []]
        , FunD ('rawSqlProcessRow) [Clause [] (NormalB rawSqlProcessRowBody) []]
```

persistent/Database/Persist/Sql/TH.hs
```diff
@@ -0,0 +1,119 @@
...
+        ]
+
+    where
+        rawSqlColsBody :: Exp
+        rawSqlColsBody =
+            LamE [VarP $ mkName "e"]
+                 (InfixE (Just $ AppE (VarE $ mkName "rawSqlCols") (VarE $ mkName "e"))
+                         (VarE $ mkName ".")
+                         (Just . VarE $ mkName $ "from" ++ show n))
```
parsonsmatt, line 93: I *generally* like the QuasiQuoter when it makes sense to use - if the AST is constant (eg not doing a `foldl1 AppT` for folding a list of arguments), then it can be a nice improvement to readability.

```suggestion
        rawSqlColsBody :: Q Exp
        rawSqlColsBody =
            [e| \e -> rawSqlCols e . $(varE (mkName $ "from" ++ show n)) |]
```

This also correctly captures `rawSqlCols` into the quote, which means you don't need to worry about `'rawSqlCols` vs `mkName "rawSqlCols"`.

persistent/Database/Persist/Sql/TH.hs
parsonsmatt: So, on another note, you can *also* make this a QuasiQuote.

```haskell
[d|
    instance $(pure constraints) => $(pure tuple) where
        rawSqlCols = $(rawSqlColsBody)
        rawSqlColCountReason = $(rawSqlColCountReasonBody)
        rawSqlProcessRow = $(rawSqlProcessRowBody)
```

The expression in a splice needs to be `Q Exp` (or something that can be automatically `lift :: Lift a => a -> Q Exp`. If you use the QQ for the `rawSqlColsBody`, then it returns a `Q Exp`, and you can use it directly.

Or, you could just have the instance completely inline with the QQ, too.

```haskell
[d|
    instance $(pure constraints) => $(pure tuple) where
        rawSqlCols = \e -> rawSqlCols e . $(varE $ mkName $ "from" ++ show n)
        -- etc...
  |]
```

parsonsmatt (changes requested):
I left a few suggestions on code style and Template Haskell usage - code looks great but I think there's a few ways to improve it.

Ultimately, though, I'm a little skeptical of including this. If we *didn't* already have all the possible tuple instances, then I'd accept this change. But I'm a bit wary of deleting all that code, and then regenerating it on every single compile, as well as incurring the TH cost of needing to recompile the module whenever any upstream dependency changes.

I genuinely appreciate the contribution but I don't think I want to incorporate it. I'll leave this PR open so if anyone else wants to voice a strong preference to merge it in, they can.

mitchellvitez:
>I left a few suggestions on code style and Template Haskell usage

Thanks so much for all the feedback! Definitely helpful for my learning, at least

>I don't think I want to incorporate it

No problem there, the reasoning makes sense

persistent/Database/Persist/Sql/TH.hs
mitchellvitez: When I try to quote the instance declaration like this
```haskell
    [d|
      instance $(pure constraints) => $(pure tuple) where
          rawSqlCols = \e -> rawSqlCols e . $(fromOrToN From n)
          rawSqlColCountReason = rawSqlColCountReason . $(fromOrToN From n)
          rawSqlProcessRow = fmap $(fromOrToN To n) . rawSqlProcessRow
      |]
```
I get this back from -ddump-splices, where the names of functions in the instance look like they're not matching up with the names within the expression (`rawSqlCols_02y` /= `rawSqlCols`). This then leads to a bunch of errors in compiling what TH generated (redefining functions, missing names, etc.)
```haskell
instance (RawSql x_a48o6, RawSql x_a48o7, RawSql x_a48o8) =>
             RawSql (x_a48o6, x_a48o7, x_a48o8) where
      rawSqlCols_02y = \ e_a48o9 -> (rawSqlCols e_a48o9 . from3)
      rawSqlColCountReason_02y = (rawSqlColCountReason . from3)
      rawSqlProcessRow_02y = (fmap to3 . rawSqlProcessRow)
```
However, trying to fix this issue by putting a `Name` there directly (trying various ways)
```haskell
    [d|
      instance $(pure constraints) => $(pure tuple) where
          $(pure 'rawSqlCols) = \e -> rawSqlCols e . $(fromOrToN From n)
          $(pure $ mkName "rawSqlColCountReason") = rawSqlColCountReason . $(fromOrToN From n)
          $(lookupValueName "rawSqlProcessRow") = fmap $(fromOrToN To n) . rawSqlProcessRow
      |]
```
yields errors like
```haskell
Pattern bindings (except simple variables) not allowed in instance declaration:
        $(pure 'rawSqlCols) = \ e -> rawSqlCols e . $(fromOrToN From n)
```

I was able to get things working by handwriting TH `FunD` constructors over quoted expressions, but not when quoting the whole instance, so I've probably misunderstood how that works

mitchellvitez:
Closing because this was a TH rewrite of code that already exists, and regenerating it via TH has compilation costs

PR #726, opened 2017-10-30
Make PostgreSQL `selectSource` stream results.

naushadh:
Address #657 for `persistent-postgres`.

With this, persistent effectively will support streaming results when using `selectSource` for all* sql backends.

*MySQL streaming is available in [persistent-mysql-haskell](https://hackage.haskell.org/package/persistent-mysql-haskell) backend.

PR #1298, opened 2021-07-15
Add API for mem-constant DB streams, Postgres implementation

ivb-supercede:
This code is based on that contained in:
https://github.com/yesodweb/persistent/pull/726

This PR has two components:

  1. A new backend class, `PersistQueryStream`, for backends which *may* allow for constant-memory streaming of query results. This is intended to replace the use of `selectSource` from `PersistQueryRead`, which is "broken" in the sense that the `ConduitT` it returns rarely actually does any streaming.
  2. An implementation of `PersistQueryStream` for `SqlBackend` which is based on an internal field, `connPrepareCursor`, which allows for backends to provide a way to run queries using an internal cursor which iterates over rows as requested, instead of returning them all at once. This internal field is provided for the Postgres version of `SqlBackend`, using functions from `postgresql-simple`.

This code differs in an important way from the above PR in that it solves a memory issue in the way the `ConduitT` was constructed that meant that the implementation using cursors was also not memory-constant. This implementation is memory-constant, tested to large sets of results.

@timds

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -47,6 +47,7 @@ module Database.Persist.Postgresql
 import qualified Database.PostgreSQL.LibPQ as LibPQ
 
 import qualified Database.PostgreSQL.Simple as PG
+import qualified Database.PostgreSQL.Simple.Cursor as PGC
 import qualified Database.PostgreSQL.Simple.FromField as PGFF
 import qualified Database.PostgreSQL.Simple.Internal as PG
 import Database.PostgreSQL.Simple.Ok (Ok(..))
@@ -356,6 +357,7 @@ createBackend logFunc serverVersion smap conn =
     maybe id setConnUpsertSql (upsertFunction upsertSql' serverVersion) $
     setConnInsertManySql insertManySql' $
     maybe id setConnRepsertManySql (upsertFunction repsertManySql serverVersion) $
+    setConnPrepareCursor (prepareCursorStmt' conn) $
     mkSqlBackend MkSqlBackendArgs
         { connPrepare    = prepare' conn
         , connStmtMap    = smap
@@ -388,6 +390,16 @@ prepare' conn sql = do
         , stmtReset = return ()
         , stmtExecute = execute' conn query
         , stmtQuery = withStmt' conn query
+        }
+
+prepareCursorStmt' :: PG.Connection -> Text -> IO Statement
+prepareCursorStmt' conn sql = do
+    let query = PG.Query (T.encodeUtf8 sql)
+    return Statement
+        { stmtFinalize = return ()
+        , stmtReset = return ()
+        , stmtExecute = error "Database.Persist.Postgresql.prepareCursorStmt': Tried to execute a read-only select. Use select/query/stmtQuery"
+        , stmtQuery = withCursorStmt' conn query
         }
 
 insertSql' :: EntityDef -> [PersistValue] -> InsertSqlResult
@@ -496,7 +508,7 @@ withStmt' conn query vals =
                 rowCount <- LibPQ.ntuples ret
                 return (ret, rowRef, rowCount, oids)
       let getters
-            = map (\(col, oid) -> getGetter conn oid $ PG.Field rt col oid) ids
+            = map (\(col, oid) -> getGetter oid $ PG.Field rt col oid) ids
       return (rt, rr, rc, getters)
 
     closeS (ret, _, _, _) = LibPQ.unsafeFreeResult ret
@@ -527,9 +539,30 @@ withStmt' conn query vals =
                                                         Errors [] -> error "Got an Errors, but no exceptions"
                                                         Ok v  -> return v
 
--- | Avoid orphan instances.
-newtype P = P PersistValue
+withCursorStmt' :: MonadIO m
+          => PG.Connection
+          -> PG.Query
+          -> [PersistValue]
+          -> Acquire (ConduitM () [PersistValue] m ())
+withCursorStmt' conn query vals =
+    foldWithCursor `fmap` mkAcquire openC closeC
+  where
+    openC = do
+        rawquery <- liftIO $ PG.formatQuery conn query (map P vals)
+        PGC.declareCursor conn (PG.Query rawquery)
+    closeC = PGC.closeCursor
+    foldWithCursor cursor = go
+      where
+        go = do
+            -- 256 is the default chunk size used for fetching
+            rows <- liftIO $ PGC.foldForward cursor 256 processRow []
+            case rows of
+                Left final -> CL.sourceList final
+                Right nonfinal -> CL.sourceList nonfinal >> go
+    processRow s row = pure $ s <> [map unP row]
 
+-- | To avoid orphan instances.
+newtype P = P { unP :: PersistValue }
 
 instance PGTF.ToField P where
     toField (P (PersistText t))        = PGTF.toField t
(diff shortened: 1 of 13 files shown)


ivb-supercede:
Interestingly, this code has some odd performance characteristics that I can't figure out the reason for. Calling `selectStream` seems to be slower than using `persistent-pagination`, when performing a like-for-like query - and the profiling information seems to suggest that it's because of row parsing being slower in the cursor code. I couldn't find the exact reason, however.

jezen:
@parsonsmatt Hope you don't mind me pinging you for review on this — just thought you'd be a likely candidate as it's closely related to `persistent-pagination` 🙂

persistent/Database/Persist/Sql/Raw.hs
```diff
@@ -74,12 +90,18 @@ getStmt sql = do
     liftIO $ getStmtConn conn sql
 
 getStmtConn :: SqlBackend -> Text -> IO Statement
-getStmtConn conn sql = do
+getStmtConn = getStmtConn' False
+
+getStmtConn' :: Bool -> SqlBackend -> Text -> IO Statement
+getStmtConn' useCursorIfPossible conn sql = do
```
parsonsmatt, line 96: I'm not really a fan of `Bool` parameters when we have some specific intent - it makes it hard to read at the use-sites. Introducing a separate data type that specifies the behavior would be preferable IMO.

persistent/Database/Persist/Class/PersistQuery.hs
```diff
@@ -54,6 +57,31 @@ class (PersistCore backend, PersistStoreRead backend) => PersistQueryRead backen
...
+    -- By default, this behaves identically to 'selectSourceRes'.
+    --
+    -- @since 2.13.2.0
+    selectSourceStream
+        :: (PersistRecordBackend record backend, MonadResource m)
+        => [Filter record]
+        -> [SelectOpt record]
+        -> ConduitM () (Entity record) (ReaderT backend m) ()
+    selectSourceStream filts opts = do
+      srcRes <- lift $ selectSourceRes filts opts
```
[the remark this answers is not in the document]
parsonsmatt: Eg - an `instance PersistQueryStream MongoContext` would compile just fine, but it would probably fail to work in a reasonable way. That seems bad to me.

persistent/Database/Persist/Class/PersistQuery.hs
```diff
@@ -54,6 +57,31 @@ class (PersistCore backend, PersistStoreRead backend) => PersistQueryRead backen
     exists :: (MonadIO m, PersistRecordBackend record backend)
            => [Filter record] -> ReaderT backend m Bool
 
+-- | Backends supporting conditional read operations, which may also support
+-- streaming the results in a memory-constant way.
+--
+-- @since 2.13.2.0
+class (PersistQueryRead backend) => PersistQueryStream backend where
```
parsonsmatt, line 64: I think adding this class makes the expectations around `selectSource` *really* weird - that's a class method on `PersistQuery`, so why do we have a separate method here?

persistent/Database/Persist/Sql/Raw.hs
```diff
@@ -74,12 +90,18 @@ getStmt sql = do
...
     smap <- liftIO $ readIORef $ connStmtMap conn
     case Map.lookup sql smap of
         Just stmt -> return stmt
         Nothing -> do
-            stmt' <- liftIO $ connPrepare conn sql
+            stmt' <- liftIO $
+                case connPrepareCursor conn of
+                    Just prepareCursor | useCursorIfPossible -> prepareCursor sql
+                    _ -> connPrepare conn sql
```
parsonsmatt, line 104: so this is where we switch - if we have `connPrepareCursor = Just ...` and we're trying to stream, then we do this thing. otherwise, we do the old thing.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -527,9 +539,30 @@ withStmt' conn query vals =
...
+    closeC = PGC.closeCursor
+    foldWithCursor cursor = go
+      where
+        go = do
+            -- 256 is the default chunk size used for fetching
+            rows <- liftIO $ PGC.foldForward cursor 256 processRow []
+            case rows of
+                Left final -> CL.sourceList final
+                Right nonfinal -> CL.sourceList nonfinal >> go
+    processRow s row = pure $ s <> [map unP row]
```
parsonsmatt, line 562: This implementation looks fine to me

parsonsmatt (changes requested):
So - 

Streaming rows from the database in a safe, composable, and fast way is apparently a hard problem.

This seems like a good solution! But you're noticing that the `persistent-pagination` approach is faster, anyway. I have no idea why that would be the case. And the tests don't cover a lot of the gnarly concurrency stuff that plague the other attempts I've seen.

So, I'm wondering - can this functionality be exposed from a 3rd party library, like `persistent-postgresql-streaming`? If not, what functionality do we need to export from `persistent` to support that library's creation?

Once `persistent-postgresql-streaming` is 'battle tested' and known to be safe, efficient, and awesome, then I'd be happy to merge it upstream. However, until it's known good, I'd prefer to err on the side of safety and not introduce features that may need to be deprecated or marked as unsafe, all due to the tricky vagaries of concurrency and foreign code.

I do appreciate the approach and the effort you spent, and I apologize for the delay in making this review.

persistent-postgresql/test/Streaming.hs
```diff
@@ -0,0 +1,58 @@
...
+
+          let resultStream = selectStream [] [ Asc StreamableNumberVal ]
+
+          let checkIncrementing expected (Entity _ (StreamableNumber actual)) = do
+                expected `shouldBe` actual
+                pure $ expected + 1
+
+          result <- runConduit $ resultStream .| CC.foldM checkIncrementing 1
+
+          result `shouldBe` numberOfRows
```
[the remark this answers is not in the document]
ivb-supercede: I'll double check it. I believe the nature of cursor queries should automatically avoid some of these concurrency issues, since streaming really fetches batches of rows from the cursor by name.

It might be necessary to make cursor statements bypass the `connStmtMap`, since the randomly-generated cursor name isn't included in the prepared query.

persistent/Database/Persist/Sql/Raw.hs
ivb-supercede: This has been replaced with a sum type.

persistent-postgresql/test/Streaming.hs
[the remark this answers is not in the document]
ivb-supercede: I'm struggling to follow the exact control flow, but: I believe this implementation is actually immune to some of these concurrency issues.

Testing has shown that certainly the single-threaded version behaves as expected. Conduits which simultaneously run the same query in a single-threaded application (one of the problematic uses mentioned in #981) yield results in the expected way, and don't affect each other. I *think* this is for the reason I mentioned above - each call to `withCursorStmt'` will cause a new temporary cursor to be created, the name of which is then used to grab rows in chunks.

The multi-threaded version might well still have issues - but testing that can wait until `ersistent-postgresql-streaming` is spun off.

ivb-supercede:
>So, I'm wondering - can this functionality be exposed from a 3rd party library, like persistent-postgresql-streaming? If not, what functionality do we need to export from persistent to support that library's creation?

I haven't tried this out yet, but I think the biggest obstacle is the inability to go from the `SqlBackend` built from a PostgreSQL `Connection` back to the `Connection` itself - so it's not possible to write a function of type
```haskell
getPGConnection :: ReaderT SqlBackend m Connection
```
which would be required to call `PGC.foldForward` (or any of the lower-level `LibPQ` functions). The only way that's possible in this PR is because the cursor code hooks into the point where the `SqlBackend` is first created and therefore has access to the `Connection` at that point.

There is, as far as I can tell, no easy way to get around that with the current polymorphic `SqlBackend`. One ugly option would be to give `SqlBackend` a field which *might* contain a `Connection`:
```
data SqlBackend = SqlBacked
  { ...
  , connPostgreSQLConnection :: Maybe Connection
  }
```

Another is for `persistent-postgresql-streaming` to simply require the user to provide the `Connection` to use `selectStream`, which is also very ugly.

parsonsmatt:
Makes sense - there was some work done for the SQLite backend to expose the 'raw' connection. If exposing the raw Postgres connection (in a similar way) works, then we can get that out the door for y'all relatively quickly and easily.

ivb-supercede:
The code from this PR is now up in [a separate library](https://github.com/SupercedeTech/persistent-postgresql-streaming).

PR #1301, opened 2021-07-25
persistent-mysql: Prevent fetching constraint info from other databases

### LINKED ISSUE #1300: persistent-mysql: Bug in fetching constraint info during migration
Having two copies of a database on the same machine results in the following error when running `runMigration migrateCore`:

```
MySQL.getColumn/getRef: error fetching constraints. Expected a single result for foreign key query for table: session and column: user_id but got: [[PersistText "user",PersistText "session_user_id_fkey",PersistInt64 1,PersistText "CASCADE",PersistText "CASCADE"],[PersistText "user",PersistText "session_user_id_fkey",PersistInt64 1,PersistText "CASCADE",PersistText "CASCADE"]]
```

I've investigated the error and it turns out that a constraint of the same name from the other database is also fetched, which results in a duplicate.

Tested on MySQL 5.7 with these package versions:
```
persistent 2.11.0.4
persistent-mysql 2.10.3.1
persistent-template 2.9.1.0
```

I'll create a PR with a fix for this issue.

dgaw:
Fixes #1300

### DIFF
--- a/persistent-mysql/ChangeLog.md
+++ b/persistent-mysql/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for persistent-mysql
 
+## 2.13.0.2
+
+* Bugfix: prevent fetching constraint info from other databases during migrations [#1301](https://github.com/yesodweb/persistent/pull/1301)
+
 ## 2.13.0.1
 
 * [#1275](https://github.com/yesodweb/persistent/pull/1275)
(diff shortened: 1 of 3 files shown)


parsonsmatt:
Great, thanks!

Can you add a note to the changelog and do a patch version bump? I'll get this released ASAP.

parsonsmatt (approved):

dgaw:
Cheers. I've updated the changelog and bumped the version.

parsonsmatt:
thanks! released as `persistent-mysql-2.13.0.2`

Issue #1156, opened 2020-11-11
Variable naming in Persistent


MaxGabriel:
When I read the Persistent codebase, it's often challenging to figure out what is going on, and I attribute a lot of this to very short variable names. I took a few examples from the codebase:

This code is littered with tiny variable names. What is `mid` exactly?

```
    (idField, primaryComposite, uniqs, foreigns) = foldl' (\(mid, mp, us, fs) attr ->
        let (i, p, u, f) = takeConstraint ps name' cols attr
            squish xs m = xs `mappend` maybeToList m
        in (just1 mid i, just1 mp p, squish us u, squish fs f)) (Nothing, Nothing, [],[]) attribs
```

What is `n` here? And what is `rest`—`rest` could be the `rest` of anything. Is it worth saving a handful of characters to write `ps` vs `persistSettings`, which is extremely clear?

```
takeUniq :: PersistSettings
         -> Text
         -> [FieldDef]
         -> [Text]
         -> UniqueDef
takeUniq ps tableName defs (n:rest)
```

What is `n`? `nu`?

```
showColumn :: Column -> Text
showColumn (Column n nu sqlType' def _defConstraintName _maxLen _ref) = T.concat
    [ escape n
    , " "
    , showSqlType sqlType'
    , " "
    , if nu then "NULL" else "NOT NULL"
    , case def of
        Nothing -> ""
        Just s -> " DEFAULT " <> s
    ]
```

In all these cases, you _can_ figure it out, but imo it's way harder than it should be. Here's some code I've written in Persistent I find much easier to read (admittedly this is very biased b/c I wrote it):

```
requireExtensions :: [[Extension]] -> Q ()
requireExtensions requiredExtensions = do
  unenabledExtensions <- filterM (fmap (not . or) . traverse isExtEnabled) requiredExtensions
```

```
getServerVersionNonEmpty :: PG.Connection -> IO (NonEmpty Word)
getServerVersionNonEmpty conn = do
  [PG.Only version] <- PG.query_ conn "show server_version";
  case AT.parseOnly parseVersion (T.pack version) of
    Left err -> throwIO $ PostgresServerVersionError $ "Parse failure on: " <> version <> ". Error: " <> err
    Right versionComponents -> case NEL.nonEmpty versionComponents of
      Nothing -> throwIO $ PostgresServerVersionError $ "Empty Postgres version string: " <> version
      Just neVersion -> pure neVersion

  where
    -- Partially copied from the `versions` package
[...]

PR #1315, opened 2021-09-07
Refactor entity constraint parsing in Quasi module

danbroooks:

After submitting your PR:

-----

A further contribution towards https://github.com/yesodweb/persistent/issues/1156, addresses the fold referenced in that issue by creating a type that holds the various possible constraints, and can be built up and merged into one. I have used `Monoid`s / `Semigroup`s however the instances are not lawful. I will put some further detail about this as a comment in the relevant area in the PR.

I have also renamed `splitExtras` to `parseEntityFields` as this is more descriptive I think. This function is called when the fields of an entity are extracted, the first step being to split the `'extra'` tokens from the rest of the entity fields which are added to a value of type `M.Map Text [ExtraLine]`, where the key is the entity name and its `[ExtraLine]` values are the entity fields.

### DIFF
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -26,7 +26,7 @@ module Database.Persist.Quasi.Internal
     , parseFieldType
     , associateLines
     , LinesWithComments(..)
-    , splitExtras
+    , parseEntityFields
     , takeColsEx
     -- * UnboundEntityDef
     , UnboundEntityDef(..)
@@ -311,7 +311,7 @@ toParsedEntityDef lwc = ParsedEntityDef
             _ -> (False, EntityNameHS entityName)
 
     (attribs, extras) =
-        splitExtras fieldLines
+        parseEntityFields fieldLines
 
 isDocComment :: Token -> Maybe Text
 isDocComment tok =
@@ -825,12 +825,12 @@ mkAutoIdField' dbName entName idSqlType =
 keyConName :: EntityNameHS -> Text
 keyConName entName = unEntityNameHS entName `mappend` "Id"
 
-splitExtras
+parseEntityFields
     :: [Line]
     -> ( [[Token]]
        , M.Map Text [ExtraLine]
        )
-splitExtras lns =
+parseEntityFields lns =
     case lns of
         [] -> ([], M.empty)
         (line : rest) ->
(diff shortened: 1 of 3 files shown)


persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -928,28 +925,72 @@ getSqlNameOr def =
...
+            { entityConstraintDefsIdField = just1 (entityConstraintDefsIdField a) (entityConstraintDefsIdField b)
+            , entityConstraintDefsPrimaryComposite = just1 (entityConstraintDefsPrimaryComposite a) (entityConstraintDefsPrimaryComposite b)
+            , entityConstraintDefsUniques = entityConstraintDefsUniques a <> entityConstraintDefsUniques b
+            , entityConstraintDefsForeigns = entityConstraintDefsForeigns a <> entityConstraintDefsForeigns b
+            }
+
+just1 :: (Show x) => Maybe x -> Maybe x -> Maybe x
+just1 (Just x) (Just y) = error $ "expected only one of: "
+  `mappend` show x `mappend` " " `mappend` show y
+just1 x y = x `mplus` y
```
danbroooks, line 947: I went with using Semigroups/Monoids here but this implementation of `just1` which is used in the semigroup instance makes it unlawful. I could re-write this to not use these typeclasses, but I thought it was a nice refactoring (means I could use `foldMap`, `mempty`).

Alternatively this could use something like `First` or `Last`, though it is probably a nicer user experience to receive this error than for it to silently proceed using one of the values over the other in the case of duplicate values. In fact that is definitely not what we would want.

It may be that now I have done an initial refactor, I could tweak things further again to not have unlawful instances, and use something instead `foldMap` / `mempty`

parsonsmatt: Hmm! I like using the `Monoid` and `Semigroup` instances here. That's neat. But I don't like having a `Semigroup` operation that will `error` on a *relatively* common case - that seems like a footgun.

I feel like a type that better represents this is something like:

```haskell
data SetOnce a 
    = SetOnce a 
     | NotSet 
     | SetManyTimes (NonEmpty a)

instance Semigroup (SetOnce a) where
    NotSet <> a = a

    SetOnce a <> NotSet = SetOnce a
    SetOnce a <> SetOnce b = SetManyTimes (a :| [b])
    SetOnce a <> SetManyTimes as = SetManyTimes (NEL.cons a as)

    a@(SetManyTimes xs) <> NotSet = a
    SetManyTimes xs <> SetOnce a = SetManyTimes (xs `snoc` a)
    SetManyTimes xs <> SetManyTimes ys = SetManyTimes (xs <> ys)

instance Monoid (SetOnce a) where
    mempty = NotSet
```



Rendering/erroring can then happen outside of the code.

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -928,28 +917,72 @@ getSqlNameOr def =
...
+                { entityConstraintDefsIdField =
+                    Just (takeId ps entityName rest)
+                }
+        _ | isCapitalizedText n ->
+            mempty
+                { entityConstraintDefsUniques =
+                    pure <$> takeUniq ps "" defs (n : rest)
+                }
+        _ ->
+            mempty
```
parsonsmatt, line 985: This is much nicer 👍🏻

PR #1317, opened 2021-09-09
Expose orderClause from PersistQuery.hs

ivb-supercede:
This is part of the followup to #1298.

This exposes `orderClause` in a similar manner to `filterClause`, to help third-party libraries which want to format their own SQL queries using Persistent's internal helpers.

It rewrites `orderClause` a bit to bring it in line with the other exposed functions, but there should be no functional changes as a result.

After submitting your PR:

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,10 +1,13 @@
 # Changelog for persistent
 
-## Unreleased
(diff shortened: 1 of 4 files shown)


parsonsmatt (approved):
looks good, thanks!

PR #1320, opened 2021-09-26
Remove mysql simple as a dependency 

belevy:
It seems strange to rely on mysql-simple when we are the high level library. We are essentially just pushing a bunch of data around for little gain seemingly since we use almost none of the library. This PR inlines much of the code that was occurring in mysql-simple. 

Potential Issues:
   - Performance of parsing: I am not using any high performance parsing library, just Text.Read. From running the tests I didnt see any obvious degredation in performance. 
   - Do we have a specific version of time that we are targeting? I only tested with 1.9.3 and may need to try with a lower version to ensure this works still.

After submitting your PR:

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -17,9 +17,9 @@ module Database.Persist.MySQL
     , createMySQLPool
     , module Database.Persist.Sql
     , MySQL.ConnectInfo(..)
-    , MySQLBase.SSLInfo(..)
+    , MySQL.SSLInfo(..)
     , MySQL.defaultConnectInfo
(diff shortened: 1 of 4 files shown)


paul-rouse:
I am wondering what this gains, since it's quite a large change to code which has apparently worked well for many years.  Is there a performance bottleneck caused by using `mysql-simple`?

belevy:
It's more about understanding what the code is actually doing. The use of mysql-simple really obfuscates how the query is getting created. 

It's also a maintenance burden to have an extra dependency that we only half use. If we wanted to support real prepared statements instead of using escape in the future it would require support to be added to mysql-simple first.

Obviously I am ok with this PR being rejected, it just started it as a feasibility test and I felt it was worth opening the PR.

PR #1333, opened 2021-11-01
Add mysql json field support

belevy:
Now that mysql-simple and mysql have been updated to properly handle JSON values coming out of a mysql db (supported since mysql 5.7.8) this PR adds support to the MySQL backend to read these values. It does not add a PersistField instance for Value's. It is probably not great that Postgres has an orphan instance for Value. Would rather there be a Json newtype and JsonB newtype. 

After submitting your PR:

### DIFF
--- a/persistent-mysql/test/JSONTest.hs
+++ b/persistent-mysql/test/JSONTest.hs
@@ -1,0 +1,33 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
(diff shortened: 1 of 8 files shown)


persistent-mysql/test/JSONTest.hs
```diff
@@ -0,0 +1,33 @@
...
+import qualified Data.Conduit.List as CL
+import Database.Persist.MySQL
+import MyInit
+
+specs :: Spec
+specs = describe "JSONTest" $ do
+    it "can select json with rawsql" $ db $ do
+        let testJSON = toJSON $ [object [ "test" .= ("value" :: Text) ]]
+        [[PersistByteString value]] <- runConduit $ rawQuery "select JSON_ARRAY(JSON_OBJECT('test', 'value'))" [] .| CL.consume
+        liftIO $ Just testJSON `shouldBe` (decode $ BSL.fromStrict value)
```
parsonsmatt, line 33: hell yeah

parsonsmatt (approved):
awesome thanks!

this is a bugfix patch for `persistent-mysql`, so I can get this out tonight.

I agree that the orphan `instance PersistValue Value` in `persistent-postgresql` isn't great. Would be better to have a `newtype` for precisely that reason.

PR #1334, opened 2021-11-03
Support ghc-9.2.1 (aeson-2, template-haskell-2.18)

ruomad:
Tentative patch to
- support  aeson-2 (introducing Key/KeyMap as an abstraction in replacement to HashMap Text Value)
- support template-haskell-2.18.0 (changes the type of some functions, esp. ConP)
which makes it possible to compile with ghc-9.2.1


parsonsmatt:
Thanks for making the PR! Was it closed accidentally?

ruomad:
No ! I closed it because of the mess in commits, I'll submit a cleaner version (the final code is the same though).
Sorry for this

PR #1335, opened 2021-11-04
Support aeson-2 and template-haskell-2.18 (ghc-9.2.1)

ruomad:
Tentative patch to

    support aeson-2 (introducing Key/KeyMap as an abstraction in replacement to HashMap Text Value)
    support template-haskell-2.18.0 (changes the type of some functions, esp. ConP)
    which makes it possible to compile with ghc-9.2.1

(cleaner version, sorry for the noise)

### DIFF
--- a/persistent/Database/Persist/Class/PersistConfig.hs
+++ b/persistent/Database/Persist/Class/PersistConfig.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE CPP #-}
+
 module Database.Persist.Class.PersistConfig
     ( PersistConfig (..)
     ) where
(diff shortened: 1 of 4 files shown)


ruomad:
Hum, I don.t know why this fails… the log is not very helpful

parsonsmatt:
It's a problem with `persistent-mongoDB` - [relevant log line](https://github.com/yesodweb/persistent/runs/4100440023?check_suite_focus=true#step:15:476)

ruomad:
Well, OK.
I indeed only use (and build) persistent-postgresql, so I'll have to take a look to other persistent modules as well.
I was supposing the specific modules were not affected by internal changes in aeson/persistent core.

brandon-leapyear:
I believe `Database.Persist.TH` is still failing. I think it's this line:
https://github.com/yesodweb/persistent/blob/29a9cbef75850fbec6b196b06dd29f0ecb180b80/persistent/Database/Persist/TH.hs#L2673
This is still using `Text.pack` for a Key.

brandon-leapyear:
Forked here: https://github.com/yesodweb/persistent/pull/1338. Feel free to merge/cherry-pick back

PR #1341, opened 2021-11-19
Add SQL Connection & Statement hooks

iand675:
This PR builds on top of https://github.com/yesodweb/persistent/pull/1327

It adds support for hooking into various aspects of the lifecycle of a connection / statements. This functionality opens up support for tracing queries within the context of a larger chunk of code: 
image

To see how this is being used, see the hs-opentelemetry project: https://github.com/iand675/hs-opentelemetry/blob/main/instrumentation/persistent/src/OpenTelemetry/Instrumentation/Persistent.hs

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -47,6 +47,7 @@ module Database.Persist.Postgresql
     , upsertManyWhere
     , openSimpleConn
     , openSimpleConnWithVersion
+    , getSimpleConn
     , tableName
     , fieldName
     , mockMigration
@@ -120,6 +121,8 @@ import Database.Persist.Sql
 import qualified Database.Persist.Sql.Util as Util
 import Database.Persist.SqlBackend
 import Database.Persist.SqlBackend.StatementCache (StatementCache, mkSimpleStatementCache, mkStatementCache)
+import qualified Data.Vault.Strict as Vault
+import System.IO.Unsafe (unsafePerformIO)
 
 -- | A @libpq@ connection string.  A simple example of connection
 -- string would be @\"host=localhost port=5432 user=test
@@ -365,6 +368,17 @@ openSimpleConnWithVersion getVerDouble logFunc conn = do
     serverVersion <- oldGetVersionToNew getVerDouble conn
     return $ createBackend logFunc serverVersion (mkStatementCache smap) conn
 
+underlyingConnectionKey :: Vault.Key PG.Connection
+underlyingConnectionKey = unsafePerformIO Vault.newKey
+{-# NOINLINE underlyingConnectionKey #-}
+
+-- | Access underlying connection, returning 'Nothing' is the 'SqlBackend'
+-- provided isn't backed by postgresql-simple.
+--
+-- @since 2.14.0
+getSimpleConn :: (BackendCompatible SqlBackend backend) => backend -> Maybe PG.Connection
+getSimpleConn = Vault.lookup underlyingConnectionKey <$> getConnVault
+
 -- | Create the backend given a logging function, server version, mutable statement cell,
 -- and connection.
 createBackend :: LogFunc -> NonEmpty Word
@@ -374,7 +388,7 @@ createBackend logFunc serverVersion smap conn =
     maybe id setConnUpsertSql (upsertFunction upsertSql' serverVersion) $
     setConnInsertManySql insertManySql' $
     maybe id setConnRepsertManySql (upsertFunction repsertManySql serverVersion) $
-    mkSqlBackend MkSqlBackendArgs
+    modifyConnVault (Vault.insert underlyingConnectionKey conn) $ mkSqlBackend MkSqlBackendArgs
         { connPrepare    = prepare' conn
         , connStmtMap    = smap
         , connInsertSql  = insertSql'
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -295,7 +295,7 @@ open' modConn getVer constructor cstr logFunc = do
     conn <- PG.connectPostgreSQL cstr
     modConn conn
     ver <- getVer conn
-    smap <- mkStatementCache <$> mkSimpleStatementCache
+    smap <- newIORef mempty
     return $ constructor (createBackend logFunc ver smap) conn
 
 -- | Gets the PostgreSQL server version
@@ -364,25 +364,25 @@ openSimpleConn = openSimpleConnWithVersion getServerVersion
 -- @since 2.9.1
 openSimpleConnWithVersion :: (PG.Connection -> IO (Maybe Double)) -> LogFunc -> PG.Connection -> IO SqlBackend
 openSimpleConnWithVersion getVerDouble logFunc conn = do
-    smap <- mkSimpleStatementCache
+    smap <- newIORef mempty
     serverVersion <- oldGetVersionToNew getVerDouble conn
-    return $ createBackend logFunc serverVersion (mkStatementCache smap) conn
+    return $ createBackend logFunc serverVersion smap conn
 
 underlyingConnectionKey :: Vault.Key PG.Connection
 underlyingConnectionKey = unsafePerformIO Vault.newKey
 {-# NOINLINE underlyingConnectionKey #-}
 
--- | Access underlying connection, returning 'Nothing' is the 'SqlBackend'
+-- | Access underlying connection, returning 'Nothing' if the 'SqlBackend'
 -- provided isn't backed by postgresql-simple.
 --
--- @since 2.14.0
+-- @since 2.13.0
 getSimpleConn :: (BackendCompatible SqlBackend backend) => backend -> Maybe PG.Connection
 getSimpleConn = Vault.lookup underlyingConnectionKey <$> getConnVault
 
 -- | Create the backend given a logging function, server version, mutable statement cell,
 -- and connection.
 createBackend :: LogFunc -> NonEmpty Word
-              -> StatementCache -> PG.Connection -> SqlBackend
+              -> IORef (Map.Map Text Statement) -> PG.Connection -> SqlBackend
 createBackend logFunc serverVersion smap conn =
     maybe id setConnPutManySql (upsertFunction putManySql serverVersion) $
     maybe id setConnUpsertSql (upsertFunction upsertSql' serverVersion) $
@@ -1523,7 +1523,7 @@ mockMigrate allDefs _ entity = fmap (fmap $ map showAlterDb) $ do
 -- with the difference that an actual database is not needed.
 mockMigration :: Migration -> IO ()
 mockMigration mig = do
-    smap <- mkStatementCache <$> mkSimpleStatementCache
+    smap <- newIORef mempty
     let sqlbackend =
             mkSqlBackend MkSqlBackendArgs
                 { connPrepare = \_ -> do
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -119,6 +119,7 @@ import Database.Persist.Postgresql.Internal
 import Database.Persist.Sql
 import qualified Database.Persist.Sql.Util as Util
 import Database.Persist.SqlBackend
+import Database.Persist.SqlBackend.StatementCache (StatementCache, mkSimpleStatementCache, mkStatementCache)
 
 -- | A @libpq@ connection string.  A simple example of connection
 -- string would be @\"host=localhost port=5432 user=test
@@ -291,7 +292,7 @@ open' modConn getVer constructor cstr logFunc = do
     conn <- PG.connectPostgreSQL cstr
     modConn conn
     ver <- getVer conn
-    smap <- newIORef $ Map.empty
+    smap <- mkStatementCache <$> mkSimpleStatementCache
     return $ constructor (createBackend logFunc ver smap) conn
 
 -- | Gets the PostgreSQL server version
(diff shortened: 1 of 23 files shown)


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -360,20 +364,31 @@ openSimpleConn = openSimpleConnWithVersion getServerVersion
...
 createBackend :: LogFunc -> NonEmpty Word
-              -> IORef (Map.Map Text Statement) -> PG.Connection -> SqlBackend
+              -> StatementCache -> PG.Connection -> SqlBackend
 createBackend logFunc serverVersion smap conn =
     maybe id setConnPutManySql (upsertFunction putManySql serverVersion) $
     maybe id setConnUpsertSql (upsertFunction upsertSql' serverVersion) $
     setConnInsertManySql insertManySql' $
     maybe id setConnRepsertManySql (upsertFunction repsertManySql serverVersion) $
-    mkSqlBackend MkSqlBackendArgs
+    modifyConnVault (Vault.insert underlyingConnectionKey conn) $ mkSqlBackend MkSqlBackendArgs
```
parsonsmatt, line 391: To get access to the underlying backend, use the `RawPostgresql` type https://github.com/yesodweb/persistent/blob/69bdefd3978aa27004fb118f301aae0fb1c40640/persistent-postgresql/Database/Persist/Postgresql.hs#L1845

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -360,20 +364,31 @@ openSimpleConn = openSimpleConnWithVersion getServerVersion
...
 openSimpleConnWithVersion getVerDouble logFunc conn = do
-    smap <- newIORef $ Map.empty
+    smap <- mkSimpleStatementCache
     serverVersion <- oldGetVersionToNew getVerDouble conn
-    return $ createBackend logFunc serverVersion smap conn
+    return $ createBackend logFunc serverVersion (mkStatementCache smap) conn
+
+underlyingConnectionKey :: Vault.Key PG.Connection
+underlyingConnectionKey = unsafePerformIO Vault.newKey
+{-# NOINLINE underlyingConnectionKey #-}
```
parsonsmatt, line 373: This doesn't feel right to me. My understanding of this behavior is:

1. We create this `Vault.Key PG.Connection` lazily, whenever the first time we call this function. Then the value is cached due to GHC's applicative-normal-form optimization. So, we generate exactly one `Key Connection` here, and it is a global constant.
2. When we call `createBackend` or `getSimpleConn` (whichever one comes first) then we create the `underlyingConnectionKey`. If we call `getSimpleConn` first, then we'll get `Nothing` in that function. If we call `createBackend` afterwards, then the `Key Conn` already exists, and we stuff the `Conn` into the `Vault`.
3. So, I think the problem with this, is that we are *overwriting* the underlying `Conn` with every call to `createBackend`. Is this a problem?

`withPostgresqlPool` delegates to `withSqlPool open'` which itself calls `return $ constructor (createBackend logFunc ver smap) conn`. Which means that, on every new `SqlBackend` that is created, we end up overwriting the value here, which means that `getSimpleConn` is going to get the most *recent* `PG.Conn` that has been created.

persistent/Database/Persist/Sql/Raw.hs
```diff
@@ -75,9 +75,10 @@ getStmt sql = do
 
 getStmtConn :: SqlBackend -> Text -> IO Statement
 getStmtConn conn sql = do
-    smap <- liftIO $ readIORef $ connStmtMap conn
-    case Map.lookup sql smap of
-        Just stmt -> return stmt
+    let cacheK = mkCacheKeyFromQuery sql
+    mstmt <- liftIO $ statementCacheLookup (connStmtMap conn) cacheK
```
parsonsmatt, line 79:
```suggestion
    mstmt <- statementCacheLookup (connStmtMap conn) cacheK
```

Technically not a necessary change - wonder why the original has `liftIO` when the code is in `IO` already 🤔

persistent/Database/Persist/Sql/Run.hs
```diff
@@ -93,15 +95,80 @@ runSqlPoolWithHooks
...
+    -- ^ Alter the backend prior to executing any actions with it.
+    , runBefore :: backend -> Maybe IsolationLevel -> m ()
+    -- ^ Run this action immediately before the action is performed.
+    , runAfter :: backend -> Maybe IsolationLevel -> m ()
+    -- ^ Run this action immediately after the action is completed.
+    , runOnException :: backend -> Maybe IsolationLevel -> UE.SomeException -> m ()
+    -- ^ This action is performed when an exception is received. The
+    -- exception is provided as a convenience - it is rethrown once this
+    -- cleanup function is complete.
+    }
```
parsonsmatt, line 119: This is nice!

To make backwards compatibility easier, I'd want to hide this constructor/accessors in an `.Internal` module, and expose an API for setting/modifying/adding hooks. Eg:

```haskell
module Database.Persist.SqlBackend.SqlPoolHooks 
  ( SqlPoolHooks
  , defaultSqlPoolHooks
  , setAlterBackend
  , modifyAlterBackend
  , addAlterBackendHook
  , addRunBeforeHook
  , addRunAfterHook
  -- etc..
  )
  where

module Database.Persist.SqlBackend.SqlPoolHooks.Internal where

data SqlPoolHooks m backend = SqlPoolHooks
    { alterBackend :: backend -> m backend
    -- ^ Alter the backend prior to executing any actions with it.
    , runBefore :: backend -> Maybe IsolationLevel -> m ()
    -- ^ Run this action immediately before the action is performed.
    , runAfter :: backend -> Maybe IsolationLevel -> m ()
    -- ^ Run this action immediately after the action is completed.
    , runOnException :: backend -> Maybe IsolationLevel -> UE.SomeException -> m ()
    -- ^ This action is performed when an exception is received. The
    -- exception is provided as a convenience - it is rethrown once this
    -- cleanup function is complete.
    }
```

persistent/Database/Persist/SqlBackend.hs
```diff
@@ -6,35 +6,45 @@ module Database.Persist.SqlBackend
       SqlBackend
     , mkSqlBackend
     , MkSqlBackendArgs(..)
+    , SqlBackendHooks(..)
```
parsonsmatt, line 9:
```suggestion
    , SqlBackendHooks
```

really would like to minimize additions to the public API that require breaking changes to modify (eg `data` constructor direct exposure)

persistent/Database/Persist/SqlBackend.hs
```diff
@@ -129,6 +139,34 @@ getConnUpsertSql
...
+
+-- | Get a tag displaying what database the 'SqlBackend' is for. Can be
+-- used to differentiate features in downstream libraries for different
+-- database backends.
+-- @since 2.14.0.0
+getRDBMS
+    :: (BackendCompatible SqlBackend backend, MonadReader backend m)
+    => m Text
+getRDBMS = do
+    asks (SqlBackend.connRDBMS . projectBackend)
```
parsonsmatt, line 168: Thanks for the docs and safe APIs ❤️

persistent/Database/Persist/SqlBackend/Internal.hs
```diff
@@ -69,7 +70,7 @@ data SqlBackend = SqlBackend
     -- When left as 'Nothing', we default to using 'defaultPutMany'.
     --
     -- @since 2.8.1
-    , connStmtMap :: IORef (Map Text Statement)
+    , connStmtMap :: StatementCache
```
parsonsmatt, line 73: i really like this abstraction, should make it a lot easier to extend and improve things

persistent/Database/Persist/SqlBackend/Internal.hs
```diff
@@ -132,6 +133,21 @@ data SqlBackend = SqlBackend
     -- When left as 'Nothing', we default to using 'defaultRepsertMany'.
     --
     -- @since 2.9.0
+    , connVault :: Vault
+    -- ^ Carry arbitrary payloads for the connection that
+    -- may be used to propagate information into hooks.
```
parsonsmatt, line 138: hm, ok, so the `Vault` is a per-`SqlBackend` thing. let me go review the `unsafePerformIO` stuff again

persistent/Database/Persist/SqlBackend/Internal/MkSqlBackend.hs
```diff
@@ -28,7 +27,7 @@ data MkSqlBackendArgs = MkSqlBackendArgs
     , connInsertSql :: EntityDef -> [PersistValue] -> InsertSqlResult
     -- ^ This function generates the SQL and values necessary for
     -- performing an insert against the database.
-    , connStmtMap :: IORef (Map Text Statement)
+    , connStmtMap :: StatementCache
```
parsonsmatt, line 30: Hmm. We may be able to avoid making this a breaking change, if there's a way to go from `IORef (Map Text Statement)` to a `StatementCache` in the function that translates a `MkSqlBackendArgs` into a `SqlBackend`.

persistent/Database/Persist/SqlBackend/StatementCache.hs
```diff
@@ -0,0 +1,65 @@
+{-# LANGUAGE RecordWildCards #-}
+module Database.Persist.SqlBackend.StatementCache
+  ( StatementCache
+  , StatementCacheKey
+  , mkCacheKeyFromQuery
+  , MkStatementCache(..)
+  , mkSimpleStatementCache
+  , mkStatementCache
+  ) where
```
parsonsmatt, line 9: blessed API 🙌🏻

persistent/Database/Persist/SqlBackend/StatementCache.hs
```diff
@@ -0,0 +1,65 @@
...
+  -- ^ Remove all statements from the cache. Implementations of this
+  -- should be sure to call `stmtFinalize` on all statements removed
+  -- from the cache.
+  --
+  -- @since 2.14.0
+  , statementCacheSize :: IO Int
+  -- ^ Get the current size of the cache.
+  --
+  -- @since 2.14.0
+  }
```
parsonsmatt, line 42:
```suggestion
data MkStatementCache = MkStatementCache
    { statementCacheLookup :: StatementCacheKey -> IO (Maybe Statement)
    -- ^ Retrieve a statement from the cache, or return nothing if it is not found.
    --
    -- @since 2.14.0
    , statementCacheInsert :: StatementCacheKey -> Statement -> IO ()
    -- ^ Put a new statement into the cache. An immediate lookup of
    -- the statement MUST return the inserted statement for the given
    -- cache key. Depending on the implementation, the statement cache MAY
    -- choose to evict other statements from the cache within this function.
    --
    -- @since 2.14.0
    , statementCacheClear :: IO ()
    -- ^ Remove all statements from the cache. Implementations of this
    -- should be sure to call `stmtFinalize` on all statements removed
    -- from the cache.
    --
    -- @since 2.14.0
    , statementCacheSize :: IO Int
    -- ^ Get the current size of the cache.
    --
    -- @since 2.14.0
    }
```

codebase uses 4 space indent

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: OK, so this is probably not an issue. While we do generate a single `Key Conn`, this is used to index into the `SqlBackend`'s internal `Vault`, not some globally available `Vault`. And, since there's only ever exactly one, then we can easily recover the `PG.Connection` for a given `SqlBackend`.

This is a bit gnarly but it shouldn't have any problems.

persistent/Database/Persist/SqlBackend/Internal/MkSqlBackend.hs
parsonsmatt: OK, yeah, this can definitely be a minor version bump! Which means there's no urgency to figure out the other breaking changes and try to bundle them together. I think this is the only breaking change in the PR.

persistent/Database/Persist/SqlBackend/StatementCache.hs
```diff
@@ -0,0 +1,65 @@
...
+    stmtMap <- newIORef Map.empty
+    pure $ MkStatementCache
+        { statementCacheLookup = \sql -> Map.lookup (cacheKey sql) <$> readIORef stmtMap
+        , statementCacheInsert = \sql stmt ->
+            modifyIORef' stmtMap (Map.insert (cacheKey sql) stmt)
+        , statementCacheClear = do
+            oldStatements <- atomicModifyIORef' stmtMap (\oldStatements -> (Map.empty, oldStatements))
+            traverse_ stmtFinalize oldStatements
+        , statementCacheSize = Map.size <$> readIORef stmtMap
+        }
```
parsonsmatt, line 59: The only thing here that needs `IO` is the `newIORef`, so if we extract+purify,

```haskell
mkSimpleStatementCache :: IORef (Map k v) -> MkStatementCache`
```

meaning that we can go from `IORef (Map Text Statement)` to a `StatementCache` purely, and can therefore do it in `mkSqlBackend`.

parsonsmatt (changes requested):
OK, having thoroughly reviewed this code:

I like it a lot. Great new feature. Excellent implementation. But we can make it a bit better.

1) Only expose data constructors and fields via the `.Internal` modules so we don't have to major-bump to add a field or change a field type. Expose accessors/modifiers/etc as appropriate. If you want to hand-this-off to me then I can take it home.
2) Don't change the type of `MkSqlBackendArgs` and instead convert from `IORef (Map Text STatement)` in `mkSqlBackend`. This allows us to release this as a minor version bump, instead of a major.

Thanks!!!

persistent-postgresql/Database/Persist/Postgresql.hs
iand675: This isn't sufficient for instrumentation of existing codebases unfortunately. The value lets us wrap existing connections without requiring pervasive downstream changes.

persistent/Database/Persist/SqlBackend/Internal/MkSqlBackend.hs
iand675: I want to mention that I'd like to ensure that _alternative_ statement caches can be provided too. In particular, the end goal here is that we could for example provide counters of the statement cache to detect memory leaks (real situation that can occur for long-lived postgresql connections currently AFAICT). Or as another example, we might want to provide an LRU version of the statement cache that can evict. That's why this was introduced as a breaking change. That could technically be deferred til a later date.

persistent-postgresql/Database/Persist/Postgresql.hs
iand675: Points 1 and 2 are intentional. We need a stable key in order to be able to reference the connection from external instrumentation tooling. AFAICT, there's no way you could call `getSimpleConn` prior to constructing a backend via `createBackend`, so you're guaranteed by virtue of the order of operations to have the key be initialized and able to return the underlying PostgreSQL connection. 

Regarding 3, I don't see how we're overwriting the underlying call in `createBackend`, since it's just an immutable function that constructs a backend from its constituent parts?

persistent/Database/Persist/SqlBackend/Internal/MkSqlBackend.hs
parsonsmatt: You should be able to write `setConnStatementCache :: StatementCache -> SqlBackend -> SqlBackend`, without breaking this.

In hindsight, I wish I made this a `Maybe` field - or even omit it from the list entirely. There's an obvious default, though it does mean that `mkSqlBackend :: MkSqlBackendArgs -> IO SqlBackend` needs to be done to produce the cache instead.

So, for a 2.13 release, can we avoid breaking this?

Then, for 2.14, we can either remove this (and only use the `setConnStatementCache`) or make it a `Maybe` field, and *also* make `mkSqlBackend` in `IO`.

parsonsmatt (approved):
looks great! thank you so much for the PR 😄

Issue #562, opened 2016-05-17
Persistent Entity Declaration QuasiQuote doesn't support type literals


pseudonom:
e.g. If you try to do

``` haskell

import Data.Finite
...

share [mkPersist sqlSettings, mkMigrate "migration"] [persistLowerCase|
  foo (Finite 20)
|]
```

you're met with an error like `Invalid field type "Finite 2001" PSFail ('2',"001")`.

This is an artifact of https://github.com/yesodweb/persistent/blob/master/persistent/Database/Persist/Quasi.hs#L63 and not any fundamental limitation. `type Twenty = 20` works fine as a workaround.

snoyberg:
I don't have an inherit objection to adding support for type level literals, just a vague concern that the error messages could end up being confusing for users. Want to take a stab at a PR to generalize the parser here?

parsonsmatt:
This should be relatively straightforward to add in the parser.

PR #1343, opened 2021-11-28
Implement Type Literal based field definitions 

danbroooks:
Implements https://github.com/yesodweb/persistent/issues/562

After submitting your PR:


persistent/Database/Persist/Types.hs
```diff
@@ -60,6 +62,7 @@ import Database.Persist.Types.Base
        , FieldCascade(..)
        , FieldDef(..)
        , FieldType(..)
+       , FieldTypeLit(..)
```
danbroooks, line 65: Given the comment at the top here:

https://github.com/yesodweb/persistent/blob/77545dc8a6a55c66ff6ed476175684c00943f8f4/persistent/Database/Persist/Types.hs#L48-L50

I wasn't sure if this was the right place for this import, but seemed like it was where everything else was re-exported (I need to use this type in `Persist/TH.hs` I think...)

Going forward should I import this from `Database.Persist.Types.Base` ? Or is re-exporting it here correct for the time being?

persistent/test/Database/Persist/QuasiSpec.hs
```diff
@@ -135,6 +135,15 @@ spec = describe "Quasi" $ do
...
                     )
 
+        it "handles numbers" $
+            parseLine "  one (Finite 1)" `shouldBe`
+                Just
+                    ( Line 2
+                        [ Token "one"
+                        , Token "Finite 1"
+                        ]
+                    )
```
danbroooks, line 145: This was an initial test I did as an exploratory thing to verify the parser was even parsing numbers at all to begin with (which it was), this test could potentially be removed now. Though it may also be worthwhile leaving in.

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -95,25 +95,59 @@ parseFieldType t0 =
...
-                     in PSSuccess (parseFieldTypePiece c a) b
-                | otherwise -> PSFail $ show (c, t')
+            Nothing -> pure PSDone
+            Just (x, xs) ->
+                parseSpace x xs
+                <|> parseParenEnclosed x xs
+                <|> parseList x xs
+                <|> parseNumericLit x xs
+                <|> parseTextLit x xs
+                <|> parseTypeCon x xs
```
parsonsmatt, line 107: Nice - this is great. Moving away from ad-hoc text munging and instead using more idiomatic Haskell classes.

persistent/Database/Persist/Types.hs
parsonsmatt: I'd rather not expose data constructors to the end-user in a non `Internal` module, since any change to those constructors necessitates a breaking change to the API version.

parsonsmatt: In this case, I'd be fine exporting it from `Database.Persist.Quasi.Internal`, since that's where it is used.

persistent/test/Database/Persist/QuasiSpec.hs
parsonsmatt: Yeah! It's good to have around as a regression. Thanks for the test!

persistent/test/Database/Persist/TH/TypeLitFieldDefsSpec.hs
```diff
@@ -0,0 +1,59 @@
...
+    one    (Finite 1)
+    twenty (Finite 20)
+
+WithLabelled
+    one    (Labelled "one")
+    twenty (Labelled "twenty")
+|]
+
+spec :: Spec
+spec = describe "TypeLitFieldDefs" $ do
```
parsonsmatt, line 47: awesome specs

parsonsmatt (approved):
Thanks! This is great :joy:

OK, so, since this is modifying a constructor on an exposed type (`FieldType`), then it's a major version bump. So it'll go in with `persistent-2.14`.

parsonsmatt:
Oh - this also needs a bump/changelog entry for `persistent-test`. And it may be a good idea to start putting a lower bound on `persistent-test` for the test-suites of the other packages.

parsonsmatt:
hmm, we have a test failure now. would like to get this merged in before 2.14, will try and get it sorted if you can't get to it soon

thanks!

Issue #1344, opened 2021-11-29
fails to build with aeson-2.0


juhp:
```
persistent                > /tmp/stack-06165844a3e2e2dd/persistent-2.13.2.1/Database/Persist/Class/PersistConfig.hs:46:36: error:
persistent                >     • Couldn't match type: Data.Aeson.KeyMap.KeyMap Value
persistent                >                      with: HashMap.HashMap k0 Value
persistent                >       Expected: HashMap.HashMap k0 Value
persistent                >         Actual: aeson-2.0.2.0:Data.Aeson.Types.Internal.Object
persistent                >     • In the second argument of ‘HashMap.lookup’, namely ‘o’
persistent                >       In the expression: HashMap.lookup "left" o
persistent                >       In the expression:
persistent                >         case HashMap.lookup "left" o of
persistent                >           Just v -> Left <$> loadConfig v
persistent                >           Nothing
persistent                >             -> case HashMap.lookup "right" o of
persistent                >                  Just v -> Right <$> loadConfig v
persistent                >                  Nothing
persistent                >                    -> fail "PersistConfig for Either: need either a left or right"
persistent                >    |
persistent                > 46 |         case HashMap.lookup "left" o of
persistent                >    |                                    ^
persistent                > 
persistent                > /tmp/stack-06165844a3e2e2dd/persistent-2.13.2.1/Database/Persist/Class/PersistConfig.hs:49:45: error:
persistent                >     • Couldn't match type: Data.Aeson.KeyMap.KeyMap Value
persistent                >                      with: HashMap.HashMap k1 Value
persistent                >       Expected: HashMap.HashMap k1 Value
persistent                >         Actual: aeson-2.0.2.0:Data.Aeson.Types.Internal.Object
persistent                >     • In the second argument of ‘HashMap.lookup’, namely ‘o’
persistent                >       In the expression: HashMap.lookup "right" o
persistent                >       In the expression:
persistent                >         case HashMap.lookup "right" o of
persistent                >           Just v -> Right <$> loadConfig v
persistent                >           Nothing
persistent                >             -> fail "PersistConfig for Either: need either a left or right"
persistent                >    |
persistent                > 49 |                 case HashMap.lookup "right" o of
persistent                >    |                                             ^
```

bergmark:
Can be closed

PR #1338, opened 2021-11-15
Fix for GHC 9.2 / template-haskell-2.18

brandon-leapyear:
Forked off of https://github.com/yesodweb/persistent/pull/1335 to fix `Database.Persist.TH` errors and CI errors. Feel free to merge this back into the other PR

PR #1351, opened 2022-01-15
aeson 2.0 support (only)

bergmark:
Issue: #1344

Cherry-picks from https://github.com/yesodweb/persistent/pull/1335 and https://github.com/yesodweb/persistent/pull/1338

seems like GHC 9.2 support for dependencies could take a while, so I would appreciate an earlier release with only aeson 2 support.

The Arbitrary Value instance in the pg tests was more restrictive than the one in aeson 2.0.3.0 so i newtyped it.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,5 @@
-resolver: lts-14.27
+resolver: nightly-2021-11-14
+
 packages:
   - ./persistent
   - ./persistent-sqlite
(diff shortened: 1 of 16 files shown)


parsonsmatt (approved):
Thank you so much! I'll get a release out soon.

stack.yaml
```diff
@@ -10,4 +11,7 @@ packages:
   - ./persistent-qq
 
 extra-deps:
-    - lift-type-0.1.0.0
+  - aeson-2.0.2.0
+  # https://github.com/yesodweb/shakespeare/pull/260
```
brandon-leapyear, line 15: This is extraneous, from a bad cherry-pick. Resolved here: https://github.com/yesodweb/persistent/pull/1338/files#diff-0de01f189d384bfbe69e806ef70b1a425b27394dbe2a9d0e62ffd010dda86c03L15

Issue #1358, opened 2022-02-09
Repeated migrations of an Int32 column in MySQL 8


paul-rouse:
*This issue describes the problem - PRs will follow both to extend the tests and to fix `persistent-mysql`.*

When running against MySQL 8.0.28, repeated, spurious migrations looking like this are produced for `Int32` columns:

```
ALTER TABLE `idempotency` CHANGE `field8` `field8` INT(11) NOT NULL
```

Internally, the behaviour for `Int64` is also dubious, but we get away with it at present - more about that below.

The problem is the "display width" - the "(11)" here - which has no effect on the stored data, and was [deprecated at MySQL 8.0.17](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html).  At some point (maybe 8.0.19) it stopped being included in the column type shown in the information_schema.  For reasons spelt out below, `findAlters` in `Database.Persist.MySQL` still expects to see "INT(11)", but now only gets "INT", so it generates a migration.  The attempted table alteration doesn't actually change anything on the server, so this recurs every time migration is calculated.

The code is wrong in two places:

  - When `parseColumnType` decodes an integer type, but the display width it expects is absent, it produces `SqlOther`.  Therefore, with MySQL 8, it gives `SqlOther "int"` and `SqlOther "bigint"` for 32- and 64-bit integers, instead of the correct decodings `SqlInt32` and `SqlInt64`.
  - `findAlters` compares the results of applying `showSqlType` both to the expected column and to the column type decoded from `parseColumnType`.  Unfortunately, `showSqlType SqlInt32` is "INT(11)" which does not match the "int" from `showSqlType (SqlOther "int")`.  This is unnecessary, even in MySQL 5.7, since eleven is the default display width for INT.  Indeed - inconsistently - `showSqlType SqlInt64` is simply "BIGINT", which is why we get away with matching it against `SqlOther "bigint"`, even though the latter isn't the intended decoding of the type.

This problem is not picked up by the migration idempotency test, since it has only a 64-bit integer column.  PR on its way to fix that first!

PR #1359, opened 2022-02-09
Further integer types in migration idempotency test

paul-rouse:
The extra fields in the migration idempotency test are intended to show the problem in #1358, and test the fix which will come separately.  Before that fix is applied, the test is *expected to fail* when run against MySQL 8.

I have kept this PR separate from the fix since it is run for other backends too, making it somewhat independent.

After submitting your PR:

- [x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts) - the current failure is the **expected** one!

### DIFF
--- a/persistent-test/ChangeLog.md
+++ b/persistent-test/ChangeLog.md
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 2.13.1.1
+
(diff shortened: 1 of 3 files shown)


paul-rouse:
The additional test passes now (after #1360), so merging.

PR #1360, opened 2022-02-09
Fix anomalies in migration of integer columns in MySQL 8 (#1358)

### LINKED ISSUE #1358: Repeated migrations of an Int32 column in MySQL 8

paul-rouse:
Fixes #1358.  I believe this remains compatible with older versions of MySQL, so should be a minor version bump.
 
After submitting your PR:

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -1023,8 +1023,10 @@ showSqlType :: SqlType
             -> String
 showSqlType SqlBlob    Nothing    _     = "BLOB"
 showSqlType SqlBlob    (Just i)   _     = "VARBINARY(" ++ show i ++ ")"
--- The non-default display width "(1)" is kept for now to avoid differences
--- between persistent versions running against older MySQL versions (eg 5.7):
+-- "tinyint(1)" has been used historically here.  In MySQL 8, the display width
+-- is deprecated, and in the future it may need to be removed here.  However,
+-- "(1)" is not the default in older MySQL versions, so for them omitting it
+-- would alter the exact form of the column type in the information_schema.
 showSqlType SqlBool    _          _     = "TINYINT(1)"
 showSqlType SqlDay     _          _     = "DATE"
 showSqlType SqlDayTime _          _     = "DATETIME"
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -827,9 +827,19 @@ data ColumnInfo = ColumnInfo
 -- @INFORMATION_SCHEMA@ tables.
 parseColumnType :: Text -> ColumnInfo -> ExceptT String IO (SqlType, Maybe Integer)
 -- Ints
-parseColumnType "tinyint" ci | ciColumnType ci == "tinyint(1)" = return (SqlBool, Nothing)
-parseColumnType "int" ci | ciColumnType ci == "int(11)"        = return (SqlInt32, Nothing)
-parseColumnType "bigint" ci | ciColumnType ci == "bigint(20)"  = return (SqlInt64, Nothing)
+-- The display width is deprecated and being removed in MySQL 8.X.  To be
+-- consistent with earlier versions, which do report it, accept either
+-- the bare type in `ciColumnType ci`, or the type adorned with the expected
+-- value for the display width (ie the defaults for int and bigint, or the
+-- value explicitly set in `showSqlType` for SqlBool).
+--
+parseColumnType "tinyint" ci
+    | ciColumnType ci == "tinyint" || ciColumnType ci == "tinyint(1)" = return (SqlBool, Nothing)
+parseColumnType "int" ci
+    | ciColumnType ci == "int"     || ciColumnType ci == "int(11)"    = return (SqlInt32, Nothing)
+parseColumnType "bigint" ci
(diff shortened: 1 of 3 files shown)


persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1013,10 +1023,12 @@ showSqlType :: SqlType
             -> String
 showSqlType SqlBlob    Nothing    _     = "BLOB"
 showSqlType SqlBlob    (Just i)   _     = "VARBINARY(" ++ show i ++ ")"
+-- The non-default display width "(1)" is kept for now to avoid differences
+-- between persistent versions running against older MySQL versions (eg 5.7):
 showSqlType SqlBool    _          _     = "TINYINT(1)"
 showSqlType SqlDay     _          _     = "DATE"
 showSqlType SqlDayTime _          _     = "DATETIME"
-showSqlType SqlInt32   _          _     = "INT(11)"
+showSqlType SqlInt32   _          _     = "INT"
```
parsonsmatt, line 1031: This comment doesn't appear to agree with the code change - can you clarify?

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -827,9 +827,19 @@ data ColumnInfo = ColumnInfo
...
+-- the bare type in `ciColumnType ci`, or the type adorned with the expected
+-- value for the display width (ie the defaults for int and bigint, or the
+-- value explicitly set in `showSqlType` for SqlBool).
+--
+parseColumnType "tinyint" ci
+    | ciColumnType ci == "tinyint" || ciColumnType ci == "tinyint(1)" = return (SqlBool, Nothing)
+parseColumnType "int" ci
+    | ciColumnType ci == "int"     || ciColumnType ci == "int(11)"    = return (SqlInt32, Nothing)
+parseColumnType "bigint" ci
+    | ciColumnType ci == "bigint"  || ciColumnType ci == "bigint(20)" = return (SqlInt64, Nothing)
```
parsonsmatt, line 841: does it maybe make sense to do a `Just _ <- Text.stripPrefix "tinyint" ` pattern? That would match `tinyint`, `tinyint(1)`, and `tinyint(2345)` etc.

parsonsmatt (approved):
A few questions but otherwise looks great, thank you!

persistent-mysql/Database/Persist/MySQL.hs
paul-rouse: It's not a great comment - I'll add a commit to try to improve it!  The point is I'm trying to avoid changing *anything* about what is put into the information schema in MySQL 5.7 and early members of the 8.X series.  In 5.7, `INT` gets a default display width `(11)` anyway, so I felt it would be clearer in the future to omit it here (as was already done for `BIGINT`).  However  `(1)` is not the default for `TINYINT`, and omitting it would make MySQL 5.7 add  `(4)` instead in the information schema.  It doesn't affect the data storage for the column itself, of course, but the different information schema might cause spurious migrations in `persistent`, or be incompatible with some external program the user might depend upon.

persistent-mysql/Database/Persist/MySQL.hs
paul-rouse: Again, I am trying to avoid changing anything for MySQL 5.7 and early 8.X.  These versions will always report the display width, so I thought it best to match it exactly, as before, to avoid including cases which were not included in `persistent-mysql-2.13.1.0`.  The addition is the matching of the bare type, eg `int`, which will only be reported by later 8.X versions

paul-rouse:
@parsonsmatt I will add a commit to improve the comment in the code, but would you like me to merge and release then?  I tend to leave things `persistent` to you, but quite happy to do this.

paul-rouse:
I've got this one - just remembered it's sitting around.

PR #1369, opened 2022-03-15
Fix JSON import

parsonsmatt:

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/TH.hs
+++ b/persistent/Database/Persist/TH.hs
@@ -81,6 +81,7 @@ import Data.Aeson
        , (.:)
        , (.:?)
        , (.=)
+       , withObject
        )
 #if MIN_VERSION_aeson(2,0,0)
 import qualified Data.Aeson.Key as Key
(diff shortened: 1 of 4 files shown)


persistent/Database/Persist/TH.hs
```diff
@@ -2646,16 +2647,17 @@ mkJSON mps (fixEntityDef -> def) = do
...
+        toKeyE = VarE 'Key.fromString
 #else
-    toKeyE <- [|pack|]
+        toKeyE = VarE 'pack
 #endif
-    dotEqualE <- [|(.=)|]
-    dotColonE <- [|(.:)|]
-    dotColonQE <- [|(.:?)|]
-    objectE <- [|object|]
-    withObjectE <- [|withObject|]
```
parsonsmatt, line 2658: Bug was here. `withObject` was not imported, so I think this turned into an `UnboundVarE` instead of a `VarE` that we want. Wow. Using the `VarE 'withObject` form forces the name to be in scope, which then becomes a `NameG` value.

@danbroooks TIL 😅

persistent/test/Database/Persist/THSpec.hs
```diff
@@ -24,7 +24,7 @@
 module Database.Persist.THSpec where
 
 import Control.Applicative (Const(..))
-import Data.Aeson hiding (Key)
+import Data.Aeson (decode, encode)
```
parsonsmatt, line 27: Surfaced problem in tests by doing this

danbroooks: Thanks for resolving this! Do you think it is worth adding a specific spec to cover imports required by users of persistent? Something like a test that only imports persistent, and defines some entities (some with derivied json instances) or do you think that might be overkill, something like:

```
module Database.Persist.TH.JsonEncodingSpec where

-- this test asserts this is the only import required
import Database.Persist.TH

JsonEncoding json
    name Text
    age  Int
    Primary name
    deriving Show Eq
JsonEncoding2 json
    name Text
    age Int
    blood Text
    Primary name blood
    deriving Show Eq
JsonEncMigrationOnly json
    name Text
    age  Int
    foo  Text MigrationOnly
|]
```

My original PR added to `JsonEncodingSpec` but that was already importing `Data.Aeson` :sweat_smile: what you have done here seems to be a more robust solution, but I wonder if we should also add a test similar to the above? Happy to do the PR for this if so :smile:

parsonsmatt: Yeah, that's a good idea.

PR #1370, opened 2022-03-15
Add spec to assert Persistent.TH is the only import required when defining entities

danbroooks:
Follow on from https://github.com/yesodweb/persistent/pull/1369. Though that PR adds some additional coverage to preventing this from happening again, this PR adds an additional spec to explicitly verify we can generate entities where `Data.Persistent.TH` is the only import required.

Screenshot from 2022-03-15 18-22-35

---

After submitting your PR:


parsonsmatt (approved):
Awesome 🙌🏻

PR #1372, opened 2022-03-16
Fix adding new column in persistent-mysql, and include MigrationTest

### LINKED ISSUE #1373: Adding a column can cause a run-time error in MySQL
In `persistent-mysql`, inserting an extra column before an existing one can result in migration trying to add that existing column again.  This produces a run-time "Duplicate column name" from MySQL.

This bug is demonstrated by PR #1372, which simply adds the standard `MigrationTest` to the set run for `persistent-mysql` - it isn't clear why it was omitted.

I will look at the code in a little while, unless someone else spots the problem first :smile:

paul-rouse:
Fixes #1373. **This PR has now been modified to include the fix**

Fixes the bug, and adds a test - an existing one, but it wasn't run for MySQL.

It **fails at present,** but that is the point!  There will be a separate issue in a minute.

After submitting your PR:

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -921,7 +921,7 @@ findAlters edef allDefs col@(Column name isNull type_ def gen _defConstraintName
     -- new fkey that didn't exist before
         [] ->
             case ref of
-                Nothing -> ([Add' col],[])
+                Nothing -> ([Add' col],cols)
                 Just cr ->
                     let tname = crTableName cr
                         cname = crConstraintName cr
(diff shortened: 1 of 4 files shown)


persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -921,7 +921,7 @@ findAlters edef allDefs col@(Column name isNull type_ def gen _defConstraintName
     -- new fkey that didn't exist before
         [] ->
             case ref of
-                Nothing -> ([Add' col],[])
+                Nothing -> ([Add' col],cols)
```
parsonsmatt, line 924: Nice

parsonsmatt (approved):

PR #1374, opened 2022-03-16
Increasing test coverage for errors thrown when parsing entity definitions

danbroooks:
Ups the test coverage in `QuasiSpec` to cover cases when there is an error, and make an assertion on the error message. 

The original motivation for this was discovering that this error here:

https://github.com/yesodweb/persistent/blob/aeaa4f61159e140fbf3a45fc2b59309b0bb4b53f/persistent/Database/Persist/Quasi/Internal.hs#L1155-L1157

Is actually `show`ing all the `UnboundField`s as illustrated in this commit prior to fixing:

https://github.com/yesodweb/persistent/pull/1374/commits/89b34a9b5e468b18b030591843c380a9ee60bb20#diff-31a6cff4e15cd1978c97554cd4fae33101b142ab0e6a6e03f427458d2851b26bR457

This PR corrects this, and extracts a function to generate the error message in a more type safe way.

Further to this I have added some similar tests around other errors in this module which do not appear to have coverage.

----

After submitting your PR:

### DIFF
--- a/persistent/test/Database/Persist/QuasiSpec.hs
+++ b/persistent/test/Database/Persist/QuasiSpec.hs
@@ -453,8 +453,10 @@ User
                 let [user] = parse lowerCaseSettings definitions
                     uniques = entityUniques (unboundEntityDef user)
                     [dbNames] = fmap snd . uniqueFields <$> uniques
-                    errMsg =
-                        "Unknown column in unique constraint: \"emailSecond\" [UnboundFieldDef {unboundFieldNameHS = FieldNameHS {unFieldNameHS = \"name\"}, unboundFieldNameDB = FieldNameDB {unFieldNameDB = \"name\"}, unboundFieldAttrs = [], unboundFieldStrict = True, unboundFieldType = FTTypeCon Nothing \"Text\", unboundFieldCascade = FieldCascade {fcOnUpdate = Nothing, fcOnDelete = Nothing}, unboundFieldGenerated = Nothing, unboundFieldComments = Nothing},UnboundFieldDef {unboundFieldNameHS = FieldNameHS {unFieldNameHS = \"emailFirst\"}, unboundFieldNameDB = FieldNameDB {unFieldNameDB = \"email_first\"}, unboundFieldAttrs = [], unboundFieldStrict = True, unboundFieldType = FTTypeCon Nothing \"Text\", unboundFieldCascade = FieldCascade {fcOnUpdate = Nothing, fcOnDelete = Nothing}, unboundFieldGenerated = Nothing, unboundFieldComments = Nothing}]\"UniqueEmail\" []"
+                    errMsg = unwords
+                        [ "Unknown column in \"UniqueEmail\" constraint: \"emailSecond\""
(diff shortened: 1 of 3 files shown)


persistent/test/Database/Persist/QuasiSpec.hs
```diff
@@ -145,6 +149,11 @@ spec = describe "Quasi" $ do
                         ]
                     )
 
+        it "should error if quotes are unterminated" $ do
+            evaluate (parseLine "    \"foo bar")
+                `shouldThrow`
+                    errorCall "Unterminated quoted string starting with foo bar"
```
danbroooks, line 155: Initially I implemented this like the `unterminated parens` test:

```hs
        it "should error on malformed input, unterminated parens" $ do
            let definitions = [st|
User
    name Text
    age  (Maybe Int
|]
            let [user] = parse lowerCaseSettings definitions
            evaluate (unboundEntityDef user)
                `shouldThrow`
                    errorCall "Unterminated parens string starting with Maybe Int"
```

However I wasn't able to come up with a realistic looking example, the following does not appear to trigger this error:

```hs
        it "should error on malformed input, unterminated quotes" $ do
            let definitions = [st|
User
    name Text
    created UTCTime default="NOW()
|]
            let [user] = parse lowerCaseSettings definitions
            evaluate (unboundEntityDef user)
                `shouldThrow`
                    errorCall "error text goes here"
```

However this did not seem to trip up the parser. It seems like the parser is expecting a token that _begins_ with a quote, as illustrated here. What is this required for in persistent? I couldnt think of anything besides the above and was unable to find examples of this in the repo.

Furthermore, I deleted this line from the tokenizer:

https://github.com/yesodweb/persistent/blob/aeaa4f61159e140fbf3a45fc2b59309b0bb4b53f/persistent/Database/Persist/Quasi/Internal.hs#L207

And the tests carried on working fine... so I am wondering if this something that was added that could possibly be made redundant now?

I am not removing that line in this PR though I feel like it's asking for trouble :D

danbroooks: So... :smile:  this is actually what I have been tinkering around with this weekend, with some success:

https://github.com/yesodweb/persistent/compare/master...danbroooks:parser-refactor-spike

:point_up: the above seems to work against the current set of unit tests, I was going to open an issue to see if there was an appetite for a change like this (as obviously its a fairly big shift in style for the quasi parser currently), I don't see why not if it doesn't bring in any extra dependencies, I guess... it's just if test coverage here is sufficient enough for a change like this (though it seems so to me?)

Also, you may have your own vision for the direction you want to take the parser in, I'd be happy to work on some changes like that steered in a particular direction, the above branch is just my take on it after a weekend hacking away at it :smile:

In terms of this comment + PR, I guess these tests at least model this current behaviour, even if it is potentially unreachable/unused

parsonsmatt: Yeah! That's awesome. I'd be happy to have that in place.

Switching to a more reasonable parser combinator style would be awesome, even if it is home-baked. I'm fine bringing in `megaparsec`, too.

PR #1375, opened 2022-03-17
Fix question marks in strings #1371

### LINKED ISSUE #1371: Question marks inside of strings cause RawSQL to be sad
Suppose you're in a dark room and you're not sure who you're interacting with

```haskell
[sqlQQ| SELECT 'hello world?' |]
```

This gives a `FormatError` because we're trying to split on the `?`, even though the `?` is *not* a parameter to substitute syntactically, but a character in the string.

We can work around this by providing the value as a query parameter:

```haskell
[sqlQQ| SELECT #{PersistText "hello world?"} |]
```

parsonsmatt:
Closes #1371 

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Sql/Raw/Internal.hs
+++ b/persistent/Database/Persist/Sql/Raw/Internal.hs
@@ -1,4 +1,6 @@
-module Database.Persist.Sql.Raw where
+-- | This module is a internal. Breaking changes to the API of this module will
+-- not be reflected in a major version bump.
+module Database.Persist.Sql.Raw.Internal where
 
 import Control.Exception (throwIO)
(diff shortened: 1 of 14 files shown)


persistent/Database/Persist/Sql/Raw/Internal.hs
```diff
@@ -1,4 +1,6 @@
-module Database.Persist.Sql.Raw where
+-- | This module is a internal. Breaking changes to the API of this module will
```
arrowd, line 1: Extra `a`?

Vlix:
This PR just exports the `Raw` and `Internal` modules, what happened to the "question marks in strings"?

PR #1379, opened 2022-04-04
Support NoFieldSelectors and DuplicateRecordFields

fumieval:
* Getters are defined in terms of explicit record pattern match instead of a field selector function
* It defines record updates verbosely in order to disambiguate duplicate fields

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/TH.hs
+++ b/persistent/Database/Persist/TH.hs
@@ -78,10 +78,10 @@ import Data.Aeson
        , Value(Object)
        , eitherDecodeStrict'
        , object
+       , withObject
        , (.:)
        , (.:?)
        , (.=)
-       , withObject
        )
 #if MIN_VERSION_aeson(2,0,0)
 import qualified Data.Aeson.Key as Key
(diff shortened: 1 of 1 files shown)


persistent/Database/Persist/TH.hs
```diff
@@ -78,10 +78,10 @@ import Data.Aeson
        , Value(Object)
        , eitherDecodeStrict'
        , object
+       , withObject
```
fumieval, line 81: This change is due to stylish-haskell

parsonsmatt:
This is a great improvement - I've definitely run into issues before where you'd have a model file like:

```
User
    profileName Text

UserProfile
    name Text
```

and it would fail to compile.

Thanks for the PR!

parsonsmatt:
Would you be willing to add a test module here? Mostly so we can avoid breaking this with a regression in the future

Probably something like:

```haskell
#if __GLASGOW_HASKELL__ >= 902
{-# LANGUAGE NoFieldSelectors #-}
{-# LANGUAGE DuplicateRecordFields #-}
#endif    

module NoFieldSelectorTest where

#if __GLASGOW_HASKELL__ >= 902

import ...

mkPersist [persistLowerCase|

something that triggers a bug

|]

spec :: Spec
spec = it "compiles" True

#else

spec :: Spec 
spec = do
    it "only works with GHC 9.2 or greater" $ do
        pendingWith "only works with GHC 9.2 or greater"    

#endif
```

fumieval:
Added (using your example as-is). Unfortunately it turns out that disambiguating fields isn't enough to solve the User vs UserProfile problem because it also defines type constructors (`UserProfile`).

parsonsmatt:
Released as `persistent-2.13.3.4`, thank you!

Issue #1323, opened 2021-10-01
Create a Unique constructor for Primary keys


parsonsmatt:
[Initial discussion here](https://github.com/yesodweb/persistent/issues/1239#issuecomment-932365831)

A `Unique rec` has two implied properties:

- They are *natural* - that is, it exists on the Haskell datatype.
- They are *distinct* from the primary key - even if you have a natural primary key.

The latter property is Bad, and this ticket intends to fix it.

The code example I gave in the prior ticket is reproduced here:

```haskell
mkPersist sqlSettings [persistLowerCase|

User
    name Text
    Primary name
|]

-- ====>

instance PersistEntity User where
    data Unique User 
        = UniqueUserKey (Key User)
```

This would allow us to write, like, `upsertBy (UniqueUserKey userKey)`.

This would be a *breaking change*, since a lot of code that previously assumed a certain amount of unique keys would no longer work. For example, `upsert` has an `OnlyOneUniqueKey` constraint. This would need to change.

Issue #985, opened 2019-11-07
Can't refer to the primary key in a Unique key


parsonsmatt:
Consider:

```
Group
  name Text

User
  Id String
  group GroupId

  UniqueUserGroup id group
```

This fails because the `id` column is unknown.

parsonsmatt:
Sigh. Impossible to fix without altering the signature of `persistUniqueKeys :: Entity record -> [Unique record]`, which is a breaking change and a major version bump.

parsonsmatt:
I think this is pointing at a larger problem that needs to be addressed in the API.

The current design is great for *surrogate* primary keys, but it kinda falls apart when you have a natural primary key (designated with `Primary`), and you can't make redundant `Unique` constraints that refer to the `id` field.

If you try to make `persistUniqueKeys` require an `Entity record` instead of a `record`, then this messes up `defaultPutMany`.

Calling `insert` with a custom Primary key breaks, so you have to do `insertKey`. The same asymmetry applies to `insertMany` vs `insertEntityMany`.

I think we need to reconcile this and make it type-safe once-and-for-all.

parsonsmatt:
Was fixed in #1383

PR #1383, opened 2022-04-12
`Primary` key implies a `Unique` constraint

parsonsmatt:

After submitting your PR:


persistent-test/src/PersistUniqueTest.hs
```diff
@@ -6,43 +6,139 @@ import Init
...
-    Just _ <- checkUnique fo -- conflicts with itself
-
-    let fo' = Fo (f + 1) b
-    Just _ <- checkUnique fo' -- conflicts with fo
-    Nothing <- checkUniqueUpdateable $ Entity k fo' -- but fo can be updated to fo'
-
-    let fo'' = Fo (f + 1) (b + 1)
-    insert_ fo''
-    Just (UniqueBar conflict) <- checkUniqueUpdateable $ Entity k fo'' -- fo can't be updated to fo''
-    conflict @== b + 1
```
parsonsmatt, line 48: I think the old `insertUnique` code was actually broken on this, since it would be totally fine doing an `insert` based on the `UniqueBar` being actually unique, but then fail with the `Primary` key constraint violation.

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -1067,7 +1074,7 @@ takeId ps entityName texts =
 --
 -- @since.2.13.0.0
 data UnboundCompositeDef = UnboundCompositeDef
-    { unboundCompositeCols :: [FieldNameHS]
+    { unboundCompositeCols :: NonEmpty FieldNameHS
```
parsonsmatt, line 1077: Might as well - we can't have a composite key without at least one field.

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -984,10 +985,16 @@ takeConstraint ps entityName defs (n :| rest) =
...
-                }
+            let
+                unboundComposite =
+                    takeComposite (unboundFieldNameHS <$> defs) rest
+            in
+                mempty
+                    { entityConstraintDefsPrimaryComposite =
+                        SetOnce unboundComposite
+                    , entityConstraintDefsUniques =
+                        Just $ pure $ compositeToUniqueDef entityName defs unboundComposite
```
parsonsmatt, line 996: This is the real change. We're adding a `UniqueDef` based on the composite primary key.

persistent/Database/Persist/Sql/Internal.hs
```diff
@@ -81,7 +81,7 @@ mkColumns
     -> BackendSpecificOverrides
     -> ([Column], [UniqueDef], [ForeignDef])
 mkColumns allDefs t overrides =
-    (cols, getEntityUniques t, getEntityForeignDefs t)
+    (cols, getEntityUniquesNoPrimaryKey t, getEntityForeignDefs t)
```
parsonsmatt, line 84: In order to preserve migration compatibility, we don't return the composite primary key in the `UniqueDef` for migrations.

persistent/Database/Persist/TH.hs
```diff
@@ -374,18 +374,18 @@ liftAndFixKeys mps emEntities entityMap unboundEnt =
                             , "\n\nYou can use the References keyword to fix this."
                             ]
                     | otherwise =
-                        zip (fmap (withDbName fieldStore) fieldNames) parentKeyFieldNames
+                        zip (fmap (withDbName fieldStore) fieldNames) (toList parentKeyFieldNames)
                   where
                     parentKeyFieldNames
-                        :: [(FieldNameHS, FieldNameDB)]
+                        :: NonEmpty (FieldNameHS, FieldNameDB)
```
parsonsmatt, line 380: just pushing this as far as we can.. even if it's `toList`ed about two lines above.

persistent/test/Database/Persist/QuasiSpec.hs
```diff
@@ -394,6 +394,18 @@ User
...
+                        { uniqueHaskell =
+                            ConstraintNameHS "UserPrimaryKey"
+                        , uniqueDBName =
+                            ConstraintNameDB "primary_key"
+                        , uniqueFields =
+                            pure (FieldNameHS "ref", FieldNameDB "ref")
+                        , uniqueAttrs =
+                            []
+                        }
+                    ]
```
parsonsmatt, line 408: super simple test for the Quasi parser

parsonsmatt (commented):
overall very happy with how easy this was to knock out

parsonsmatt:
Fixes #1323

parsonsmatt:
Fixes #985

Issue #1377, opened 2022-03-24
Feature Request: Add ability to turn off foreign key constraints


erikd:
I work on a project that uses Persistent on a relatively large database (40 or so tables and total database size of 120GB).

I recently ran a test where I generated the database schema using Persistent and then manually deleted all foreign key constraints of the form:
```
ALTER TABLE "slot_leader" ADD CONSTRAINT "slot_leader_pool_hash_id_fkey" FOREIGN KEY("pool_hash_id")
    REFERENCES "pool_hash"("id") ON DELETE CASCADE  ON UPDATE RESTRICT ;
```
There were 64 instance of these and I found that with the removal of these constraints alone, I was able to my populate the database **2-4 times faster** than before (-23 days for a full db populate instead of 6+ days).

That means we would like to a directive to turn off the generation of foreign key constraints.  This would appear in the Haskell schema as `NoForeignKeyConstraint` and be used much like `OnDeleteCascade`.

I am fully aware that this shifts a lot of onus for correct handling of deletions, updates and other issues to the Persistent user. Is there a better way to do this? Would a high quality change implementing this be accepted?

parsonsmatt:
You can do this by running `getMigration :: Migration -> SqlPersistT m [Text]` and then running `partition (\sql -> "ADD CONSTRAINT` isInfixOf` sql && "REFERENCES" `isInfixOf` sql)` on that. Then you can get both the foreign key migrations as well as the other migrations in separate `[Text]` and run them how you'd like.

erikd:
Thanks @parsonsmatt . Yes, you are right, I can do this myself in my code. No changes needed in Persistent.

PR #1431, opened 2022-10-16
Fix typo in FieldAttrNoreference comment

kderme:
Typo fix.
I guess `FieldAttrNoreference` is an alternative answer to https://github.com/yesodweb/persistent/issues/1377

cc @erikd


parsonsmatt:
good catch, thank you!

PR #1437, opened 2022-11-04
Add existsBy

blx:
This adds `existsBy` in `PersistUniqueRead`, as a more efficient version of `isJust <$> getBy uniq`.

The intent is to provide an (a) more convenient to type and (b) more efficient way to query existence of a row by unique key.

The SQL is a combination of the `exists` and `getBy` impls, and looks like `SELECT EXISTS (SELECT 1 FROM ...constraints...)`.

The performance difference shouldn't be large, but conceivably on wider tables (or with heavy columns eg. JSON or binary blobs), `existsBy` would require less serialization (just a single response column) and may allow the query planner to do an index-only plan.

What do you think?

---

After submitting your PR:

### DIFF
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -136,7 +136,7 @@ import Data.Bits (shiftR)
 import Data.Bson (ObjectId(..))
 import qualified Data.ByteString as BS
 import Data.Conduit
-import Data.Maybe (fromJust, mapMaybe)
+import Data.Maybe (fromJust, isJust, mapMaybe)
 import Data.Monoid (mappend)
 import qualified Data.Pool as Pool
 import qualified Data.Serialize as Serialize
@@ -604,6 +604,8 @@ instance PersistUniqueRead DB.MongoContext where
       where
         t = entityDef $ Just rec
         rec = dummyFromUnique uniq
+
+    existsBy uniq = isJust <$> getBy uniq
 
 instance PersistUniqueWrite DB.MongoContext where
     deleteBy uniq =
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -136,7 +136,7 @@ import Data.Bits (shiftR)
 import Data.Bson (ObjectId(..))
 import qualified Data.ByteString as BS
 import Data.Conduit
-import Data.Maybe (fromJust, isJust, mapMaybe)
+import Data.Maybe (fromJust, mapMaybe)
 import Data.Monoid (mappend)
 import qualified Data.Pool as Pool
 import qualified Data.Serialize as Serialize
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -605,8 +605,6 @@ instance PersistUniqueRead DB.MongoContext where
         t = entityDef $ Just rec
         rec = dummyFromUnique uniq
 
-    existsBy uniq = isJust <$> getBy uniq
-
 instance PersistUniqueWrite DB.MongoContext where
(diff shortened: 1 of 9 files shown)


persistent-mongoDB/Database/Persist/MongoDB.hs
```diff
@@ -605,6 +605,8 @@ instance PersistUniqueRead DB.MongoContext where
         t = entityDef $ Just rec
         rec = dummyFromUnique uniq
 
+    existsBy uniq = isJust <$> getBy uniq
```
blx, line 608: I'm not very familiar with MongoDB and saw in the readme that the persistent-mongoDB is "on hiatus", so here's a valid impl that doesn't have any performance benefits.

persistent/Database/Persist/Sql/Orphan/PersistUnique.hs
```diff
@@ -8,18 +8,25 @@ import Control.Exception (throwIO)
...
-import Database.Persist.Sql.Util (dbColumns, parseEntityValues, updatePersistValue, mkUpdateText')
+import Database.Persist.Sql.Raw
+import Database.Persist.Sql.Types.Internal
+import Database.Persist.Sql.Util
+       ( dbColumns
+       , mkUpdateText'
+       , parseEntityValues
+       , parseExistsResult
+       , updatePersistValue
+       )
```
blx, line 29: imports diff here was just stylish-haskell

blx:
CI is failing with a bunch of
```
Database/MongoDB/Query.hs:1692:32: error:
Error:     Not in scope: type constructor or class ‘MonadFail’
     |
1692 | aggregateCursor :: (MonadIO m, MonadFail m) => Collection -> Pipeline -> AggregateConfig -> Action m Cursor
     |                                ^^^^^^^^^

Database/MongoDB/Query.hs:1711:18: error:
Error:     Not in scope: type constructor or class ‘MonadFail’
     |
1711 |   :: (MonadIO m, MonadFail m)
     |                  ^^^^^^^^^
```
which looks like it's the `mongoDB` third-party package which I haven't touched here.

I wonder, if I hadn't edited anything in `persistent-mongoDB` here, would the CI have passed with pre-existing cached artifacts, or something, and not have run into any issues here?

persistent/Database/Persist/Sql/Orphan/PersistUnique.hs
```diff
@@ -110,11 +117,34 @@ instance PersistUniqueRead SqlBackend where
         t = entityDef $ dummyFromUnique uniq
         toFieldNames' = toList . fmap snd . persistUniqueToFieldNames
 
+    existsBy uniq = do
```
blx, line 120: Just realized that maybe this could be simplified to something like
```hs
existsBy uniq = exists $ persistUniqueToValues uniq
```
if
* we had access to `PersistQueryRead` here, for `exists`
* and the signature of `exists`, or possibly an inner helper, was adapted to also accept `[PersistValue]`, which is what is obtained in `exists` already after calling `getFiltsValues`

parsonsmatt (changes requested):
Let's put a default impl on the class so it's a safer for backwards compatibility

persistent/ChangeLog.md
```diff
@@ -2,13 +2,18 @@
...
 
+## 2.14.5.0
+
+* [#1437](https://github.com/yesodweb/persistent/pull/1437)
+    * Add `existsBy` to `PersistUniqueRead`
+
 ## 2.14.4.4
 
-* [#1460] https://github.com/yesodweb/persistent/pull/1460
+* [#1460](https://github.com/yesodweb/persistent/pull/1460)
```
blx, line 16: This matches the format in the rest of the changelog, where the PR number links to the link, but the URL itself is not shown. I can revert these changes if you want.

parsonsmatt: That's great - thank you!

parsonsmatt (approved):
This looks great, thank you so much!

PR #1440, opened 2022-11-17
NFData PersistValue

avieth:
Fully evaluating a `PersistValue` is important for gathering correct metrics. E.g. to get an accurate estimation of how long the SQL itself takes to run, it's essential to force the `[PersistValue]` which is used to construct the statement _before_ starting the timer.

### DIFF
--- a/persistent/Database/Persist/PersistValue.hs
+++ b/persistent/Database/Persist/PersistValue.hs
@@ -11,6 +11,7 @@ module Database.Persist.PersistValue
     , LiteralType(..)
     ) where
 
+import Control.DeepSeq
 import qualified Data.ByteString.Base64 as B64
 import qualified Data.Text.Encoding as TE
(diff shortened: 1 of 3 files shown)


persistent/Database/Persist/PersistValue.hs
```diff
@@ -70,6 +71,25 @@ data PersistValue
     -- @since 2.12.0.0
     deriving (Show, Read, Eq, Ord)
 
+instance NFData PersistValue where
```
parsonsmatt, line 74:
```suggestion
-- |
-- @since 2.14.4.0
instance NFData PersistValue where
```

parsonsmatt (commented):
Please add the changelog and version bump and I'll have it released soon

thanks!

avieth:
> Please add the changelog and version bump and I'll have it released soon

Awesome! Changelog updated and `@since` added.

Issue #1454, opened 2022-12-19
Namespaces in models


arrowd:
Many ERP systems allow its user to write applications on top of them. Applications may store their stuff in the system's database, which means that each application has its own set of entities. If we're to write such a system using Persistent, how can we handle a case when two unrelated applications create entities with same names?

In other words, if we have two Haskell modules `M1` and `M2`, both of which containing something like

```haskell
share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
User
   name Text
|]
```

then running

```haskell
import qualified M1
import qualified M2

insert $ M1 "User1"
insert $ M1 "User2"
```

will place the data into the same table.

Would it make sense to implement some sort of namespacing in Persistent? If yes, what approach should be taken?
I'm thinking about taking a fully qualified module name (like, `packagename-1.2.3.Database.Persist.Foo`), computing a hash based on it and appending it to the table name.

arrowd:
This issue [1] is highly related, although the solution proposed for it is only a safe belt preventing data mixing for some cases.

[1] https://github.com/yesodweb/persistent/issues/576

parsonsmatt:
I think the proper approach here is to support multiple database schema: #93 

Then a library can use a specific schema to separate the tables it uses. Oof, I do feel bad linking a 10 year old ticket...

A work-around until then is for the library to export the `[UnboundEntityDef]` that the `QuasiQuoter` generates:

```haskell
librarySchema :: [UnboundEntityDef]
librarySchema = [persistLowerCase|

User
    name Text
|]
```

A client of the API can then call `mkPersist` to generate the models. But - they also have the ability to map over the `UnboundEntityDef`, prefixing all of the `EntityNameHS` with a tag - I think with `transformBi (\(EntityNameHS n) -> (EntityNameHS (prefix <> n)) :: UnboundEntityDef -> UnboundEntityDef` it shouldn't be too tricky to write.

arrowd:
Are database schemas a standard concept? Can we expect it to work for most SQL backends (existing and not yet written)?

> (prefix <> n)

Generating that `prefix` is also a problem. How'd we generate an unique one and be sure that it won't interfere with some existing table? The same question goes for schema names. Or we just offload that problem to the user?

parsonsmatt:
[...]

PR #1466, opened 2023-01-17
WIP: Schema support

arrowd:
This is a first take on #1454 

It adds `entitySchema` field and teach sqlite backend to use it. I'm a bit worried that I had to import Internal modules, so I decided to ask for early feedback.


arrowd:
At this stage the schema support is added to Sqlite and PostgreSQL backend. There is a simple test that passes for these backends too.

thanhvh2205:
@arrowd  Currently, we have a hardcode in the `doesTableExist` function. I also need help creating a new table with the same name as the existing one but in a new schema. Your PR can be added to the `doesTableExist` function to support the schema name. What do you think?

thanhvh2205:
@arrowd Currently, we have a hardcode `doesTableExist` function. I also need help creating a new table with the same name as the existing one but in a new schema. I think your PR can be added to the `doesTableExist` function. What do you think?

arrowd:
I'm not sure what function you're talking about. I did adapt `doesTableExist` for PostgreSQL: https://github.com/yesodweb/persistent/pull/1466/files#diff-3a6af517565f4d723a74260939388b56ca107830b6a9ae68db5a877ca084f520L598

thanhvh2205:
@arrowd Sorry, you are right. Your PR will be able to solve my current problem. Can you tell me when the PR will be ready? I hope it can be merged soon.

arrowd:
At the moment this PR lacks 2 things:

1. MySQL and MongoDB changes.
2. A bit more tests

Once this is done a review from Persistent developers would be required, which may take some time.

persistent/Database/Persist/Types/Base.hs
```diff
@@ -131,6 +131,8 @@ data EntityDef = EntityDef
     -- ^ The name of the entity as Haskell understands it.
     , entityDB      :: !EntityNameDB
     -- ^ The name of the database table corresponding to the entity.
+    , entitySchema  :: !(Maybe Text)
+    -- ^ The schema name of the database table.
```
parsonsmatt, line 135: Let's make another `newtype SchemaNameDB` for this - will make it easier to disambiguate what is and isn't a schema vs any other bit of text

persistent/Database/Persist/Sql/Types.hs
```diff
@@ -34,6 +34,7 @@ data Column = Column
 -- | This value specifies how a field references another table.
 --
 -- @since 2.11.0.0
+-- TODO: what about schema name there?
```
parsonsmatt, line 37: This does seem important to resolve. Unfortunately, the `ColumnReference` type is exposed, meaning that we will incur a major version bump if we modify it.

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt, line 1: Since `ColumnReference` isn't supported yet, this addition will fail for things that need to use it - which is specifying foreign key relationships in migrations. I suspect that a test that works on this schema would fail:

```haskell
  Foo schema=foo_schema
    name Int

  Bar 
    foo FooId 
|]
```

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1094,12 +1110,16 @@ findAlters defs edef col@(Column name isNull sqltype def _gen _defConstraintName
...
+                                            && getEntityDBName ent == crTableName colRef)
+                                        defs
+                        (schema, _) = schemaNamePair edef
+                    in case mbRefdef of
                         Just refdef
                             | Just _oldName /= fmap fieldDB (getEntityIdField edef)
                             ->
                             [AddReference
-                                (crTableName colRef)
+                                (schema, crTableName colRef)
```
parsonsmatt, line 1122: Oh, nice, you handle this appropriately here. Well done.

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -731,6 +738,7 @@ mayDefault def = case def of
     Just d -> " DEFAULT " <> d
 
 type SafeToRemove = Bool
+type SchemaEntityName = (Text, EntityNameDB)
```
parsonsmatt, line 741: I'd prefer to see a `newtype` or a `data` with named fields to make this more explicit

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -678,6 +686,16 @@ sqlUnique (UniqueDef _ cname cols _) = T.concat
...
 
+schemaNamePair :: EntityDef -> SchemaEntityName
+schemaNamePair ent = schemaNamePair' (getEntityDBSchema ent) (getEntityDBName ent)
+
+schemaNamePair' :: Maybe Text -> EntityNameDB -> SchemaEntityName
+schemaNamePair' mbSchema entName = case mbSchema of
+    Nothing -> ("", entName)
+    Just "main" -> ("", entName)
+    Just "temp" -> ("temp", entName)
+    Just schema -> ("", EntityNameDB $ (schema <> "_") <> unEntityNameDB entName)
```
parsonsmatt, line 697: Intriguing - so sqlite only has two schema, `main` and `temp`? And we mimic that by using `schema_table_name`? I'd be inclined to use `__` (two underscore) to make the distinction more clear - that way you can see `foo__bar_baz` and know that `foo` is the supposed schema, rather than `foo_bar_baz` potentially being either `BarBaz schema=foo` or `FooBarBaz` (no schema).

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1320,24 +1343,33 @@ showAlter table (Update' c s) = T.concat
...
     ]
 
+schemaNamePair :: EntityDef -> SchemaEntityName
+schemaNamePair ent = schemaNamePair' (getEntityDBSchema ent) (getEntityDBName ent)
+
+schemaNamePair' :: Maybe Text -> EntityNameDB -> SchemaEntityName
+schemaNamePair' mbSchema entName = case mbSchema of
+    Nothing -> ("", entName)
+    Just "public" -> ("", entName)
+    Just schema -> (schema, entName)
```
parsonsmatt, line 1371: Uses of this function case on `""` and treat it differently - I think that's a good reason to keep this as a `Maybe Text` instead of a `Text`.

I'm also curious about treating `"public"` differently here - can you explain why we don't just keep the `public` prefix?  Presumably, the user must have specified that - ie `Foo schema=public` - and I'm hesitant to override a user choice.

parsonsmatt (commented):
This is a fantastic start! Apologies for the delayed review.

With the current change set, this is a minor bump to `persistent`, which will require new lower bounds on `persistent-sqlite` and `persistent-postgresql` to take advantage of the feature.

arrowd:
I guess this now can be closed in favor of https://github.com/yesodweb/persistent/pull/1561

PR #1468, opened 2023-01-23
Remove extraneous `map toPersistValue` call.

arrowd:
The list produced by `toPersistFields entity` is already of type `[PersistValue]`. This can be checked by putting a type hole inside, which will reveal that `toPersistValue` call in this context has `PersistValue -> PersistValue` type and boils down to `id`.

- [X] Ran `stylish-haskell` on any changed files.
- [X] Adhered to the code style (see the `.editorconfig` file for details)

After submitting your PR:

- [X] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,5 +1,11 @@
 # Changelog for persistent
 
+## unreleased
+
+* [#1460] https://github.com/yesodweb/persistent/pull/1468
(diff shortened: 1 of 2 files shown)


parsonsmatt:
Good catch, thank you! Please add a changelog entry, under a new `unreleased` section (since it isn't a bugfix or feature, just a refactor)

PR #1476, opened 2023-03-02
Avoid parse error occurred when the field name generated by TH matches any of Haskell keywords

ccycle:
To avoid the problem that GHC fails compiling due to parse error when the field name generated by Template Haskell matches any of Haskell keywords, this PR fixes `mkRecordName` to suffix `_` if the field name matches any of keywords.

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/TH.hs
+++ b/persistent/Database/Persist/TH.hs
@@ -3100,7 +3100,7 @@ mkEntityLensName mps entDef fieldDef =
 
 mkRecordName :: MkPersistSettings -> Maybe Text -> EntityNameHS -> FieldNameHS -> Name
 mkRecordName mps prefix entNameHS fieldNameHS =
-    mkName $ T.unpack $ fromMaybe "" prefix <> lowerFirst recName
+    mkName $ T.unpack . avoidKeyword $ fromMaybe "" prefix <> lowerFirst recName
   where
     recName :: Text
     recName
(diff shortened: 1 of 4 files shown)


persistent/Database/Persist/TH.hs
```diff
@@ -3115,6 +3115,17 @@ mkRecordName mps prefix entNameHS fieldNameHS =
...
+    avoidKeyword :: Text -> Text
+    avoidKeyword name = if name `elem` keywords then name ++ "_" else name
+
+    keywords :: [Text]
+    keywords =
+      ["case","class","data","default","deriving","do","else"
+      ,"if","import","in","infix","infixl","infixr","instance","let","module"
+      ,"newtype","of","then","type","where","_"
+      ,"foreign"
+      ]
```
parsonsmatt, line 3127: Lots of list lookups like this is pretty inefficient - would prefer to see a `Set` lookup
```suggestion
    avoidKeyword name = if name `Set.member` haskellKeywords then name ++ "_" else name

haskellKeywords :: Set Text
haskellKeywords = Set.fromList
      ["case","class","data","default","deriving","do","else"
      ,"if","import","in","infix","infixl","infixr","instance","let","module"
      ,"newtype","of","then","type","where","_"
      ,"foreign"
      ]
```

Floating it to the top-level ensures we aren't constructing the `Set` for each call to `mkRecordName`, though I would guess GHC is capable of performing that optimization on it's own

parsonsmatt (changes requested):
Thanks for writing this up! One minor performance quibble.

At a future point, it would be nice to customize this behavior.

Since anyone writing `type` as a field name currently will receive GHC error, I think we can safely call this a bugfix. Mind doing a patch version bump and adding a changelog notice?

PR #1207, opened 2021-03-16
async exception issue

parsonsmatt:
Before submitting your PR, check that you've:

- [x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

- [x] Update the Changelog.md file with a link to your PR
- [x] Bumped the version number if there isn't an `(unreleased)` on the Changelog
- [x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

test initially thanks to @codygman 

investigating #1199 
maybe even fixes #1199

PR #1482, opened 2023-03-14
Retryable transactions + async exception handling

jship:
The main change in this PR is adding support for retryable transactions. For example, when using the various `runSqlPool*` functions at a repeatable-read or serializable isolation level, application authors need a means to retry transactions that encounter serialization failures. Without official support for this in `persistent`, `runSqlPool*` users could technically still manually catch a serialization failure exception around the whole `runSqlPool*` and retry the transaction, but this is a nonstarter as the connection would need to be returned then reacquired from the pool. To support retryable transactions, a `runSqlPoolWithExtensibleHooksRetry` function was added that takes in an exception predicate to determine retrying the transaction. The existing `runSqlPoolWithExtensibleHooks` is now implemented in terms of the new function.

Another change in this PR is around `runSqlPoolWithExtensibleHooks`'s async exception handling. The previous version of this function would not run the `runOnException` hook when the user-specified database action was aborted via async exception, as async exceptions were ignored entirely due to use of `unliftio`'s `catchAny`. This is not as problematic as it sounds on the surface: if an async exception came in, the enclosing `withResource` on the pool would catch it and terminate the connection. For PostgreSQL, when the connection is terminated, the database discards whatever changes were made in the transaction even though there was no explicit rollback. With the change in this PR, if users have custom logic defined in their `runOnException` hook on top of just rolling back the transaction, they should now be able to rely on `persistent` to execute this hook when the user-specified database action encounters _any_ type of exception.

There were also multiple spots where the masking state was being restored (basically everything was in a `restore` except for the installation of the `catchAny` handler). The masking has been changed in this PR such that `alterBackend` is still in a `restore`, as is the user-specified action, but `runBefore` and `runAfter` are no longer in a `restore`. The previous version's `restore` usage came in from #1207, so I verified that the `conn-killed` binary still produces `Right` with the new masking. Additionally, it's worth noting that `runOnException` is now implicitly in an `uninterruptibleMask` (via `unliftio`'s `withException`).

The PR might be easiest to review commit-by-commit, as intentionally failing tests were added prior to changes being made to the libraries.

---

- [X] Ran `stylish-haskell` on any changed files.

After submitting your PR:


persistent/Database/Persist/Sql/Run.hs
```diff
@@ -105,28 +107,63 @@ runSqlPoolWithHooks r pconn i before after onException =
...
+-- rethrown and will never trigger a retry of the transaction.
+--
+-- Considering @persistent@ abstracts over specific SQL backends, you will
+-- likely need to reach for a backend-specific exception type when defining an
+-- exception predicate.
+--
+-- @since 2.14.6.0
+runSqlPoolWithExtensibleHooksRetry
+    :: forall backend m a. (MonadUnliftIO m, BackendCompatible SqlBackend backend)
+    => (UE.SomeException -> Bool)
```
jship, line 138: It's worth pointing out that the API has been intentionally kept simple here: so long as exceptions match the predicate, the transaction will always be retried. In the case of serialization failures at repeatable-read and serializable isolation specifically, the recommendation from the PostgreSQL docs and elsewhere is to retry these failing transactions unconditionally.

That guidance is what drove this simpler API, but users may have [other failures](https://www.postgresql.org/docs/current/mvcc-serialization-failure-handling.html) (e.g. uniqueness violations) they would like to only retry up to a fixed number of times or retry based on some more sophisticated policy. The simple exception predicate approach in this PR would not work for those more complicated cases. Considering the more complicated cases are rare, it seemed prudent to keep the retry API as simple as possible for now. However, depending upon the exception predicate the user specifies, they may get themselves into a situation where `persistent` indefinitely retries. For these cases, they could wrap their `runSqlPoolWithExtensibleHooksRetry` call in a `timeout`.

jship:
We've had success running these changes in production for a few months now. Is there anything I can help with in regards to moving the PR along?

konrad-npx:
A potential issue with this design is that `parseCheckJsonBody` only works once. I ended up catching exceptions for my project.

PR #1497, opened 2023-05-18
Always creates `SymbolToField "id"` instance

hw202207:
## Context

- Regardless the type of primary key, there is always a `{EntityName}Id` type hence I think it's safe to always create a `SymbolToField` instance for ID.

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,5 +1,10 @@
 # Changelog for persistent
 
+## Unreleased
+
+* [1497](https://github.com/yesodweb/persistent/pull/1497)
+    * Always generates `SymbolToField "id"` instance
+
 ## 2.14.5.1
 
 * [#1496](https://github.com/yesodweb/persistent/pull/1496)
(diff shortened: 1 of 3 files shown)


persistent/ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog for persistent
 
+## Unreleased
+
+* [1497](https://github.com/yesodweb/persistent/pull/1497)
+    * Always generates `SymbolToField "id"` instance
```
parsonsmatt, line 6: This will need to be a minor version bump - may be best to roll up with #1503

hw202207: Sounds reasonable to me. Anything I need to do / pay attention to?

hw202207: I'll rebase after #1503 has been merged (adjust the changelog section)

parsonsmatt:
This builds and tests just fine locally so I am going to merge. Thank you!

PR #1499, opened 2023-05-22
Make it build with ghc-9.6

erikd:


erikd:
This could also be fixed with a package revision on Hackage.

erikd:
Ping?

parsonsmatt:
Ahh, looks like the CI error is more related to the newer version of `aeson` which has reformatted parse errors. A constraint in the test suite's build dependencies should get that passing.

erikd:
@parsonsmatt ?

.github/workflows/haskell.yml
```diff
@@ -45,15 +45,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        cabal: ["3.6"]
+        cabal: ["3.10"]
         ghc:
-          - "8.4.4"
```
parsonsmatt, line 50: `persistent` still supports 8.4 per the cabal file's `base >= 4.11.1.0` bound. I'd prefer to retain support here if we can manage it.

persistent/persistent.cabal
```diff
@@ -17,7 +17,7 @@ extra-source-files: ChangeLog.md README.md
 library
     build-depends:   
         base                     >= 4.11.1.0     && < 5
-      , aeson                    >= 1.0 && < 2.2
+      , aeson                    == 2.1.*
```
parsonsmatt, line 20: Why is this bound so much tighter now?

erikd: Was trying to avoid the issue with Aeson error message changes causing tests to fail.

erikd: And of course it failed.

erikd: The problem is that `persistent` depends on `shakespeare` which has a base dependency of `(>=4.12 && <5)` (ie only build with `ghc-8.6` and later). 

I do not see an easy solution to this other than dropping `ghc-8.4` (the version in that series was release on 14th October 2018).

parsonsmatt: Ahhh, I see, so the solver needs to pick a newer version of `shakespeare`  to satisfy the `aeson > 2`, and *that* `shakespeare` also bumps the base bound. So `shakespeare` is blocking any package from compiling with GHC 8.4 that also supports `aeson >= 2`. That's annoying!

Testing locally, `shakespeare` HEAD builds fine with GHC 8.4. So it's not even necessary.

erikd: So what do you propose is the solution required to get this PR merged?

parsonsmatt: I'm OK dropping GHC 8.4 support; I'll make a follow-on issue to either restore it or finish the job (and delete relevant CPP bits for compatibility)

erikd:
Last commit can be dropped when the new `shakespeare` has been released.

andreasabel:
NB: The latest released version of `persistent` builds with GHC after a TH bound relaxation: https://hackage.haskell.org/package/persistent-2.14.5.1/revisions/

PR #1502, opened 2023-06-20
Add attoparsec-aeson for aeson-2.2

9999years:


persistent/persistent.cabal
```diff
@@ -15,10 +15,11 @@ bug-reports:     https://github.com/yesodweb/persistent/issues
 extra-source-files: ChangeLog.md README.md
 
 library
-    build-depends:   
+    build-depends:
         base                     >= 4.11.1.0     && < 5
-      , aeson                    >= 1.0 && < 2.2
+      , aeson                    >= 1.0 && < 2.3
       , attoparsec
+      , attoparsec-aeson         >= 2.2 && <3
```
parsonsmatt, line 22: hmm, this is unfortunate

`attoparsec-aeson` itself has a dependency on `aeson >= 2.2.0.0 && < 2.3`, so this effectively means that `persistent` now requires `aeson >= 2.2`.

`attoparsec-aeson` should *probably* have another version available, perhaps `2.1.0.0`, which has a `aeson < 2.1` bound, which doesn't define any modules. Hmm.

parsonsmatt: I don't think this is necessary, actually, since we don't depend on `Data.Aeson.Parser` ?

parsonsmatt: Made an issue anyway, since this is a problem for providing support for multiple versions: https://github.com/haskell/aeson/issues/1048

Issue #1297, opened 2021-07-14
Support `-- ^` comment style


parsonsmatt:
Currently, you must use the `-- |` comment style for documenting fields and entities. This is not idiomatic Haskell, and folks often want to use `-- ^` for documenting fields. 

Related to #1296 

Right now, you need to document like this:

```
-- | A table
SomeTable
    -- | Some Field
    someField Text
```

Ideally, we would be able to document like this:

```
-- | A table
SomeTable
    someField Text 
    -- ^ Some Field
```

Supporting same-line comments seems useful too.

```
SomeTable
    someField Text -- ^ This is a doc comment.
```

It seems less obvious that documenting *entities* like this makes sense. I've never see this on a Haskell datatype and I'm not sure it's even valid syntax.

```
SomeTable
    someField Text -- ^ Doc Comment
-- ^ I am a doc comment for SomeTable?? Weird.
```

Issue #1296, opened 2021-07-14
Support single pipe comment style


parsonsmatt:
In `persistent`'s QuasiQuoter, you can add documentation using a `-- |` style doc-comment. Like in `purescript`, the `|` must be present on every comment.

```
-- | This is a table.
-- | This is another line of the doc comment.
SomeTable
    -- | This is a field.
    -- | This is another line of the field comment.
    someField Text
```

However, this requirement is odd to Haskellers, so we should support having only a single pipe.

```
-- | This is a table.
-- This is another line of the doc comment.
SomeTable
    -- | This is a field.
    -- This is another line of the field comment
    someField Text
```

This issue is done when the two code blocks create equal commentary on the produced `EntityDef`s.

PR #1503, opened 2023-06-27
Create Haddocks from entity documentation comments

### LINKED ISSUE #1462: Doc comments should be included in Haddock, optionally
Recent GHC allow you to put documentation comments in Haddock, using `putDoc` and `getDoc`.

We should support this, optionally - I suspect that these operations are actually somewhat slow, so I want to hide it behind a flag.

blujupiter32:
This PR adds a flag to `MkPersistSettings` which enables Haddock generation from entity documentation comments. The default is `False` to avoid any unexpected performance impacts from upgrading `persistent`.

Fixes #1462.

---

  - Changes to the documentation assume that this feature will be made available in `persistent` version `2.14.6.0`.
  - `stylish-haskell` fails on the `#if`-`#else`-`#endif` construct in the import list in `TH.hs` as all three preprocessor directives are ignored, resulting in two identifier lists for `Language.Haskell.TH.Lib`.

After submitting your PR:

### DIFF
--- a/persistent/persistent.cabal
+++ b/persistent/persistent.cabal
@@ -145,6 +145,7 @@ test-suite test
       , silently
       , template-haskell         >= 2.4
       , text
+      , th-lift
       , th-lift-instances
       , time
       , transformers
@@ -172,6 +173,7 @@ test-suite test
         Database.Persist.TH.CompositeKeyStyleSpec
         Database.Persist.TH.DiscoverEntitiesSpec
         Database.Persist.TH.EmbedSpec
+        Database.Persist.TH.EntityHaddockSpec
         Database.Persist.TH.ForeignRefSpec
         Database.Persist.TH.ImplicitIdColSpec
         Database.Persist.TH.JsonEncodingSpec
--- a/persistent/persistent.cabal
+++ b/persistent/persistent.cabal
@@ -145,7 +145,6 @@ test-suite test
       , silently
       , template-haskell         >= 2.4
       , text
-      , th-lift
       , th-lift-instances
       , time
       , transformers
--- a/persistent/persistent.cabal
+++ b/persistent/persistent.cabal
@@ -1,5 +1,5 @@
 name:            persistent
-version:         2.14.5.0
+version:         2.14.6.0
 license:         MIT
 license-file:    LICENSE
 author:          Michael Snoyman <michael@snoyman.com>
--- a/persistent/test/Database/Persist/TH/EntityHaddockSpec.hs
+++ b/persistent/test/Database/Persist/TH/EntityHaddockSpec.hs
@@ -8,7 +8,7 @@ import TemplateTestImports
 #if MIN_VERSION_template_haskell(2,18,0)
 import Database.Persist.TH.CommentSpec (CommentModel (..))
 import Language.Haskell.TH (DocLoc (DeclDoc), getDoc)
-import Language.Haskell.TH.Lift (lift)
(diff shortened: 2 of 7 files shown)


parsonsmatt:
This is fantastic, thanks!

Can you clarify the limitatiion around documenting fields?

Can you post the problematic test? That smells like a GHC bug to me, and might be worth reporting upstream.

For a test, I'd be happy to see a rendered Haddock that contains the docs - a screenshout would be fine.

blujupiter32:
On closer inspection, the limitation only applies to `withDecDoc`, so I've added Haddock support for fields using `putDoc`. As for the test, I've added a spec module in https://github.com/blujupiter32/persistent/pull/5 which fails with the error. Building the Haddock documentation for `persistent` from this branch should show an additional `Test.HaddockExample` module with entries generated from the doc comments.

persistent/test/Database/Persist/TH/EntityHaddockSpec.hs
```diff
@@ -0,0 +1,36 @@
...
+{-# LANGUAGE TemplateHaskell #-}
+
+module Database.Persist.TH.EntityHaddockSpec (spec) where
+
+import TemplateTestImports
+
+#if MIN_VERSION_template_haskell(2,18,0)
+import Database.Persist.TH.CommentSpec (CommentModel (..))
+import Language.Haskell.TH (DocLoc (DeclDoc), getDoc)
+import Language.Haskell.TH.Lift (lift)
```
parsonsmatt, line 11:
```suggestion
import Language.Haskell.TH.Syntax (lift)
```

Should work to allow removing the `th-lift` dependency

parsonsmatt (changes requested):

persistent/persistent.cabal
```diff
@@ -145,6 +145,7 @@ test-suite test
       , silently
       , template-haskell         >= 2.4
       , text
+      , th-lift
```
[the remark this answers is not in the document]
blujupiter32: Fixed in 3fbb38362faba6465d4894bc64ae0ec907fc1ee8.

persistent/test/Database/Persist/TH/EntityHaddockSpec.hs
blujupiter32: Fixed in 3fbb38362faba6465d4894bc64ae0ec907fc1ee8.

blujupiter32:
Regarding version bumping, would this change require a minor version bump (to `2.15.0.0`) as `mpsEntityHaddocks` has been added to the public API?

Also, would it be worth addressing #1296 and #1297 (and merging #1505) before releasing this change so that documentation comments are syntactically equivalent to the format supported by Haddock? Alternatively, these issues could be fixed and released in subsequent patch versions if frequent releases aren't a problem.

parsonsmatt:
This would be a minor version bump, which put be `2.14.6.0`, since PVP has two major versions.

I think I'd be fine releasing this separately from changing the way the comment syntax works.

blujupiter32:
e2420bc78130dc6614b517860fe0a09abb6f6ecb conflicts with 0bc89423a6c7452a6345a9a1e33e63c92d5d9d19, so would you prefer rebasing onto `master` or merging `master` into the feature branch? PRs seem to be squash-merged so I don't think it would make much difference either way.

parsonsmatt:
You can do either; I generally find that merges make viewing diffs slightly easier on GitHub since I can see "changes since last review," but you're right that it'll get squash/merged into `master` anyway.

PR #1508, opened 2023-07-11
Store object IDs as blobs in SQL

tfausak:
I'm mostly opening this PR for discussion. 

The `persistent-mongoDB` package provides a `PersistFieldSql` instance for the `ObjectId` type. Unfortunately the implementation just uses `SqlOther` with a message to the developer. Usually this doesn't matter because MongoDB doesn't use `PersistFieldSql`. The instance could have anything in it and everything would still work.

However, a proper instance for the `ObjectId` type is useful if you want to store object IDs in a SQL database. In that case, storing them as a `BLOB` makes sense. In fact that's what the `PersistField` instance already does. 

Does this change make sense? Am I missing something else about these instances?

### DIFF
--- a/persistent-mongoDB/Database/Persist/MongoDB.hs
+++ b/persistent-mongoDB/Database/Persist/MongoDB.hs
@@ -238,10 +238,10 @@ instance PersistField DB.ObjectId where
     fromPersistValue _ = Left $ T.pack "expected PersistObjectId"
 
 instance Sql.PersistFieldSql DB.ObjectId where
-    sqlType _ = Sql.SqlOther "doesn't make much sense for MongoDB"
+    sqlType = const Sql.SqlBlob
 
 instance Sql.PersistFieldSql (BackendKey DB.MongoContext) where
-    sqlType _ = Sql.SqlOther "doesn't make much sense for MongoDB"
+    sqlType = const Sql.SqlBlob
 
 
 withConnection :: (Trans.MonadIO m)


tfausak:
Ah well, there is one problem with this 😄 

https://github.com/yesodweb/persistent/blob/0bc89423a6c7452a6345a9a1e33e63c92d5d9d19/persistent-postgresql/Database/Persist/Postgresql/Internal.hs#L61C1-L62

I think that instance could change to encode `PersistObjectId` the same as `PersistByteString`. But I'm not sure if I should keep pulling on that thread or not.

parsonsmatt:
Yeah, I think I'd rather a) support it in a sensible manner, or b) put a `TypeError` instance on it, or c) delete the instance entirely and let folks do orphans if they want.

I'm not familiar enough with what an `ObjectId` really is to make an informed decision here. There's some value in having, say, `Key Foo` as a column type automatically being inferred as a foreign key, which does provide a bit of type safety - if you have a `Key Foo` that's sourced from a table with a foreign key constraint, then you have some knowledge that the record does exist in that database (or did, when the record was retrieved).

tfausak:
This page describes object IDs from MongoDB's perspective: https://www.mongodb.com/docs/manual/reference/method/ObjectId/

Object IDs are a special type in the BSON spec: https://bsonspec.org/spec.html

As far as Haskell is concerned, an `ObjectId` is just a `Word32` and a `Word64`: https://hackage.haskell.org/package/bson-0.4.0.1/docs/Data-Bson.html#t:ObjectId

Personally, I think there are a few SQL representations for object IDs that would work alright:

- `SqlBlob`: Essentially the existing one that uses a byte string. 
- `SqlString`: Serialize it as a hexadecimal string. 
- `SqlNumeric`: Somehow putting it into a 96-bit integer, I think? I'm not too familiar with `SqlNumeric`.

tfausak:
Realistically I don't think I have the time to make this change.

Issue #1458, opened 2022-12-30
Cannot turn off PathPiece and ToHttpApiData newtype instances for keys


lf-:
Hi!

I'm writing a not-a-webapp and I think it's somewhat odd that Persistent will generate `deriving newtype instance FromHttpApiData` and likewise for `PathPiece` for all keys, even though this should not be necessary for non-web users; furthermore there is a mechanism in MkPersistSettings to set `mpsDeriveInstances` to derive more instances, which could be used to achieve this for web users that want it.

I think that the default `mkPersistSettings` should probably be changed to include these extra instances, such that it can be overridden if they are not desirable. This would unfortunately be a breaking change for those overriding `mpsDeriveInstances` while discarding the default value (which was previously empty-list).

https://github.com/yesodweb/persistent/blob/7222127f812edd51c0de7f5dbc0121c80aacd444/persistent/Database/Persist/TH.hs#L1748-L1750

Issue #649, opened 2017-01-11
Automatic (TH) PathPiece generation is poorly documented, and confusing


tebello-thejane:
The documentation for `mkPersist` says 
> Create data types and appropriate `PersistEntity` instances for the given `EntityDefs`. Works well with the persist quasi-quoter.

I'm struggling to understand why it also creates `PathPiece` instances. In particular, I have hit a confusing issue where, with more than two tables being defined, using an enum type (whose instances where generated via `derivePersistField`) in the one table causes the compiler to complain that the data type does not have the appropriate `path-pieces` instances, but enums in another table do not cause the same error.

The source code for the module reads
>  -- | declare the key type and associated instances
>  -- **@'PathPiece'@, @'ToHttpApiData'@ and @'FromHttpApiData'@ instances are only generated for a Key with one field**
> `mkKeyTypeDec :: MkPersistSettings →  EntityDef →  Q (Dec, [Dec])`

What does this even mean?? Note that my project does not use Yesod (yet), so I'm finding it difficult to justify creating `PathPiece` and `*HttpApiData` instances just to satisfy the compiler when there aren't any actual paths in the code base, or any other hint of HTTP.

On a possibly-related note, declaring `DeriveGeneric` and `DeriveAnyClass` as default extensions in my project causes `mkPersist` to blow up and start complaining about missing `Generic` instances. That is, I guess this might be a weird extensions issue, but right now my guesses are worthless.

If necessary, I can try and create a minimal example producing the issue. But I'm hoping this rings a bell and has been encountered by someone else before.

tebello-thejane:
Here is a minimal example.

In a separate module, we have:
```
import Database.Persist.TH

data Bug = Ant | Bee | Roach deriving (Show, Read, Eq, Ord)
derivePersistField "Bug"        
```

Then our main program has:
```
share [mkPersist sqlSettings] [persistLowerCase|
Foo                                                                          
    bar Bug
    Primary idea
 |]    
```

The compile will complain with
```
Main.hs:10:1: error:
    • No instance for (path-pieces-0.2.1:Web.PathPieces.PathPiece Bug)
        arising from the 'deriving' clause of a data type declaration
      Possible fix:
        use a standalone 'deriving instance' declaration,
          so you can specify the instance context yourself
    • When deriving the instance for (path-pieces-0.2.1:Web.PathPieces.PathPiece
                                        (Key Foo))

Main.hs:10:1: error:
[...]

PR #1509, opened 2023-07-16
Optionally derive `PathMultiPiece` for composite keys

blujupiter32:
This PR enables `PathMultiPiece` instance derivation for entities with composite keys using `mpsDeriveInstances`.

Unlike `PathPiece`, an instance is **not** derived automatically. This should avoid the coupling concerns expressed in #649 and #1458. A similar technique could be used for `PathPiece` etc. in a future (breaking) change to address these issues.

---

  - Changes to the documentation assume that this feature will be made available in `persistent` version `2.14.6.0`.
  - `stylish-haskell` suggests fixes to `Database.Persist.Class.PersistEntity` which are unrelated to this PR: blujupiter32#10

After submitting your PR:


parsonsmatt (changes requested):
I don't think I'm in favor of this change. There's a bit too much special-case logic here.

Currently, there are two ways of deriving instances for keys: the `mps` setting, and manually.

```haskell
    Blah
      name String
  |]

instance PathMultiPiece (Key Blah) where ...
```

I'm OK with introducing features to make manual instances easier - ie exporting the function defaults that are based on `PersistEntity a => ...`, and a `newtype` for `DerivingVia`.

persistent/Database/Persist/Quasi.hs
```diff
@@ -171,6 +171,26 @@ As a key with multiple fields, it is also a "composite key."
...
+This enables composite keys to be used in Yesod routes, for example.
+
+@
+Email
+    firstPart   Text
+    secondPart  Text
+
+    Primary firstPart secondPart
+
+    deriving PathMultiPiece
```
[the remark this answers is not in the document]
blujupiter32: This usage, and the Haddock change documenting it here, have been removed in 950098cf5a4369b37c0ce84de19538bfb4c5f8dc.

persistent/Database/Persist/Class/PersistEntity.hs
```diff
@@ -180,6 +182,27 @@ class ( PersistField (Key record), ToJSON (Key record), FromJSON (Key record)
...
+        Right key <- keyFromValues <$> mapM fromPathPiece pieces
+        pure key
+    -- | 'toPathMultiPiece' wrapper. The default implementation uses
+    -- 'keyToValues' and 'toPathPiece', which are provided by the context.
+    keyToPieces :: Key record -> [Text]
+    keyToPieces = map toPathPiece . keyToValues
+
+instance PersistPathMultiPiece record => PathMultiPiece (Key record) where
+    fromPathMultiPiece = keyFromPieces
+    toPathMultiPiece = keyToPieces
```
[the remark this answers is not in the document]
blujupiter32: I've replaced `PersistPathMultiPiece` with the more idiomatic approach in 950098cf5a4369b37c0ce84de19538bfb4c5f8dc.

persistent/Database/Persist/TH.hs
```diff
@@ -1780,6 +1776,18 @@ mkKeyTypeDec mps entDef = do
...
 
+    -- Always use StockStrategy for Show/Read.
+    -- This means e.g. (FooKey 1) shows as ("FooKey 1"), rather than just "1".
+    -- This is much better for debugging/logging purposes:
+    -- cf. https://github.com/yesodweb/persistent/issues/1104
+    decideStrategy :: Name -> DerivStrategy
+    decideStrategy typeclass
+        | typeclass `elem` [''Show, ''Read] = StockStrategy
+        | typeclass == ''PathMultiPiece =
+            ViaStrategy $ ConT ''ViaPersistEntity `AppT` recordType
```
parsonsmatt, line 1790: Apologies, I should have been more clear - I'm not going to accept a PR that has a special-case for how an instance is derived. 

```suggestion
```

parsonsmatt (changes requested):

persistent/Database/Persist/TH.hs
blujupiter32: OK, so regarding your first comment,

> I'm OK with introducing features to make manual instances easier - ie exporting the function defaults that are based on `PersistEntity a => ...`, and a `newtype` for `DerivingVia`.

should I revert my changes to `Database.Persist.TH` and remove `CompositeKeyPathMultiPieceSpec`? That would leave `ViaPersistEntity` and its `PathMultiPiece` instance as the additions from this PR. If those don't seem useful on their own, I'm fine with closing the PR.

I'll keep the rest on my fork as automatic `PathMultiPiece` derivation is helpful for my specific use case, although I don't think it would be used often if it was added here (no issues have been raised to request it).

parsonsmatt: Yes, that works.

I need to document more about how to hook into the derivation stuff without modifying this function here.

The way that I would approach this is with an additional function that operates on the `[UnboundEntityDef]` from a thing. So, you have,

```haskell
$(share [mkPersist sqlSettings, derivePathMultiPiece] ...)
```

Where `derivePathMultiPiece :: [UnboundEntityDef] -> Q [Dec]`, and you can do something like:

```haskell
derivePathMultiPiece = 
  concat <$> traverse derivePathMultiPieceFor
  where 
    derivePathMultiPieceFor unboundEntityDef = 
      let keyTypeName = ...
      [d| deriving via (ViaPersistentEntity $(keyTypeName)) instance PathMultiPiece $(keyTypeName) |]
```

This gives you the same automatic instance generation, without modifying the upstream code.

parsonsmatt (approved):
Thanks! This is great 🎉

PR #1518, opened 2023-09-20
postgres(normalize types to eliminate noop migrations)

### LINKED ISSUE #1517: Postgres: BIGINT and INT8 are an alias and should not require migration.
I'm using Persistent on a project where the postgres schema is already defined by an outside source. This source uses `BIGINT` for a lot of columns. I'm using persistent migrations to check if my persistent model matches that database schema. After all, if the migration is empty then everything should be in order.

However, I'm hitting a lot of false positives because the migration contains a lot of such statements:

```
ALTER TABLE "table" ALTER COLUMN "column" TYPE INT8;
```

Judging by the following:

```
SELECT * FROM pg_type WHERE typname IN ('int8', 'bigint');

|typname|
|-------|
|int8   |
```

Persistent should recognize that there is nothing to migrate.

HugoPeters1024:
Fixes #1517. I've kept the list non exhaustive because I simply was not sure how to handle every case. Notably when the alias did not have a placeholder variable but the base name did (e.g. timestamptz). I also looked around in the `Postgresql.Simple` modules to see if something like this already existed but it did not look like it.

At the risk of sounding selfish, this fixes my immediate problem does not make the code worse.

<hr>

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -1103,7 +1103,15 @@ getColumn _ _ columnName _ =
 -- | Intelligent comparison of SQL types, to account for SqlInt32 vs SqlOther integer
 sqlTypeEq :: SqlType -> SqlType -> Bool
 sqlTypeEq x y =
-    T.toCaseFold (showSqlType x) == T.toCaseFold (showSqlType y)
+    let
+        -- Non exhaustive helper to map postgres aliases to the same name. Based on
+        -- https://www.postgresql.org/docs/9.5/datatype.html.
+        -- This prevents needless `ALTER TYPE`s when the type is the same.
+        normalize "int8"    = "bigint"
+        normalize "serial8" = "bigserial"
+        normalize v         = v
+    in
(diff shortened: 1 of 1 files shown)


parsonsmatt:
This looks good to me. If you want to add a test and finish the PR checklist I'd be happy to do a patch release.

persistent/ChangeLog.md
```diff
@@ -1,4 +1,9 @@
-# Changelog for persistent
+# Changelog for persistentChan
+
+## 2.14.6.1 (unreleased)
+
+* [#1518](https://github.com/yesodweb/persistent/pull/1518)
+    * Normalize postgres type aliases to prevent noop migrations
```
parsonsmatt, line 6: The change is actually in `persistent-postgresql`, not `persistent`, so you'll want to make the version bump and changelog modification there 😄

parsonsmatt (changes requested):

parsonsmatt:
Released as `persistent-postgresql-2.13.6.1`

persistent/ChangeLog.md
HugoPeters1024: thanks for fixing it :)

PR #1519, opened 2023-09-30
Add keyAndEntityFieldsDatabase function to return key with migration only fields

josephsumabat:
This PR adds a version of the `keyAndEntityFields` function which includes `MigrationOnly` and `SafeToRemove` fields. We would like to make use of this in `persistent-documentation` in order to generate documentation for fields marked by those modifiers.

After submitting your PR:


parsonsmatt (commented):
Thanks, looks great! Do you mind adding a Changelog entry? I haven't released 2.14.6.0 yet so we can fold this in there.

PR #1526, opened 2023-11-28
persistent-mysql: treat tinyint as SqlOther "tinyint", not SqlBool

fumieval:
I noticed that our application reports schema mismatch after upgrading to MySQL 8.0, even if I change the column type to `SqlOther "tinyint"` (removing the width). I realised that the recent versions of persistent-mysql parses tinyint as SqlBool, but it doesn't work in our usecase, where some tinyint columns represent non-Boolean values.

After this change, I confirmed that our application works both on MySQL 5.7 and MySQL 8.0, but I'm not quite confident that this change is in the right direction. What are your thoughts?

----

After submitting your PR:

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -827,14 +827,16 @@ data ColumnInfo = ColumnInfo
 -- @INFORMATION_SCHEMA@ tables.
 parseColumnType :: Text -> ColumnInfo -> ExceptT String IO (SqlType, Maybe Integer)
 -- Ints
--- The display width is deprecated and being removed in MySQL 8.X.  To be
--- consistent with earlier versions, which do report it, accept either
+-- The display width is deprecated and being removed in MySQL 8.X
+-- with [an exception of tinyint(1) which is used for boolean values](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-19.html#mysqld-8-0-19-deprecation-removal).
+-- To be consistent with earlier versions, which do report it, accept either
 -- the bare type in `ciColumnType ci`, or the type adorned with the expected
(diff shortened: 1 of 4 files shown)


paul-rouse:
This looks reasonable to me.  It relies on the exception documented in the [MySQL 8.0.19 release notes](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-19.html#mysqld-8-0-19-deprecation-removal) concerning the deprecation of the width specification for integer data types - it does appear that there is an intention to retain `tinyint(1)` as a special case.

There are a couple of things you could do to tidy up the code:
  * the comment on `parseColumnType` needs to be updated to refer to this exception to the general deprecation of integer width specification, and the comment in `showSqlType` specifically about the `Bool` case can be removed.
  * I can't see any reason for the second condition (the one with `stripPrefix`), since removing the `tinyint` alternative from the condition will make anything other than `tinyint(1)` fall through to the catch-all case of `parseColumnType` at the bottom.  Doesn't that do what you want?

paul-rouse:
@parsonsmatt can you see any issues with this?  Just to make sure we don't break anything!

parsonsmatt:
Code looks fine to me. I'm curious if you could write a test to demonstrate the behavior you're looking for?

fumieval:
I added a unit test that tests this behaviour. Would that suffice?

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -74,6 +74,7 @@ import qualified Data.Text.IO as T
 import GHC.Stack
 import System.Environment (getEnvironment)
 
+import Database.Persist.MySQL.Internal
```
parsonsmatt, line 77: I think this file wasn't checked in

parsonsmatt (approved):
Fantastic, thanks! I'll go ahead and release this.

PR #1530, opened 2024-04-29
Document the `!no-migrate` Entity Attribute

prikhi:
Add docs to the quasiquoter module about how the `!no-migrate` entity attribute will prevent any migrations from being generated for an entity. This is currently only "documented" in the source code:

https://github.com/yesodweb/persistent/blob/97013604d9ca9bf39b03a36005287027b621bfff/persistent/Database/Persist/TH.hs#L2769

This has existed for a while, does it need a `@since` declaration or CHANGELOG entry in here?

### DIFF
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -352,6 +352,17 @@ userAttrs = do
 -- [["sad"],["sogood"]]
 @
 
+== @!no-migrate@
+
+To prevent @migrateModels@ from generating _any_ migrations for an entity, add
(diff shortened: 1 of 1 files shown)


persistent/Database/Persist/Quasi.hs
```diff
@@ -352,6 +352,17 @@ userAttrs = do
 -- [["sad"],["sogood"]]
 @
 
+== @!no-migrate@
```
prikhi, line 355: I picked this spot since it's right after the attribute docs. I could make it go one level deeper so it's nested within them :shrug:

PR #1537, opened 2024-05-30
Elaborate on 'upsertManyWhere'

lfborjas:
Expand the documentation of the Postgres version of `upsertManyWhere`: it wasn't clear how the second argument behaves and at least two coders interpreted it backwards introducing a subtle production bug; hopefully the examples now make it clear!

I don't have a dev env setup for this so I didn't run `stylish-haskell`, nor do I know if a documentation update warrants a version bump — happy to comply if necessary though!

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -1714,7 +1714,7 @@ upsertWhere record updates filts =
 
 -- | Postgres specific 'upsertManyWhere'.
 --
--- The first argument determines a list of new records to insert.
+-- The first argument is a list of new records to insert.
 --
 -- If there's unique key collisions for some or all of the proposed insertions,
 -- you can use the second argument to specify which fields, under which
@@ -1744,14 +1744,22 @@ upsertWhere record updates filts =
 --
 -- @
 --   upsertManyWhere
---     [record] -- new records to insert if there's no conflicts
---     [ copyField recordField1 -- for each conflicting existing row, replace the value of recordField1 with the one present in the conflicting new record
---     , copyUnlessEq recordField2 -- only replace the existing value if it's different from the one present in the conflicting new record
---     , copyUnlessNull recordField3 -- only replace the existing value if the new value is non-NULL (i.e. don't replace existing values with NULLs.)
+--     [record]                         -- (1) 
+--     [ copyField recordField1         -- (2) 
+--     , copyUnlessEq recordField2      -- (3) 
+--     , copyUnlessNull recordField3    -- (4) 
 --     ]
---     [recordField4 =. arbitraryValue] -- update recordField4 with an arbitrary new value
---     [recordField4 !=. anotherValue] -- only apply the above updates for conflicting rows that meet this condition
+--     [recordField4 =. arbitraryValue] -- (5) 
+--     [recordField4 !=. anotherValue]  -- (6) 
 -- @
+-- 
+-- 1. new records to insert if there's no conflicts
+-- 2. for each conflicting existing row, replace the value of recordField1 with the one present in the conflicting new record
+-- 3. only replace the existing value if it's different from the one present in the conflicting new record
+-- 4. only replace the existing value if the new value is non-NULL (i.e. don't replace existing values with NULLs.)
+-- 
+-- 5. update recordField4 with an arbitrary new value
+-- 6. only apply the above updates for conflicting rows that meet this condition
 --
 -- Called thusly, this method will insert a new record (if none exists) OR it
 -- will copy three fields (@recordField1@, @recordField2@, @recordField3@) from
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -1712,16 +1712,70 @@ upsertWhere
 upsertWhere record updates filts =
   upsertManyWhere [record] [] updates filts
 
--- | Postgres specific 'upsertManyWhere'. This method does the following:
--- It will insert a record if no matching unique key exists.
--- If a unique key exists, it will update the relevant field with a user-supplied value, however,
--- it will only do this update on a user-supplied condition.
--- For example, here's how this method could be called like such:
+-- | Postgres specific 'upsertManyWhere'.
 --
--- upsertManyWhere [record] [recordField =. newValue] [recordField !=. newValue]
+-- The first argument determines a list of new records to insert.
 --
--- Called thusly, this method will insert a new record (if none exists) OR update a recordField with a new value
--- assuming the condition in the last block is met.
+-- If there's unique key collisions for some or all of the proposed insertions,
+-- you can use the second argument to specify which fields, under which
+-- conditions (using 'HandleUpdateCollision' strategies,) will be copied from
+-- each proposed, conflicting new record onto its corresponding row already present
+-- in the database. Helpers such as 'copyField', 'copyUnlessEq',
+-- 'copyUnlessNull' and 'copyUnlessEmpty' are exported from this module to help
+-- with this, as well as the constructors for 'HandleUpdateCollision'.
+-- For example, you may have a patch of updates to apply, some of which are NULL
+-- to represent "no change"; you'd use 'copyUnlessNull' for each applicable
+-- field in this case so any new values in the patch which are NULL don't
+-- accidentally overwrite existing values in the database which are non-NULL.
+--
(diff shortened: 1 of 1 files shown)


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1712,16 +1712,70 @@ upsertWhere
...
+-- @
+--   upsertManyWhere
+--     [record] -- new records to insert if there's no conflicts
+--     [ copyField recordField1 -- for each conflicting existing row, replace the value of recordField1 with the one present in the conflicting new record
+--     , copyUnlessEq recordField2 -- only replace the existing value if it's different from the one present in the conflicting new record
+--     , copyUnlessNull recordField3 -- only replace the existing value if the new value is non-NULL (i.e. don't replace existing values with NULLs.)
+--     ]
+--     [recordField4 =. arbitraryValue] -- update recordField4 with an arbitrary new value
+--     [recordField4 !=. anotherValue] -- only apply the above updates for conflicting rows that meet this condition
+-- @
```
parsonsmatt, line 1754: This looks a little noisy - can we try and get the line length below 80 columns? Putting the comments directly above the item in question

```suggestion
-- @
--   upsertManyWhere
--     [record]                         -- (1) 
--     [ copyField recordField1         -- (2) 
--     , copyUnlessEq recordField2      -- (3) 
--     , copyUnlessNull recordField3    -- (4) 
--     ]
--     [recordField4 =. arbitraryValue] -- (5) 
--     [recordField4 !=. anotherValue]  -- (6) 
-- @
-- 
-- 1. new records to insert if there's no conflicts
-- 2. for each conflicting existing row, replace the value of recordField1 with the one present in the conflicting new record
-- 3. only replace the existing value if it's different from the one present in the conflicting new record
-- 4. only replace the existing value if the new value is non-NULL (i.e. don't replace existing values with NULLs.)
-- 
-- 5. update recordField4 with an arbitrary new value
-- 6. only apply the above updates for conflicting rows that meet this condition
```

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1712,16 +1712,70 @@ upsertWhere
...
 
--- | Postgres specific 'upsertManyWhere'. This method does the following:
--- It will insert a record if no matching unique key exists.
--- If a unique key exists, it will update the relevant field with a user-supplied value, however,
--- it will only do this update on a user-supplied condition.
--- For example, here's how this method could be called like such:
+-- | Postgres specific 'upsertManyWhere'.
 --
--- upsertManyWhere [record] [recordField =. newValue] [recordField !=. newValue]
+-- The first argument determines a list of new records to insert.
```
parsonsmatt, line 1717: `determines` feels off here, I'd say `is` instead

```suggestion
-- The first argument is a list of new records to insert.
```

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1712,16 +1712,70 @@ upsertWhere
...
+-- you can use the second argument to specify which fields, under which
+-- conditions (using 'HandleUpdateCollision' strategies,) will be copied from
+-- each proposed, conflicting new record onto its corresponding row already present
+-- in the database. Helpers such as 'copyField', 'copyUnlessEq',
+-- 'copyUnlessNull' and 'copyUnlessEmpty' are exported from this module to help
+-- with this, as well as the constructors for 'HandleUpdateCollision'.
+-- For example, you may have a patch of updates to apply, some of which are NULL
+-- to represent "no change"; you'd use 'copyUnlessNull' for each applicable
+-- field in this case so any new values in the patch which are NULL don't
+-- accidentally overwrite existing values in the database which are non-NULL.
```
parsonsmatt, line 1729: This wording is a little confusing to me (but, hey, the function itself is a bit confusing!) I think if I were to write it, I'd say something like:

> The second list describes what happens if one of the records to be inserted conflicts with a unique key in the database. The 'HandleUpdateCollision' type has more documentation on this. Briefly, we can:
> 
> 1. `copyField` will copy the value from the record we tried to insert into the record in the database table.
> 2. `copyUnlessEq` (describe this)
> 3. `copyUnlessNull` (describe this)
> 4. `etc`

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1712,16 +1712,70 @@ upsertWhere
...
+--
+-- Further updates to the matching records already in the database can be
+-- specified in the third argument, these are arbitrary updates independent of
+-- the new records proposed for insertion. For example, you can use this to
+-- update a timestamp or a counter for all conflicting rows.
```
parsonsmatt, line 1734: I might start with "The third argument is a list containing updates..." or similar, so the reader knows that we're now talking about the third argument. I'd also front-load that the update *only* happens if a conflict occurred.

Possibly,

> The third argument specifies arbitrary `Update`s you can perform on conflicting rows.

With some more elaboration

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -1712,16 +1712,70 @@ upsertWhere
...
+--
+-- You can use the fourth argument to provide arbitrary conditions to constrain
+-- the above updates. This way, you can choose to only alter a subset of the
+-- conflicting existing rows while leaving the others alone. For example, you
+-- may want to copy fields from new records and update a timestamp only when the
+-- corresponding existing row hasn't been soft-deleted.
```
parsonsmatt, line 1740: Fantastic example!

persistent-postgresql/Database/Persist/Postgresql.hs
parsonsmatt: It may also be helpful to refer to the example you gave above, where you don't want to hit something with a soft-delete.

Possibly,

```haskell
  Product
    name         Text
    stock        Int Maybe
    price        Int Maybe
    updateCount  Int
    deletedAt    UTCTime Maybe

    UniqueProduct name
|]

do
  products <- loadCSV
  upsertManyWhere 
    products
    [ copyUnlessNull ProductStock
    , copyField ProductPrice
    ]
    [ updatedCount +=. 1 ]
    [ ProductDeletedAt !=. Nothing ]
```

parsonsmatt (commented):
This is a big improvement, thank you! Left a few comments for further improvement

persistent-postgresql/Database/Persist/Postgresql.hs
lfborjas: I like the more abridged approach — I'm not 100% sure but, if I follow [the docs](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT) I think the `one of the records` part might be a lil bit misleading, since there could be multiple collisions each of which will be resolved in the `ON CONFLICT` clause (in my mind `one of` means that only one collision would be resolved? Maybe I'm reading too deep into it.) 

Maybe something like `what happens for a record proposed for insertion whose unique key conflicts with an existing record in the database` (as you can see, the more precise I attempt to get, the more I veer into verbosity lol)

parsonsmatt: That's a great point! Thanks. 

It is a rather tricky thing to articulate!

> The second list describes what happens to the records which have unique key conflicts with existing rows in the database. The `HandleUpdateCollision` type documents this more thoroughly. Briefly,
>
> 1. `copyField` will update the row in the database with the value of the field of the record we attempted to insert.

is another attempt, but I'm not thrilled with it.

But let me try and break this down:

1. We have a `[record]` that we want to `upsert` into the database.
2. If the `record` does not exist, then we `insert` it.
3. If the `record` already exists by some `Unique record`, then we use `[HandleUpdateCollision record]` to specify how to handle the collision for that record.

You know, it's occuring to me that this is *much* easier to express in the singular. For example, `upsertWhere :: rec -> [HandleUpdateCollision rec] -> [Update rec] -> [Filter rec] -> m ()`. We can talk about that like:

> Attempt to insert the `record` into the database. If there is a unique key conflict, and the `[Filter rec]` are satisfied for the record we are attempting to insert, then we use the `[HandleUpdateCollision record]` and `[Update rec]` to specify how to update the record.

I think that `[Update rec]` aren't included in `upsertWhere` because it's easy enough to just edit the record itself and copy the field in.

But if `upsertWhere` is documented in an easy to explain manner, then we can make `upsertManyWhere` as:

> Like `upsertWhere`, but does a batch upsert for all of the records in the list.

PR #1540, opened 2024-07-09
Derive `{From,To}JSON` for `EntityDef`

evanrelf:
...and its dependencies.

Motivated by some orphan instances we have at `$WORK` to dump our `persistent` entities (i.e. our database schema) to JSON for manipulation elsewhere.

### DIFF
--- a/persistent/ChangeLog.md
+++ b/persistent/ChangeLog.md
@@ -1,5 +1,10 @@
 # Changelog for persistent
 
+## 2.14.7.0 (unreleased)
+
+* [#1540](https://github.com/yesodweb/persistent/pull/1540)
+    * Derive `FromJSON`, `ToJSON`, and `Generic` for `EntityDef` and its dependencies.
+
 ## 2.14.6.1
 
 * [#1528](https://github.com/yesodweb/persistent/pull/1528)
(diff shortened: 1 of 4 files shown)


evanrelf:
Not sure if you want `@since` annotations on `Generic`, `FromJSON`, and `ToJSON`, so I've omitted them for now.

evanrelf:
Not sure that upstream wants this, but I'm interested in building stuff with this at work, so maybe others would find it useful.

persistent/ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog for persistent
 
+## 2.14.7.0 (unreleased)
```
parsonsmatt, line 3:
```suggestion
## 2.14.7.0
```

I can release this asap

parsonsmatt (approved):
I'm a *little* hesitant to incorporate this, because it sort of makes the JSON representation part of the public API. 

Would you mind adding `@since` comments to the JSON instances with a comment saying that a breaking change to the JSON representation may not be considered a breaking change for PVP reasons?

evanrelf:
> I'm a _little_ hesitant to incorporate this, because it sort of makes the JSON representation part of the public API.

Yeah, I figured it wouldn't be a 100% uncontroversial change! So I don't mean to push for this strongly if you or other maintainers feel it's not the right move.

evanrelf:
We moved further away from this internally so I don't see this ever getting merged. Anyone who needs these can still write orphan instances downstream.

PR #1541, opened 2024-07-12
Use toEncoding for toPersistValueJSON

matthewbauer:
This should give us a performance benefit for free. Instead of creating a bunch of intermediate values and serializing them, we can just go straight to serialization with toEncoding. No new constraints are needed because ToJSON provides a default toEncoding.

- [ ] ~Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)~
- [ ] ~Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock~

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Class/PersistEntity.hs
+++ b/persistent/Database/Persist/Class/PersistEntity.hs
@@ -53,9 +53,10 @@ import Data.Aeson
        , (.=)
        )
 import qualified Data.Aeson.Parser as AP
-import Data.Aeson.Text (encodeToTextBuilder)
+import Data.Aeson.Encoding (encodingToLazyByteString)
 import Data.Aeson.Types (Parser, Result(Error, Success))
 import Data.Attoparsec.ByteString (parseOnly)
+import qualified Data.ByteString as B
 import Data.Functor.Identity
 import Web.PathPieces (PathMultiPiece(..), PathPiece(..))
 
@@ -71,8 +72,6 @@ import Data.Maybe (isJust)
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as TE
-import qualified Data.Text.Lazy as LT
-import qualified Data.Text.Lazy.Builder as TB
 import GHC.Generics
 import GHC.OverloadedLabels
 import GHC.TypeLits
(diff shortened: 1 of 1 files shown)


persistent/Database/Persist/Class/PersistEntity.hs
```diff
@@ -452,7 +451,7 @@ idField = "_id"
 --   toPersistValue = toPersistValueJSON
 -- @
 toPersistValueJSON :: ToJSON a => a -> PersistValue
-toPersistValueJSON = PersistText . LT.toStrict . TB.toLazyText . encodeToTextBuilder . toJSON
+toPersistValueJSON = PersistByteString . B.toStrict . encodingToLazyByteString . toEncoding
```
parsonsmatt, line 454: I'm a little concerned that changing the `PersistValue` constructor here will cause downstream issues for consumers, since the constructor tag is used to determine things like which postgres type is used.

ie anyone who is expecting for `case toPersistValueJSON x of PersistText x -> ...` is going to be broken (which would be *odd*, of course)

It seems like we can do this in a safer way by just dropping the `. toJSON` in the original?

```suggestion
toPersistValueJSON = PersistText . LT.toStrict . TB.toLazyText . encodeToTextBuilder
```

since [`encodeToTextBuilder`](https://www.stackage.org/haddock/lts-22.28/aeson-2.1.2.1/Data-Aeson-Text.html#v:encodeToTextBuilder) is `ToJSON a => a -> Builder` already.

Though the docs for that says: "Uses `toJSON`", and immediately above, we have [`encodeToLazyText`](https://www.stackage.org/haddock/lts-22.28/aeson-2.1.2.1/Data-Aeson-Text.html#v:encodeToLazyText), which changes this to:

```suggestion
toPersistValueJSON = PersistText . LT.toStrict . encodeToLazyText
```

parsonsmatt (changes requested):

persistent/Database/Persist/Class/PersistEntity.hs
matthewbauer: yeah I saw that encodeToTextBuilder uses toJSON, but didn't realize encodeToLazyText uses toEncoding

matthewbauer: Yeah PersistText seems fine. I think it's a little extra work though - it has to decode utf8 then encode it again. I think there's some fusion things GHC has that may mean it doesn't actually do this internally though.

PR #1543, opened 2024-08-05
cabal.project: Remove redundant allow-newer stanzas

erikd:
A new version of postgresql-simple has been released.

Nothing needs to be changed on Hackage as `cabal.project` does not end up on Hackage.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -12,15 +12,6 @@ constraints:
   -- https://github.com/mongodb-haskell/mongodb/pull/152
   mongoDB < 2.7.1.3 
 
--- GHC 9.4 shims for persistent
-
--- These need hackage revisions but otherwise test fine in the repo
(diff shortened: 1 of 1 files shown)


.github/workflows/haskell.yml
```diff
@@ -47,7 +47,6 @@ jobs:
       matrix:
         cabal: ["3.10"]
         ghc:
-          - "8.4.4"
```
parsonsmatt, line 50: Can we restore this? I think the shakespeare issue should be fixed by upstream hackage revision

erikd: Restored.

erikd: And all tests pass  :tada:

erikd:
Thanks @parsonsmatt !

PR #1545, opened 2024-08-29
Update mongoDB backend mongoDB driver interface

darycabrera:
The [mongoDB driver library](https://github.com/mongodb-haskell/mongodb) introduced breaking changes to the current API calls used in persistent when interfacing with a mongoDB Server version 6.0 and above. Two of those changes, namely the change in behavior of the delete and update functions, seem irrational to me, but regardless they're already officially incorporated. 

This PR makes the necessary changes to restore expected functionality whilst remaining backwards compatible with previous mongoDB versions, namely MongoDB 5.0, the [EOL for which comes up this October](https://www.mongodb.com/legal/support-policy/lifecycles). As alluded to, these changes were prompted for the mongoDB upgrade effort at work.

The mongo driver package itself needs a little more work to conform the server spec changes brought along with the Mongo Wire version 17, however, that is independent of these changes.


parsonsmatt (commented):
I think this PR needs some additional version bounds on the underlying mongo library. It seems like the interface is unchanged so it would be a patch version bump.

parsonsmatt:
Hm, incompatible with GHC 8.4.4 because of `mongoDB`'s version constraints - would require dropping support for GHC 8.4 and 8.6. That's unfortunate. GHC 8.4 is ~8 years old now, and 8.6 is 6 years old. I think I'm OK dropping those versions from our CI.

We can do a follow-up PR to tighten our own base bounds and see what things we can adopt with 8.8 as our oldest compiler target.

parsonsmatt:
OK, changes look good to me! Please add a changelog entry and a version bump for `persistent-mongoDB` and we'll get it released.

parsonsmatt (approved):
AWesome thanks! I'll have this uploaded today.

Issue #1416, opened 2022-08-09
Consider changing On{Update,Delete} FK behaviour from RESTRICT to NO ACTION


hdgarrood:
At the moment, the OnUpdate and OnDelete behaviour for a foreign key declared in a persistentmodels file defaults to RESTRICT if not specified. However, I think NO ACTION might be a better default: it matches all supported SQL databases' defaults, and for databases that support it, it offers a little more flexibility without losing data integrity guarantees (because the FK is still checked, just not until the end of the transaction, as opposed to immediately). 

* In MySQL, there is no difference between RESTRICT and NO ACTION: https://dev.mysql.com/doc/refman/5.6/en/create-table-foreign-keys.html#foreign-key-referential-actions, so changing the default will make no difference.
* In PostgreSQL, the difference is that RESTRICT checks immediately, whereas NO ACTION defers the check until the end of the database transaction. https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK
* In sqlite, NO ACTION is the default, and (as far as I can tell) it works the same as PostgreSQL: https://www.sqlite.org/foreignkeys.html#fk_actions

I didn't see much previous discussion of this in the change that initially implemented the ability to specify OnUpdate/OnDelete in persistentmodels files (#1145) but please do let me know if I've missed anything.

This change would of course have to wait for a breaking release if we were going to do it, and it would generate a number of new migrations for existing databases that were previously considered to have their DB schemas up to date, but I think we usually consider this acceptable in breaking releases?

parsonsmatt:
This seems like a reasonable change.

To make migration easier, probably best to have a new configuration option for "default foreign key action", which is used when nothing is specified. Then users can say "Continue using `RESTRICT`, I don't want to update this behavior."

ldub:
Should there be a way to specify `NO ACTION` in persistent? Reading [the docs](https://hackage.haskell.org/package/persistent-2.12.1.1/docs/Database-Persist-Quasi.html#:~:text=OnUpdate%20and%20OnDelete) I expected to see an `OnDeleteNoAction` and `OnUpdateNoAction` options.

hdgarrood:
Yeah, I think we'd need a new `NoAction` constructor for the `CascadeAction` data type first.

PR #1560, opened 2024-12-30
Make `NO ACTION` the default action for foreign keys

samtay:
See #1416 for motivation, but tl;dr is that `NO ACTION` is a more sensible default across the supported databases.

---

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql.hs
@@ -1193,7 +1193,7 @@ findAlters defs edef col@(Column name isNull sqltype def _gen _defConstraintName
                 )
 
 -- We check if we should alter a foreign key. This is almost an equality check,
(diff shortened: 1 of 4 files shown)


parsonsmatt:
Sorry I've taken so long to get to this!

I'm amenable to the rationale for the changes in this, however, I'd like to think through the upgrade path. If we release this as-is, then every `persistent` user is going to get a new migration suggestion to change their existing `RESTRICT` foreign key changes to `NO ACTION`, for every foreign key. While this may be preferable, it's also a pretty harsh UX. We'd essentially be requiring all users to modify all foreign key relationships and specify `Restrict` instead of using that as an implied default.

Instead, it'd be better to do this in phases:

1. Introduce the `NoAction` constructor and type, but `Restrict` is still default.
2. Allow the end user to customize the default in settings somewhere. Instead of using `Restrict` as a hardcoded default, we'll refer to the relevant settings record.
3. Switch the default to `NoAction` and publicize the change.

Issue #93, opened 2012-08-20
persistent should support schema-qualified table names


joeyadams:
My particular problem involves PostgreSQL and schema-qualified table names:

    CREATE TABLE acme.groups (
        id      SERIAL PRIMARY KEY,
        name    TEXT NOT NULL
    );

    CREATE TABLE acme.widgets (
        id      SERIAL PRIMARY KEY,
        "group" INT NOT NULL REFERENCES acme.groups(id),
        name    TEXT NOT NULL
    );

However, this won't do what I want:

    mkPersist sqlSettings [persist|
    Widget sql=acme.widgets
        group   GroupId
        name    Text
        deriving Show
    Group sql=acme.groups
        name    Text
        deriving Show
    |]

because "acme.widgets" is quoted as a single identifier, which Postgres treats as a single identifier, rather than a schema-qualified name.

Because my database uses schema-qualified names for everything, it looks like I can't use persistent with it.  This issue is related to [pull request #82][1], where someone wanted a similar feature to access nested MongoDB objects, where the schema was already defined.

The easy fix would be to update each backend's escape function to not escape `.`, but this would be technically incorrect (in PostgreSQL, a quoted identifier may include "." in its name).  The harder fix would be to support nested labels.

 [1]: https://github.com/yesodweb/persistent/pull/82

snoyberg:
Closing out old issues, please reopen if still relevant

gregwebs:
Issue is still present and should at least be documented

EvanCarroll:
Someone encountering this bug from StackOverflow: http://stackoverflow.com/q/43504368/124486

And seriously, Persistent doesn't support schemas? That's just basic sql-quantifier information. Everything has a catalog and a schema, implicit or not.

sbchaos:
I understand, the project was started to provide database access for Yesod web-apps. In a web-app the tables are generally present in single schema.

In micro-service based architectures,  the tables are likely to be sitting in different schema. Which makes persistent not so friendly in those cases.

JavierJF:
I wanted help with this issue, sadly the most I have been able to do is tracking the parsing of the "sql=" attribute to [this file](https://github.com/yesodweb/persistent/blob/master/persistent/Database/Persist/Quasi.hs). The logic involved doesn't look trivial and I don't find any documentation in which I could stand for starting the task. Anyone with higher knowledge of the internals that could enlighten me a bit?

parsonsmatt:
[...]

PR #1561, opened 2024-12-30
Add schema support to Persistent (issue #93)

curranosaurus:
Addresses https://github.com/yesodweb/persistent/issues/93
I worked on this together with @benjonesy.

This PR is in draft form, we're going to do a little more work on it still.

# Design challenges and decisions

## "Schema" means many things to many backends

In this PR, we use the word "schema" to refer to various devices that backends offer for namespacing tables.

- :heavy_check_mark: In PostgreSQL, a schema exists in between the level of a table and a database.
- :heavy_check_mark: In MySQL and SQLite, schemas and databases are synonymous concepts.
- :x: MongoDB and Redis do not have a concept analogous to schemas.

The semantics of schemas in the three relational backends are quite different, so Persistent's schema support will inherit these differences in semantics. For example, MySQL can enforce foreign key references across databases, but PostgreSQL and SQLite cannot. So a Persistent model with a cross-schema foreign key reference may lead to runtime errors with PostgreSQL and SQLite backends, but it will work fine with the MySQL backend.

## Schemas may or may not exist

We considered designing this PR so that Persistent would create schemas for the user if they did not already exist. But we chose not to do that. This means that local development with Persistent using schemas will require some manual SQL migrations to create the schemas that the user wants. But for conventional use-cases, schema creation will be a rare event, so this is not very onerous for developers. Below are some issues one would have to solve to make Persistent automate schema creation for the developer:

### Three-stage migrations

RIght now, the `CREATe TABLE` DDL statements are sorted to the top of the migration output, so that the (potentially recursive) foreign-key constraints are created only after all of the tables they reference exist. If we also wanted Persistent to ensure the existence of schemas, we would need to sort backend-appropriate versions of `CREATE SCHEMA` statements above the `CREATe TABLE` statements.

### Granting permissions in MySQL

Before a connection can use a newly created database in MySQL, one must run a `GRANT` statement to give the current user the requisite DDL operations to run `CREATe TABLE` statements. We would somehow need to make Persistent aware of the 

### File name and directory in SQLite

SQLite's invocation for creating a new database also specifies the file name and location for that database's `.db` file. Persistent would need to be opinionated about this directory scheme, or we would have to add a configuration option, if we wanted to support schema creation in SQLite.

# PR Template checklist

After submitting your PR:

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -151,7 +151,7 @@ openMySQLConn ci logFunc = do
                 , connCommit     = const $ MySQL.commit   conn
                 , connRollback   = const $ MySQL.rollback conn
                 , connEscapeFieldName = T.pack . escapeF
-                , connEscapeTableName = T.pack . escapeE . getEntityDBName
+                , connEscapeTableName = \ent -> escapeET (getEntityDBName ent) (getEntitySchema ent)
                 , connEscapeRawName = T.pack . escapeDBName . T.unpack
                 , connNoLimit    = "LIMIT 18446744073709551615"
                 -- This noLimit is suggested by MySQL's own docs, see
@@ -192,7 +192,7 @@ insertSql' ent vals =
     (fieldNames, placeholders) = unzip (Util.mkInsertPlaceholders ent escapeFT)
     sql = T.concat
         [ "INSERT INTO "
-        , escapeET $ getEntityDBName ent
+        , escapeET (getEntityDBName ent) (getEntitySchema ent)
         , "("
         , T.intercalate "," fieldNames
         , ") VALUES("
@@ -359,6 +359,7 @@ migrate' :: MySQL.ConnectInfo
          -> IO (Either [Text] CautiousMigration)
 migrate' connectInfo allDefs getter val = do
     let name = getEntityDBName val
+    let schema = getEntitySchema val
     let (newcols, udefs, fdefs) = mysqlMkColumns allDefs val
     old <- getColumns connectInfo getter val newcols
     let udspair = map udToPair udefs
@@ -368,7 +369,7 @@ migrate' connectInfo allDefs getter val = do
             let uniques = do
                     (uname, ucols) <- udspair
                     pure
-                        $ AlterTable name
+                        $ AlterTable name schema
                         $ AddUniqueConstraint uname
                         $ map (findTypeAndMaxLen name) ucols
 
@@ -376,11 +377,12 @@ migrate' connectInfo allDefs getter val = do
                     Column { cName=cname, cReference=Just cRef } <- newcols
                     let refConstraintName = crConstraintName cRef
                     let refTblName = crTableName cRef
+                    let refSchmName = crSchemaName cRef
                     let refTarget =
-                          addReference allDefs refConstraintName refTblName cname (crFieldCascade cRef)
+                          addReference allDefs refConstraintName refTblName refSchmName cname (crFieldCascade cRef)
 
                     guard $ Just cname /= fmap fieldDB (getEntityIdField val)
-                    return $ AlterColumn name refTarget
+                    return $ AlterColumn name schema refTarget
 
 
             let foreignsAlt =
@@ -393,8 +395,10 @@ migrate' connectInfo allDefs getter val = do
                             in
                                 AlterColumn
                                     name
+                                    schema
                                     (AddReference
                                         (foreignRefTableDBName fdef)
+                                        (foreignRefSchemaDBName fdef)
                                         (foreignConstraintNameDBName fdef)
                                         childfields
                                         parentfields
@@ -432,9 +436,9 @@ migrate' connectInfo allDefs getter val = do
                         $ partitionEithers
                         $ old'
                 acs' =
-                    map (AlterColumn name) acs
+                    map (AlterColumn name schema) acs
                 ats' =
-                    map (AlterTable  name) ats
+                    map (AlterTable  name schema) ats
             return
                 $ Right
                 $ map showAlterDb
@@ -455,7 +459,7 @@ addTable :: [Column] -> EntityDef -> AlterDB
 addTable cols entity = AddTable $ concat
     -- Lower case e: see Database.Persist.Sql.Migration
     [ "CREATe TABLE "
-    , escapeE name
+    , escapeE name schema
     , "("
     , idtxt
     , if null nonIdCols then [] else ","
@@ -467,6 +471,8 @@ addTable cols entity = AddTable $ concat
         filter (\c -> Just (cName c) /= fmap fieldDB (getEntityIdField entity) ) cols
     name =
         getEntityDBName entity
+    schema =
+        getEntitySchema entity
     idtxt =
         case getEntityId entity of
             EntityIdNaturalKey pdef ->
@@ -549,12 +555,14 @@ addReference
     -- ^ Foreign key name
     -> EntityNameDB
     -- ^ Referenced table name
+    -> (Maybe SchemaNameDB)
+    -- ^ Referenced schema name
     -> FieldNameDB
     -- ^ Column name
     -> FieldCascade
     -> AlterColumn
-addReference allDefs fkeyname reftable cname fc =
-    AddReference reftable fkeyname [cname] referencedColumns fc
+addReference allDefs fkeyname reftable refschema cname fc =
+    AddReference reftable refschema fkeyname [cname] referencedColumns fc
   where
     errorMessage =
         error
@@ -562,7 +570,7 @@ addReference allDefs fkeyname reftable cname fc =
             ++ " (allDefs = " ++ show allDefs ++ ")"
     referencedColumns =
         fromMaybe errorMessage $ do
-            entDef <- find ((== reftable) . getEntityDBName) allDefs
+            entDef <- find (\e -> getEntityDBName e == reftable && getEntitySchema e == refschema) allDefs
             return $ map fieldDB $ NEL.toList $ getEntityKeyFields entDef
 
 data AlterColumn = Change Column
@@ -576,6 +584,7 @@ data AlterColumn = Change Column
                  -- | See the definition of the 'showAlter' function to see how these fields are used.
                  | AddReference
                     EntityNameDB -- Referenced table
+                    (Maybe SchemaNameDB) -- Referenced table schema
                     ConstraintNameDB -- Foreign key name
                     [FieldNameDB] -- Referencing columns
                     [FieldNameDB] -- Referenced columns
@@ -588,8 +597,8 @@ data AlterTable = AddUniqueConstraint ConstraintNameDB [(FieldNameDB, FieldType,
                 deriving Show
 
 data AlterDB = AddTable String
-             | AlterColumn EntityNameDB AlterColumn
-             | AlterTable EntityNameDB AlterTable
+             | AlterColumn EntityNameDB (Maybe SchemaNameDB) AlterColumn
+             | AlterTable EntityNameDB (Maybe SchemaNameDB) AlterTable
              deriving Show
 
 
@@ -650,7 +659,7 @@ getColumns connectInfo getter def cols = do
       where ref rs c = case cReference c of
                 Nothing -> rs
                 (Just r) -> (unFieldNameDB $ cName c, r) : rs
-    vals = [ PersistText $ pack $ MySQL.connectDatabase connectInfo
+    vals = [ PersistText $ fromMaybe (pack $ MySQL.connectDatabase connectInfo) $ fmap unSchemaNameDB $ getEntitySchema def
            , PersistText $ unEntityNameDB $ getEntityDBName def
         --   , PersistText $ unDBName $ fieldDB $ getEntityId def
            ]
@@ -659,7 +668,7 @@ getColumns connectInfo getter def cols = do
         where
           getIt row = fmap (either Left (Right . Left)) .
                       liftIO .
-                      getColumn connectInfo getter (getEntityDBName def) row $ ref
+                      getColumn connectInfo getter (getEntityDBName def) (getEntitySchema def) row $ ref
             where ref = case row of
                     (PersistText cname : _) -> (Map.lookup cname refMap)
                     _ -> Nothing
@@ -679,19 +688,20 @@ getColumn
     => MySQL.ConnectInfo
     -> (Text -> IO Statement)
     -> EntityNameDB
+    -> Maybe SchemaNameDB
     -> [PersistValue]
     -> Maybe ColumnReference
     -> IO (Either Text Column)
-getColumn connectInfo getter tname [ PersistText cname
-                                   , PersistText null_
-                                   , PersistText dataType
-                                   , PersistText colType
-                                   , colMaxLen
-                                   , colPrecision
-                                   , colScale
-                                   , default'
-                                   , generated
-                                   ] cRef =
+getColumn connectInfo getter tname tschema [ PersistText cname
+                                           , PersistText null_
+                                           , PersistText dataType
+                                           , PersistText colType
+                                           , colMaxLen
+                                           , colPrecision
+                                           , colScale
+                                           , default'
+                                           , generated
+                                           ] cRef =
     fmap (either (Left . pack) Right) $
     runExceptT $ do
         -- Default value
@@ -761,6 +771,7 @@ getColumn connectInfo getter tname [ PersistText cname
         -- Foreign key (if any)
         stmt <- lift . getter $ T.concat
             [ "SELECT KCU.REFERENCED_TABLE_NAME, "
+            ,   "KCU.TABLE_SCHEMA, "
             ,   "KCU.CONSTRAINT_NAME, "
             ,   "KCU.ORDINAL_POSITION, "
             ,   "DELETE_RULE, "
@@ -777,7 +788,7 @@ getColumn connectInfo getter tname [ PersistText cname
             ,   "KCU.COLUMN_NAME"
             ]
         let vars =
-                [ PersistText $ pack $ MySQL.connectDatabase connectInfo
+                [ PersistText $ fromMaybe (pack $ MySQL.connectDatabase connectInfo) $ fmap unSchemaNameDB tschema
                 , PersistText $ unEntityNameDB tname
                 , PersistText cname
                 , PersistText $ pack $ MySQL.connectDatabase connectInfo
@@ -795,14 +806,22 @@ getColumn connectInfo getter tname [ PersistText cname
 
         cntrs <- liftIO $ with (stmtQuery stmt vars) (\src -> runConduit $ src .| CL.consume)
         pure $ case cntrs of
-            [] ->
-                Nothing
-            [[PersistText tab, PersistText ref, PersistInt64 pos, PersistText onDel, PersistText onUpd]] ->
+            [] -> Nothing
+            [[PersistText tab, PersistText schema, PersistText ref, PersistInt64 pos, PersistText onDel, PersistText onUpd]] ->
                 if pos == 1
-                then Just $ ColumnReference (EntityNameDB tab) (ConstraintNameDB ref) FieldCascade
-                    { fcOnUpdate = parseCascadeAction onUpd
-                    , fcOnDelete = parseCascadeAction onDel
-                    }
+                then Just $
+                  let colSchema =
+                        if T.null schema || schema == (pack $ MySQL.connectDatabase connectInfo)
+                          then Nothing
+                          else Just $ SchemaNameDB schema
+                   in ColumnReference
+                        (EntityNameDB tab)
+                        colSchema
+                        (ConstraintNameDB ref)
+                        FieldCascade
+                          { fcOnUpdate = parseCascadeAction onUpd
+                          , fcOnDelete = parseCascadeAction onDel
+                          }
                 else Nothing
             xs -> error $ mconcat
               [ "MySQL.getColumn/getRef: error fetching constraints. Expected a single result for foreign key query for table: "
@@ -813,7 +832,7 @@ getColumn connectInfo getter tname [ PersistText cname
               , show xs
               ]
 
-getColumn _ _ _ x  _ =
+getColumn _ _ _ _ x _ =
     return $ Left $ pack $ "Invalid result from INFORMATION_SCHEMA: " ++ show x
 
 -- | Extra column information from MySQL schema
@@ -928,7 +947,8 @@ findAlters edef allDefs col@(Column name isNull type_ def gen _defConstraintName
                 Just cr ->
                     let tname = crTableName cr
                         cname = crConstraintName cr
-                        cnstr = [addReference allDefs cname tname name (crFieldCascade cr)]
+                        sname = crSchemaName cr
+                        cnstr = [addReference allDefs cname tname sname name (crFieldCascade cr)]
                     in
                         (Add' col : cnstr, cols)
         Column _ isNull' type_' def' gen' _defConstraintName' maxLen' ref' : _ ->
@@ -941,12 +961,12 @@ findAlters edef allDefs col@(Column name isNull type_ def gen _defConstraintName
                             []
                 refAdd  =
                     case (ref == ref', ref) of
-                        (False, Just ColumnReference {crTableName=tname, crConstraintName=cname, crFieldCascade = cfc })
+                        (False, Just ColumnReference {crTableName=tname, crSchemaName=sname, crConstraintName=cname, crFieldCascade = cfc })
                             | tname /= getEntityDBName edef
                             , Just idField <- getEntityIdField edef
                             , unConstraintNameDB cname /= unFieldNameDB (fieldDB idField)
                             ->
-                            [addReference allDefs cname tname name cfc]
+                            [addReference allDefs cname tname sname name cfc]
                         _ -> []
                 -- Type and nullability
                 modType | showSqlType type_ maxLen False `ciEquals` showSqlType type_' maxLen' False && isNull == isNull' = []
@@ -1010,7 +1030,7 @@ showColumn showReferences (Column n nu t def gen _defConstraintName maxLen ref) 
         Just cRef | showReferences ->
             mconcat
                 [ " REFERENCES "
-                , escapeE (crTableName cRef)
+                , escapeE (crTableName cRef) (crSchemaName cRef)
                 , " "
                 , T.unpack (renderFieldCascade (crFieldCascade cRef))
                 ]
@@ -1045,19 +1065,19 @@ showSqlType (SqlOther t) _        _     = T.unpack t
 -- | Render an action that must be done on the database.
 showAlterDb :: AlterDB -> (Bool, Text)
 showAlterDb (AddTable s) = (False, pack s)
-showAlterDb (AlterColumn t ac) =
-    (isUnsafe ac, pack $ showAlter t ac)
+showAlterDb (AlterColumn t s ac) =
+    (isUnsafe ac, pack $ showAlter t s ac)
   where
     isUnsafe Drop{} = True
     isUnsafe _      = False
-showAlterDb (AlterTable t at) = (False, pack $ showAlterTable t at)
+showAlterDb (AlterTable t s at) = (False, pack $ showAlterTable t s at)
 
 
 -- | Render an action that must be done on a table.
-showAlterTable :: EntityNameDB -> AlterTable -> String
-showAlterTable table (AddUniqueConstraint cname cols) = concat
+showAlterTable :: EntityNameDB -> Maybe SchemaNameDB -> AlterTable -> String
+showAlterTable table schema (AddUniqueConstraint cname cols) = concat
     [ "ALTER TABLE "
-    , escapeE table
+    , escapeE table schema
     , " ADD CONSTRAINT "
     , escapeC cname
     , " UNIQUE("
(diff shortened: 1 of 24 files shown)


persistent-sqlite/test/main.hs
```diff
@@ -37,13 +37,14 @@ import qualified MigrationColumnLengthTest
 import qualified MigrationOnlyTest
 import qualified MpsCustomPrefixTest
 import qualified MpsNoPrefixTest
-import qualified PersistUniqueTest
 import qualified PersistentTest
+import qualified PersistUniqueTest
```
curranosaurus, line 41: Sorted imports due to `stylish-haskell`.

persistent-sqlite/test/SqliteInit.hs
```diff
@@ -43,22 +45,42 @@ module SqliteInit (
...
+       , assertNotEqual
+       , isTravis
+       , liftA2
+       , truncateTimeOfDay
+       , truncateToMicro
+       , truncateUTCTime
+       , (==@)
+       , (@/=)
+       , (@==)
+       )
```
curranosaurus, line 65: Reformatted by `stylish-haskell`.

persistent-qq/test/Spec.hs
```diff
@@ -4,23 +4,23 @@
...
 
+import qualified CodeGenTest
 import Database.Persist.Class.PersistEntity
 import Database.Persist.Sql
 import Database.Persist.Sql.Raw.QQ
 import Database.Persist.Sqlite
-import PersistTestPetType
 import PersistentTestModels
-import qualified CodeGenTest
+import PersistTestPetType
```
curranosaurus, line 23: Imports sorted by `stylish-haskell`.

persistent-postgresql/test/main.hs
```diff
@@ -48,14 +48,15 @@ import qualified MigrationReferenceSpec
 import qualified MigrationTest
 import qualified MpsCustomPrefixTest
 import qualified MpsNoPrefixTest
-import qualified PersistUniqueTest
 import qualified PersistentTest
+import qualified PersistUniqueTest
```
curranosaurus, line 52: Imports sorted by `stylish-haskell`.

persistent-mysql/test/main.hs
```diff
@@ -46,8 +46,9 @@ import qualified MigrationOnlyTest
 import qualified MigrationTest
 import qualified MpsCustomPrefixTest
 import qualified MpsNoPrefixTest
-import qualified PersistUniqueTest
 import qualified PersistentTest
+import qualified PersistUniqueTest
```
curranosaurus, line 50: Imports sorted by `stylish-haskell`.

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -539,7 +541,7 @@ getCopyTable :: [EntityDef]
              -> EntityDef
              -> IO [(Bool, Text)]
 getCopyTable allDefs getter def = do
```
curranosaurus, line 543: The changes to this function were non-trivial. A naive change to this function would allow tables in different databases to clobber each other when copying them into the temporary schema. So we do some manual namespacing by prefixing the table name with `<schema>_` when copying it to the temporary schema.

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -470,6 +470,8 @@ migrate' allDefs getter val = do
   where
     def = val
     table = getEntityDBName def
+    schema = getEntitySchema def
+    sqliteMaster = maybe "sqlite_master" (\schema' -> escapeS schema' <> ".sqlite_master") schema
```
curranosaurus, line 474: There is one copy of the `sqlite_master` table in each schema, and it only contains information about the tables in its own schema.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -795,14 +808,25 @@ getColumn connectInfo getter tname [ PersistText cname
...
-            [[PersistText tab, PersistText ref, PersistInt64 pos, PersistText onDel, PersistText onUpd]] ->
+            [] -> Nothing
+            [[PersistText tab, schema, PersistText ref, PersistInt64 pos, PersistText onDel, PersistText onUpd]] ->
                 if pos == 1
-                then Just $ ColumnReference (EntityNameDB tab) (ConstraintNameDB ref) FieldCascade
-                    { fcOnUpdate = parseCascadeAction onUpd
-                    , fcOnDelete = parseCascadeAction onDel
-                    }
+                then Just $
+                  let colSchema =
```
curranosaurus, line 815: This is one of the most complex pieces of logic in the `persistent-mysql` diff, it's worth scrutinizing whether this makes sense.

curranosaurus:
I've run the test suites for the changed packages, and they all pass except for `persistent-mysql`. But the test failure I get there also occurs on `master` branch for me, so seems unrelated to these changes. Here is the failure output I get, for reference:

```
Failures:

  src/ForeignKey.hs:198:12: 
  1) foreign keys options deletes cascades with field self reference to the whole chain
       expected: []
        but got: [Entity {entityKey = Chain3Key {unChain3Key = SqlBackendKey {unSqlBackendKey = 92}}, entityVal = Chain3 {chain
3Name = 2, chain3Previous = Just (Chain3Key {unChain3Key = SqlBackendKey {unSqlBackendKey = 91}})}},Entity {entityKey = Chain3K
ey {unChain3Key = SqlBackendKey {unSqlBackendKey = 93}}, entityVal = Chain3 {chain3Name = 3, chain3Previous = Just (Chain3Key {
unChain3Key = SqlBackendKey {unSqlBackendKey = 92}})}}]

  To rerun use: --match "/foreign keys options/deletes cascades with field self reference to the whole chain/" --seed 6043809

Randomized with seed 6043809
```

persistent-qq/test/Spec.hs
```diff
@@ -37,6 +40,7 @@ runConn f = do
 db :: SqlPersistT (LoggingT (ResourceT IO)) a -> IO a
 db actions = do
   runResourceT $ runConn $ do
+      rawSql @(Single Int64) ("attach 'animals.db' as animals") []
```
curranosaurus, line 43: This ensures that the tests referencing the `animals` schema succeed.

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -661,7 +681,11 @@ sqlForeign fdef = T.concat $
     , " FOREIGN KEY("
     , T.intercalate "," $ map (escapeF . snd. fst) $ foreignFields fdef
     , ") REFERENCES "
-    , escapeE $ foreignRefTableDBName fdef
+    , -- It's a syntax error in SQLite to use a dot-qualified table name.
+      -- In general, it's not possible for SQLite to maintain foreign key
+      -- constraints across databases (which Persistent calls "schemas").
+      -- So we omit the schema here.
```
curranosaurus, line 687: Should we instead throw an exception here if there is a cross-schema foreign key reference?

curranosaurus: If we don't, one of two things will occur.

1. The table holding the fkey reference shares a schema with a table sharing the same name as the intended target. Then we will create a fkey reference to the _wrong_ table.
2. There is no such table sharing the schema with the origin table. In this case, `sqlite` will emit an error saying that there is no table with the name we're looking for in the origin table's schema.

As I write this out, case (1) seems to tell me that we _should_ throw an error if there is a cross-schema foreign key reference being attempted.

curranosaurus: How does the `persistent` team feel about using `error`? Can I use that to handle this case, or should I lift this function to a monad that can throw synchronous exceptions?

curranosaurus: I'm seeing some examples of `error` scattered around the library, so I'll go ahead with that approach.

curranosaurus:
I asked @parsonsmatt if he would be able to review this PR, but he is busy right now. He suggested that I ping you, @psibi and @snoyberg. Would either of you be available to take a look at it? Thanks!

persistent-test/src/SchemaTest.hs
```diff
@@ -0,0 +1,58 @@
...
+            SchemaEntity
+                { schemaEntityBar = 42
+                }
+        _ <- insert $
+            DefaultSchemaEntity
+                { defaultSchemaEntityBar = 43
+                }
+        rawBar  <- rawSql "SELECT bar FROM schema_entity" []
+        liftIO $ rawBar @?= [Single (43 :: Int)]
+        return ()
```
parsonsmatt, line 58: Very nice!

persistent-sqlite/test/SqliteInit.hs
```diff
@@ -99,9 +124,10 @@ runConn f = do
     let debugPrint = not travis && _debugOn
     let printDebug = if debugPrint then print . fromLogStr else void . return
     void $ flip runLoggingT (\_ _ _ s -> printDebug s) $ do
-        withSqlitePoolInfo sqlite_database 1 $ runSqlPool f
+        withSqlitePoolInfo sqlite_database 1 $ runSqlPool $ do
+          rawSql @(Single Int64) ("attach '" <> sqlite_foo_database_file <> "' as foo") []
```
parsonsmatt, line 128: hmm we may want to expose this functionality from `persistent-sqlite` directly so we can support this more

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -552,6 +554,9 @@ getCopyTable allDefs getter def = do
            , (False, dropTmp)
            ]
   where
+    tableInfo = case schema of
+        Nothing -> "table_info"
+        Just schema' -> escapeS schema' <> ".table_info"
```
parsonsmatt, line 559: nice

persistent-sqlite/persistent-sqlite.cabal
```diff
@@ -44,7 +44,7 @@ flag use-stat4
 
 library
     build-depends:   base                    >= 4.9         && < 5
-                   , persistent              >= 2.13.3      && < 3
+                   , persistent              >= 2.14.7      && < 3
```
parsonsmatt, line 47:
```suggestion
                   , persistent              >= 2.15.0      && < 3
```

I'll make Hackage revisions on these packages to put upper bounds of `2.15` on them

persistent-sqlite/test/SqliteInit.hs
parsonsmatt:
```haskell
attachDatabaseFile :: (MonadIO m) => Text -> SqlPersistT m Int64
attachDatabaseFile = ...
```

Along with a note that we need to use that?

persistent-sqlite/ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog for persistent-sqlite
 
+## 2.13.4
+
+* [#1561](https://github.com/yesodweb/persistent/pull/1561)
+    * Add schema support to Persistent.
```
parsonsmatt, line 6: I think this is actually a patch version: `2.13.3.1` since we aren't adding anything to the API.

If we add `attachDatabase` as a function + advertise it, then that would be a minor bump to `2.13.4`.

persistent-qq/persistent-qq.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 1.12
 name:           persistent-qq
-version:        2.12.0.6
+version:        2.12.1
```
parsonsmatt, line 3:
```suggestion
version:        2.12.0.7
```

Since API hasn't changed, we can do a patch bump

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -626,14 +626,19 @@ withStmt' conn query vals =
 
 doesTableExist :: (Text -> IO Statement)
                -> EntityNameDB
+               -> (Maybe SchemaNameDB)
+               -- ^ @since 2.13.7
```
parsonsmatt, line 630: This function isn't exported, so I don't think it will have any haddocks rendered. Changing the signature isn't breaking since it's not exported.

persistent-postgresql/ChangeLog.md
```diff
@@ -1,9 +1,11 @@
 # Changelog for persistent-postgresql
 
-## Unreleased
+## 2.13.7
```
parsonsmatt, line 3: Since the API didn't chagne, I think this is 
```suggestion
## 2.13.6.3
```

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -795,14 +808,25 @@ getColumn connectInfo getter tname [ PersistText cname
...
+                        case schema of
+                          PersistNull -> Nothing
+                          PersistText schemaName ->
+                            if T.null schemaName || schemaName == (pack $ MySQL.connectDatabase connectInfo)
+                              then Nothing
+                              else Just $ SchemaNameDB schemaName
```
parsonsmatt, line 821: it may be nicer to use `do` and `guard` to make this logic a little more linear:

```suggestion
                          PersistText schemaName -> do
                            guard $ not (T.null schemaName)
                            guard $ schemaName /= (pack $ MySQL.connectDatabase connectInfo)
                            pure $ SchemaNameDB schemaName
```

very much a style thing, please ignore if it doesn't feel better to you

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -795,14 +808,25 @@ getColumn connectInfo getter tname [ PersistText cname
...
+            [[PersistText tab, schema, PersistText ref, PersistInt64 pos, PersistText onDel, PersistText onUpd]] ->
                 if pos == 1
-                then Just $ ColumnReference (EntityNameDB tab) (ConstraintNameDB ref) FieldCascade
-                    { fcOnUpdate = parseCascadeAction onUpd
-                    , fcOnDelete = parseCascadeAction onDel
-                    }
+                then Just $
+                  let colSchema =
+                        case schema of
+                          PersistNull -> Nothing
```
parsonsmatt, line 817: I think this is missing some cases?

persistent-mysql/Database/Persist/MySQL.hs
parsonsmatt: also may want to have a fallthrough:
```suggestion
                          PersistText schemaName ->
                            if T.null schemaName || schemaName == (pack $ MySQL.connectDatabase connectInfo)
                              then Nothing
                              else Just $ SchemaNameDB schemaName
                          _ -> Nothing
```

persistent-mysql/persistent-mysql.cabal
```diff
@@ -1,5 +1,5 @@
 name:            persistent-mysql
-version:         2.13.1.5
+version:         2.13.2
```
parsonsmatt, line 2: No API changes, so we can do a patch: 
```suggestion
version:         2.13.1.6
```

persistent-mysql/persistent-mysql.cabal
```diff
@@ -28,7 +28,7 @@ extra-source-files: ChangeLog.md
 
 library
     build-depends:   base             >= 4.9      && < 5
-                   , persistent       >= 2.13.3   && < 3
+                   , persistent       >= 2.14.7   && < 3
```
parsonsmatt, line 31:
```suggestion
                   , persistent       >= 2.15   && < 3
```

parsonsmatt (commented):
The PR looks great to me, thank you!

Please update with the CI changes in master to get CI running.

The support in different databases feels interesting, but ultimately fine. I think there's further work to support this on redis and mong, even if it's just `mconcat [schemaName, "_", tableName]`. But I'll defer that decision to someone who is using those backends.

persistent-sqlite/test/SqliteInit.hs
curranosaurus: That's a great idea!

persistent-sqlite/ChangeLog.md
curranosaurus: Okay cool, I'll add `attachDatabase` as a function and mention it in the changelog here.

persistent-postgresql/Database/Persist/Postgresql.hs
curranosaurus: Okay, I'll delete this line then.

persistent/test/Database/Persist/THSpec.hs
```diff
@@ -71,14 +71,13 @@ import qualified Database.Persist.TH.SharedPrimaryKeySpec as SharedPrimaryKeySpe
 import qualified Database.Persist.TH.SumSpec as SumSpec
 import qualified Database.Persist.TH.ToFromPersistValuesSpec as ToFromPersistValuesSpec
 import qualified Database.Persist.TH.TypeLitFieldDefsSpec as TypeLitFieldDefsSpec
-
```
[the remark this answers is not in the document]
curranosaurus: Sure, I'll add it back!

persistent/Database/Persist/Sql/Internal.hs
```diff
@@ -178,27 +178,28 @@ mkColumns allDefs t overrides =
...
+            (tableName_, schema, _) <- ref c fe as
+            pure (tableName_, schema, ConstraintNameDB x)
         _ -> ref c fe as
 
 refName :: EntityNameDB -> FieldNameDB -> ConstraintNameDB
 refName (EntityNameDB table) (FieldNameDB column) =
     ConstraintNameDB $ Data.Monoid.mconcat [table, "_", column, "_fkey"]
 
-resolveTableName :: [EntityDef] -> EntityNameHS -> EntityNameDB
+resolveTableName :: [EntityDef] -> EntityNameHS -> (EntityNameDB, Maybe SchemaNameDB)
```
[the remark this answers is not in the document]
curranosaurus: I haven't made this change so far, because it doesn't seem to make a functional difference, and because this comment doesn't seem to clearly say you want the change. Let me know if you are sure you want me to do this. (I would be fine doing it if you wanted it.)

persistent-mysql/Database/Persist/MySQL.hs
curranosaurus: Thanks, I applied your suggested changes.

persistent/Database/Persist/Sql/Types.hs
```diff
@@ -36,9 +36,13 @@ data Column = Column
...
 data ColumnReference = ColumnReference
     { crTableName :: !EntityNameDB
-    -- ^ The table name that the
+    -- ^ The foreign table's name.
     --
     -- @since 2.11.0.0
+    , crSchemaName :: !(Maybe SchemaNameDB)
+    -- ^ The name of the schema that the foreign table belongs to.
+    --
+    -- @since 2.14.7
```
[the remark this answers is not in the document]
curranosaurus: I bumped the version.

persistent/test/Database/Persist/THSpec.hs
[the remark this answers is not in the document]
curranosaurus: Done

persistent-sqlite/test/SqliteInit.hs
curranosaurus: I added the function `attachDatabaseFile` to `Database.Persist.Sqlite` and called it from the handful of places in the test suite that I was using an `attach` command.

persistent-sqlite/ChangeLog.md
curranosaurus: Done

persistent-mysql/Database/Persist/MySQL.hs
curranosaurus: Fixed!

persistent-postgresql/Database/Persist/Postgresql.hs
curranosaurus: Deleted.

curranosaurus:
Alright, I think I made it through all of your requests, @parsonsmatt. This should be ready for another round of review. Thanks for your advice!

> I think there's further work to support this on redis and mong, even if it's just mconcat [schemaName, "_", tableName]. But I'll defer that decision to someone who is using those backends.

Yeah, I suspect you're right that people will want to apply this feature to those backends. But I don't know what namespacing conventions people will expect for those data stores. So I agree that we should leave that decision to a future contributor.

.github/workflows/haskell.yml
```diff
@@ -68,6 +68,10 @@ jobs:
           cabal-version: ${{ matrix.cabal }}
       - name: Check MySQL connection
         run: mysql -utest -ptest -h127.0.0.1 --port=33306 test -e "SELECT 1;"
+      - name: Create 'foo' schema in MySQL
+        run: mysql -utest -ptest -h127.0.0.1 --port=3306 test -e "CREATE DATABASE foo; GRANT ALL ON foo.* to 'test'@'localhost';"
```
parsonsmatt, line 72:
```suggestion
        run: mysql -utest -ptest -h127.0.0.1 --port=33306 test -e "CREATE DATABASE foo; GRANT ALL ON foo.* to 'test'@'localhost';"
```

curranosaurus:
Looks like the MySQL schema setup failed:
```

Run mysql -utest -ptest -h127.0.0.1 --port=33306 test -e "CREATE DATABASE foo; GRANT ALL ON foo.* to 'test'@'localhost';"
  mysql -utest -ptest -h127.0.0.1 --port=33306 test -e "CREATE DATABASE foo; GRANT ALL ON foo.* to 'test'@'localhost';"
  shell: /usr/bin/bash -e {0}
  env:
    CONFIG: --enable-tests --enable-benchmarks
Warning: arning] Using a password on the command line interface can be insecure.
ERROR 1044 (42000) at line 1: Access denied for user 'test'@'%' to database 'foo'
Error: Process completed with exit code 1.
```

I am guessing it's because the `test` user doesn't have rights to create databases or grant itself access to those databases. I'll probably need to rewrite the CLI invocation to log in as root user?

.github/workflows/haskell.yml
curranosaurus: Thanks

.github/workflows/haskell.yml
```diff
@@ -68,6 +68,10 @@ jobs:
           cabal-version: ${{ matrix.cabal }}
       - name: Check MySQL connection
         run: mysql -utest -ptest -h127.0.0.1 --port=33306 test -e "SELECT 1;"
+      - name: Create 'foo' schema in MySQL
+        run: mysql -uroot -ptest -h127.0.0.1 --port=33306 test -e "CREATE DATABASE foo; GRANT ALL ON foo.* to 'test'@'localhost';"
```
curranosaurus, line 72: I think `-uroot -ptest` will work because near the top of this file there's a declaration `MYSQL_ROOT_PASSWORD: test`.

curranosaurus:
```
mysql -uroot -ptest -h127.0.0.1 --port=33306 test -e "CREATE DATABASE foo; GRANT ALL ON foo.* to 'test'@'localhost';"
ERROR 1410 (42000) at line 1: You are not allowed to create a user with GRANT
```
Maybe the host isn't called `localhost`?

parsonsmatt:
This may be relevant: https://stackoverflow.com/questions/50177216/how-to-grant-all-privileges-to-root-user-in-mysql-8-0

curranosaurus:
My current theory is that the user `test@<?>` exists because of the `MYSQL_USER: test` declaration on line 32. But that instead of `<?>` being `localhost`, it should be `127.0.0.1`, matching the way that we log in to the db using the `mysql` command.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -650,7 +660,7 @@ getColumns connectInfo getter def cols = do
       where ref rs c = case cReference c of
                 Nothing -> rs
                 (Just r) -> (unFieldNameDB $ cName c, r) : rs
-    vals = [ PersistText $ pack $ MySQL.connectDatabase connectInfo
+    vals = [ PersistText $ fromMaybe (pack $ MySQL.connectDatabase connectInfo) $ fmap unSchemaNameDB $ getEntitySchema def
```
rjmholt, line 663:
```suggestion
    vals = [ PersistText $ maybe (pack $ MySQL.connectDatabase connectInfo) unSchemaNameDB $ getEntitySchema def
```

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -777,7 +790,7 @@ getColumn connectInfo getter tname [ PersistText cname
             ,   "KCU.COLUMN_NAME"
             ]
         let vars =
-                [ PersistText $ pack $ MySQL.connectDatabase connectInfo
+                [ PersistText $ fromMaybe (pack $ MySQL.connectDatabase connectInfo) $ fmap unSchemaNameDB tschema
```
rjmholt, line 793: Same `maybe` thing

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -795,14 +808,26 @@ getColumn connectInfo getter tname [ PersistText cname
 
         cntrs <- liftIO $ with (stmtQuery stmt vars) (\src -> runConduit $ src .| CL.consume)
         pure $ case cntrs of
-            [] ->
-                Nothing
-            [[PersistText tab, PersistText ref, PersistInt64 pos, PersistText onDel, PersistText onUpd]] ->
+            [] -> Nothing
```
rjmholt, line 811: This would arguably be easier to review if it weren't so inlined. A comment at the top or breaking it out into a named function would make it much clearer what the intended semantics are

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -795,14 +808,26 @@ getColumn connectInfo getter tname [ PersistText cname
...
+            [[PersistText tab, schema, PersistText ref, PersistInt64 pos, PersistText onDel, PersistText onUpd]] ->
                 if pos == 1
-                then Just $ ColumnReference (EntityNameDB tab) (ConstraintNameDB ref) FieldCascade
-                    { fcOnUpdate = parseCascadeAction onUpd
-                    , fcOnDelete = parseCascadeAction onDel
-                    }
+                then Just $
+                  let colSchema =
+                        case schema of
+                          PersistNull -> Nothing
```
rjmholt, line 817: We didn't get a schema name at all, so there's no schema

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -795,14 +808,26 @@ getColumn connectInfo getter tname [ PersistText cname
...
+                          PersistText schemaName -> do
+                            guard $ not (T.null schemaName)
+                            guard $ schemaName /= (pack $ MySQL.connectDatabase connectInfo)
+                            pure $ SchemaNameDB schemaName
```
rjmholt, line 821: We got a schema name. If it's empty or the default, we return nothing, otherwise we return the name.

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -795,14 +808,26 @@ getColumn connectInfo getter tname [ PersistText cname
...
+                          _ -> Nothing -- this should never happen
```
rjmholt, line 822: I suppose this is the right thing to do in library code that needs to be robust, but this is one of those Haskell antipatterns that irks me where we might silently throw away the fact that we're hitting a "should never happen" case for years before someone realises we quietly ate a bug with our value result

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -795,14 +808,26 @@ getColumn connectInfo getter tname [ PersistText cname
...
+                   in ColumnReference
+                        (EntityNameDB tab)
+                        colSchema
+                        (ConstraintNameDB ref)
+                        FieldCascade
+                          { fcOnUpdate = parseCascadeAction onUpd
+                          , fcOnDelete = parseCascadeAction onDel
+                          }
```
rjmholt, line 830: Yeah this looks correct

persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -974,7 +988,11 @@ getColumn getter tableName' [ PersistText columnName
         , cGenerated = fmap stripSuffixes generationExpression'
         , cDefaultConstraintName = Nothing
         , cMaxLen = Nothing
-        , cReference = fmap (\(a,b,c,d) -> ColumnReference a b (mkCascade c d)) ref
+        , -- The ColumnReference always has a non-null SchemaNameDB. The default schema name
+          -- in Postgres is "public", but Postgres doesn't know whether a table with
+          -- schema "public" was explicitly given that schema by the Persistent
+          -- app developer.
+          cReference = fmap (\(a,schemaName,c,d,e) -> ColumnReference a (Just schemaName) c (mkCascade d e)) ref
```
rjmholt, line 995: Given that one variable has been renamed here, it might be worthwhile just to rename all of them. If someone looks at this code later without the context of this particular diff, it's going to look really weird that most of the variables are just assigned letters

persistent-sqlite/Database/Persist/Sqlite.hs
rjmholt: This might be a good comment to put into the code

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -561,30 +567,38 @@ getCopyTable allDefs getter def = do
                 return $ name : names
             Just y -> error $ "Invalid result from PRAGMA table_info: " ++ show y
     table = getEntityDBName def
-    tableTmp = EntityNameDB $ unEntityNameDB table <> "_backup"
+    schema = getEntitySchema def
+    tableIdentifier = entityIdentifier def
+    -- Temporary tables cannot have qualified names, so we prepend
+    -- the name with the intended schema to provide namespacing.
+    tableTmp =
+        EntityNameDB $ T.intercalate "_" $ catMaybes [unSchemaNameDB <$> schema, Just $ unEntityNameDB table, Just "backup"]
```
rjmholt, line 575: Is the philosophy here that Persistent will provide ersatz schema support by using a name prefix here, even though sqlite itself doesn't allow schemas for temp tables?

I can imagine that's maybe defensible, but ideally Persistent would just reflect what the underlying implementation supports, even in partial cases like this (i.e. you need to know as an implementor that sqlite won't support this, so if you try adding it in Persistent, it will throw an error or something).

I don't know Persistent's general philosophy though, so I might be off base there.

Also, I don't really know how you get Persistent to interface with a temporary table anyway, and even then this feels like a corner case...

persistent-sqlite/Database/Persist/Sqlite.hs
```diff
@@ -646,27 +660,42 @@ sqlColumn noRef (Column name isNull typ def gen _cn _maxLen ref) = T.concat
...
     , escapeC $ foreignConstraintNameDBName fdef
     , " FOREIGN KEY("
     , T.intercalate "," $ map (escapeF . snd. fst) $ foreignFields fdef
     , ") REFERENCES "
-    , escapeE $ foreignRefTableDBName fdef
+    , -- It's not possible for SQLite to maintain foreign key
+      -- constraints across databases (which Persistent calls "schemas").
+      -- In fact, it's a syntax error to use a schema qualifier in this
+      -- part of the SQL expression. So we omit the schema here, and throw
+      -- an error if this is a cross-schema reference.
```
rjmholt, line 689: Yeah this bit feels a bit weird to me. Does sqlite actually support what we've referred to elsewhere as "schemas" (i.e. namespaces within the same database)? (Looks like [no](https://stackoverflow.com/a/33962375)?) Or are we grasping at some other sqlite functionality that we imagine could be made analogous to schemas? What happens if sqlite adds support for true schemas later?

Should sqlite maybe fall into the same category as MongoDB, where the actual backend just doesn't support them?

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -349,6 +352,9 @@ toParsedEntityDef lwc = ParsedEntityDef
     (attribs, extras) =
         parseEntityFields fieldLines
 
+    schemaName =
+      fmap SchemaNameDB $ listToMaybe $ mapMaybe (T.stripPrefix "schema=") entAttribs
```
rjmholt, line 356: Depending on what packages are available, `Data.List.Extra.firstJust` does this nicely

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -1390,6 +1397,9 @@ takeForeign ps entityName = takeRefTable
                                 EntityNameHS refTableName
                             , foreignRefTableDBName =
                                 EntityNameDB $ psToDBName ps refTableName
+                            , foreignRefSchemaDBName =
+                                Nothing
+                            -- ^ This will be determined in the TH phase ('fixForeignRefSchemaDBName').
```
rjmholt, line 1402: Even though this is written as a haddock comment, because it's in actual implementation code, people will have to read the internal module to see it.

If this is information about the lifecycle contract of the values on the record type, it might be worth hoisting this to something like the type definition so it's apparent in docs

persistent/Database/Persist/Sql/Internal.hs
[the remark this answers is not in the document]
rjmholt: Yeah I think a record would be nicer

persistent/Database/Persist/Sql/Types.hs
```diff
@@ -36,9 +36,13 @@ data Column = Column
 -- @since 2.11.0.0
 data ColumnReference = ColumnReference
     { crTableName :: !EntityNameDB
-    -- ^ The table name that the
+    -- ^ The foreign table's name.
```
rjmholt, line 39: It's not necessarily foreign right? You can have a self-referential FK. I guess it's not totally clear by what's meant by foreign in this context

Issue #1567, opened 2025-02-06
Include source file information when parsing `UnboundEntityDef`


parsonsmatt:
Also may be nice to have source location information for fields, unique identifiers, etc to make `TAGS` and other utilities easier to do

lf-:
Working on this :)

PR #1569, opened 2025-02-12
Add position info to EntityDef

lf-:
Internal Mercury bug: DUX-3027
Fixes: #1567 

This is an initial implementation that we can API-compatibly improve in
the future by providing more accurate information. However, to do that,
we would have to rewrite the entity definition parser with e.g.
megaparsec. This is a reasonable course of action but rewriting it is
going to be more work and we can ship position info we could improve
later with the existing parser.

Currently this gives users the entire file in which the model is defined, but not anything closer since the parser throws all that stuff out.

After submitting your PR:


lf-:
idk if i parsed the PVP correctly here, this might actually be non breaking so perhaps should be the C instead of B component, but it will blow up anyone using the internals of the quasiquoter/TH stuff.

persistent/Database/Persist/Types/Base.hs
```diff
@@ -155,6 +173,14 @@ data EntityDef = EntityDef
     -- ^ Optional comments on the entity.
     --
     -- @since 2.10.0
+    , entitySpan :: !(Maybe Span)
+    -- ^ Source code span occupied by this entity. May be absent if it is not
+    -- known.
```
parsonsmatt, line 178: I guess we wouldn't have this if we skip the quasiquoter entirely and write the instances ourselves. Valid

persistent/persistent.cabal
```diff
@@ -157,7 +157,7 @@ test-suite test
     hs-source-dirs:
         test/
 
-    ghc-options: -Wall
+    ghc-options: -Wall -Wno-incomplete-uni-patterns -dth-dec-file
```
parsonsmatt, line 160: What's `-dth-dec-file` do?

persistent/test/Database/Persist/THSpec.hs
```diff
@@ -76,6 +77,10 @@ import qualified Database.Persist.TH.TypeLitFieldDefsSpec as TypeLitFieldDefsSpe
 -- machinery
 type TextId = Text
 
+-- | Location above the block defining Person below. Must be before it. Do not move!
+-- Used to test TH definition positions are plausible.
+personDefBeforeLoc :: SourceLoc
+personDefBeforeLoc = $(TH.lift @_ @SourceLoc . sourceLocFromTHLoc =<< TH.location)
```
parsonsmatt, line 83: I think there may be a better way of doing this - 

```haskell
do
  firstLoc <- [d| firstLocation = $(TH.location) |]
  decls <- share [mkPersistWith sqlSettings ...] [persistUpperCase| ... |]
  secondLoc <- [d| secondLocation = $(TH.location) |]
  pure (concat [firstLoc, decls, secondLoc])
```

This would require a little less spooky magic

persistent-test/persistent-test.cabal
```diff
@@ -76,7 +76,7 @@ library
       , monad-logger             >= 0.3.25
       , mtl
       , path-pieces              >= 0.2
-      , persistent               >= 2.14        && < 2.15
+      , persistent               >= 2.15        && < 2.16
```
parsonsmatt, line 79: Since there are no source changes, we can probably just do

```suggestion
      , persistent               >= 2.14        && < 2.16
```

parsonsmatt (commented):
nice! a couple suggestions. don't need to worry about the version bump on `persistent-test`, once we release I'll revise the Hackage page and/or we can do a patch version bump

persistent/persistent.cabal
parsonsmatt: Ohhh that's super cool! Thanks for the reference 😄 

(full circle: i wrote the bug you diagnosed :joy:)

parsonsmatt:
Build failure is due to `persistent-th`'s benchmark looks like

persistent/test/Database/Persist/THSpec.hs
parsonsmatt: Ah good point. My apologies

It looks like this is failing on GHC 8.8 - complaining about `TH.lift @_ @SourceLoc`. My guess is that GHC 8.8 has only one type variable for `lift`

parsonsmatt:
Could you make an issue with your specific use case? In the meantime, you can depend on the internal module (it is exposed). The docs for it say you can define entities separately and then use that to make a migration. However, that *shouldn't* be necessary with the way `migrateModels` works now. I left a comment in the linked issue.

PR #1576, opened 2025-03-18
style: apply `cabal-fmt -i **/*.cabal`

ncaq:
Inconsistent order and uneven indentation existed not even on a file-by-file basis,
but even within files, causing warnings.
I suggest using `cabal-fmt` to let the machine take care of the non-essential annoyances and not worry about them.


.github/workflows/haskell.yml
```diff
@@ -91,3 +91,13 @@ jobs:
...
+  cabal-fmt:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: haskell-actions/setup@v2
+        with:
+          ghc-version: "9.6"
+          cabal-version: "3.6"
+      - run: cabal install cabal-fmt
+      - run: cabal-fmt --check **/*.cabal
```
parsonsmatt, line 103: It looks like we can add `cabal-fmt` to `restyled` - do you mind adding it there instead?

parsonsmatt (commented):
Thanks, this is great! If we can get this on `restyled` CI check then that would be fantastic.

.github/workflows/haskell.yml
ncaq: I was not sure which one to add, but I made the decision to add it to the workflow that runs cabal, based on the analogy that the restyled workflow may only want to run tools made by restyled.
Of course, I will make the change according to the maintainer's decision.

.github/workflows/restyled.yml
```diff
@@ -18,3 +18,13 @@ jobs:
...
+  cabal-fmt:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: haskell-actions/setup@v2
+        with:
+          ghc-version: "9.6"
+          cabal-version: "3.6"
+      - run: cabal install cabal-fmt
+      - run: cabal-fmt --check **/*.cabal
```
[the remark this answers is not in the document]
ncaq: I did not know that it was supported.
I assumed it was missing because it was not in the repository list.

parsonsmatt:
Wonderful, thank you!!

PR #1580, opened 2025-04-03
Rework Entity parsing implementation, foundation for comment improvements 

danbroooks:
This PR refactors the way that the initial entity definition blocks are parsed by the `parse` function before they then get turned into an `UnboundEntityDef`. This is an initial change that hopefully will make it easier to implement https://github.com/yesodweb/persistent/issues/1297 and https://github.com/yesodweb/persistent/issues/1296.

I have added an example test file that illustrates various possible comment styles and what the parse currently is or isn't picking up. This test can be used when tackling the issues linked above. The content of this test is here:

https://github.com/yesodweb/persistent/blob/54be9d68125984249587f81971269f1c7c599a4a/persistent/test/Database/Persist/TH/CommentsSpec.hs

I'm not sure exactly what behavior is expected for comments, so I'd welcome any input on what the behavior should be as I might have made some assumptions that are not desired for the library. But I'd assume everything I've included there persistent would want to be parsed as comments.

The only actual behavioral change regarding comments in this change is illustrated here:

https://github.com/yesodweb/persistent/commit/54be9d68125984249587f81971269f1c7c599a4a#diff-a2bb675c726f8d466d83a85584878a9053dd6304d85aedf0511b8affb8074c56

With this change, that style of comment on the fields is now picked up by the parser, everything else should be the same.

I've also re-worked some of the unit tests, as I've changed some of the types exposed by this module. I've tried to make the tests a bit more flexible as whenever making a change to this module there is often quite a lot to change in the tests. Hopefully `ParsedEntityDef` and `ParsedFieldDef` can better replace `LinesWithComments` which is a lot more low level. I appreciate the coverage you get from the lower level tests, so I have tried to balance things here between flexibility and coverage, I've been through all the tests I had to remove and ensured that we're still covered by a new test based around the types I've added here.

---

- [ ] Ran `fourmolu` on any changed files (`restyled` will do this for you, so
- [ ] Adhered to the code style (see the `.editorconfig` and `fourmolu.yaml` files for details)

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -20,16 +20,13 @@ module Database.Persist.Quasi.Internal
     , upperCaseSettings
     , lowerCaseSettings
     , toFKNameInfixed
-    , Token (..)
     , Line (..)
+    , Comment(..)
     , SourceLoc(..)
     , sourceLocFromTHLoc
     , preparse
     , parseLine
     , parseFieldType
-    , associateLines
-    , LinesWithComments(..)
-    , parseEntityFields
     , takeColsEx
     -- * UnboundEntityDef
     , UnboundEntityDef(..)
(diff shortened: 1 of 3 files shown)


parsonsmatt:
Oh, shoot, I'm sorry - we are about to land a PR that will port the parser to `megaparsec`

danbroooks:
Ha 😅 no worries, I think that is for the best definitely.

PR #1584, opened 2025-04-12
use Megaparsec for first-pass parsing of entity definitions

dtpowl:
This PR uses Megaparsec to reimplement the tokenization and initial parsing of entity definitions.

**It also slightly changes the semantics of documentation comments**; it is now possible to write Haddock-style pre-comment blocks:

```
-- | This is a multi-line documentation block 
-- in the style of Haddock pre-comments.

-- | This is a multi-line documentation block
-- | in the old Persistent style, which is still
-- | supported.
```

Other than this, there are no intentional changes in behavior. Before merging this, I think it would be a very good idea to test that some other packages depending on persistent (e.g. Esqueleto) build and run correctly with these changes. **I have not done this yet**, but I think I should open the PR anyway so I can get feedback.

---

- [x] Ran `fourmolu` on any changed files (`restyled` will do this for you, so
- [ ] Adhered to the code style (see the `.editorconfig` and `fourmolu.yaml` files for details)

After submitting your PR:

### DIFF
--- a/persistent/Database/Persist/EntityDef.hs
+++ b/persistent/Database/Persist/EntityDef.hs
@@ -19,7 +19,7 @@ module Database.Persist.EntityDef
     , getEntityKeyFields
     , getEntityComments
     , getEntityExtra
-    , getEntitySourceSpan
+    , getEntitySpan
     , isEntitySum
     , entityPrimary
     , entitiesPrimary
@@ -209,12 +209,12 @@ overEntityFields
 overEntityFields f ed =
     setEntityFields (f (getEntityFieldsDatabase ed)) ed
 
--- | Gets the 'SourceSpan' of the definition of the entity.
+-- | Gets the 'Source' of the definition of the entity.
 --
 -- Note that as of this writing the span covers the entire file or quasiquote
 -- where the item is defined due to parsing limitations. This may be changed in
 -- a future release to be more accurate.
 --
--- @since 2.16.0.0
-getEntitySourceSpan :: EntityDef -> Maybe SourceSpan
-getEntitySourceSpan = entitySourceSpan
+-- @since 2.15.0.0
+getEntitySpan :: EntityDef -> Maybe SourceSpan
+getEntitySpan = entitySpan
--- a/persistent/Database/Persist/EntityDef.hs
+++ b/persistent/Database/Persist/EntityDef.hs
@@ -19,7 +19,7 @@ module Database.Persist.EntityDef
     , getEntityKeyFields
     , getEntityComments
     , getEntityExtra
-    , getEntitySpan
+    , getEntitySourceSpan
     , isEntitySum
     , entityPrimary
     , entitiesPrimary
@@ -43,7 +43,7 @@ import Database.Persist.FieldDef
 
 import Database.Persist.Names
 import Database.Persist.Types.Base
-       (ForeignDef, Span, UniqueDef(..), entityKeyFields)
+       (ForeignDef, SourceSpan, UniqueDef(..), entityKeyFields)
 
 -- | Retrieve the list of 'UniqueDef' from an 'EntityDef'. This does not include
 -- a @Primary@ key, if one is defined. A future version of @persistent@ will
@@ -209,12 +209,12 @@ overEntityFields
 overEntityFields f ed =
     setEntityFields (f (getEntityFieldsDatabase ed)) ed
 
--- | Gets the 'Span' of the definition of the entity.
+-- | Gets the 'SourceSpan' of the definition of the entity.
 --
 -- Note that as of this writing the span covers the entire file or quasiquote
 -- where the item is defined due to parsing limitations. This may be changed in
 -- a future release to be more accurate.
 --
--- @since 2.15.0.0
-getEntitySpan :: EntityDef -> Maybe Span
-getEntitySpan = entitySpan
+-- @since 2.16.0.0
+getEntitySourceSpan :: EntityDef -> Maybe SourceSpan
+getEntitySourceSpan = entitySourceSpan
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE DeriveLift #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
@@ -51,6 +50,7 @@ import Prelude hiding (lines)
 import Control.Applicative (Alternative ((<|>)))
 import Control.Monad
 import Data.Char (isDigit, isLower, isSpace, isUpper, toLower)
+import Data.Foldable (toList)
 import Data.List (find, foldl')
 import Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.List.NonEmpty as NEL
@@ -309,10 +309,10 @@ data UnboundEntityDef
     -- the field?" yet, so we defer those to the Template Haskell execution.
     --
     -- @since 2.13.0.0
-    , unboundEntityDefSpan :: Maybe Span
+    , unboundEntityDefSourceSpan :: Maybe SourceSpan
     -- ^ The source code span of this entity in the models file.
     --
-    -- @since 2.15.0.0
+    -- @since 2.16.0.0
     }
     deriving (Eq, Ord, Show, Lift)
 
@@ -336,7 +336,7 @@ unbindEntityDef ed =
             ed
         , unboundEntityFields =
             map unbindFieldDef (entityFields ed)
-        , unboundEntityDefSpan = entitySpan ed
+        , unboundEntityDefSourceSpan = entitySourceSpan ed
         }
 
 -- | Returns the @['UnboundFieldDef']@ for an 'UnboundEntityDef'. This returns
@@ -547,7 +547,7 @@ mkUnboundEntityDef ps parsedEntDef =
                     DefaultKey (FieldNameDB $ psIdName ps)
         , unboundEntityFields =
             cols
-        , unboundEntityDefSpan = parsedEntityDefSpan parsedEntDef
+        , unboundEntityDefSourceSpan = parsedEntityDefSourceSpan parsedEntDef
         , unboundEntityDef =
             EntityDef
                 { entityHaskell = entNameHS
@@ -571,7 +571,7 @@ mkUnboundEntityDef ps parsedEntDef =
                     case parsedEntityDefComments parsedEntDef of
                         [] -> Nothing
                         comments -> Just (T.unlines comments)
-                , entitySpan = parsedEntityDefSpan parsedEntDef
+                , entitySourceSpan = parsedEntityDefSourceSpan parsedEntDef
                 }
         }
   where
@@ -581,19 +581,11 @@ mkUnboundEntityDef ps parsedEntDef =
     attribs =
         parsedEntityDefFieldAttributes parsedEntDef
 
-    fieldComments =
-        parsedEntityDefFieldComments parsedEntDef
-
     cols :: [UnboundFieldDef]
-    cols = foldMap (f . commentedField ps) (zip attribs fieldComments)
-      where
-        f = \case
-            Just unb -> [unb]
-            _ -> []
+    cols = foldMap (toList . commentedField ps) attribs
 
     textAttribs :: [[Text]]
-    textAttribs =
-        fmap tokenContent <$> attribs
+    textAttribs = fmap tokenContent . fst <$> attribs
 
     entityConstraintDefs =
         foldMap
@@ -612,6 +604,14 @@ mkUnboundEntityDef ps parsedEntDef =
             SetOnce a -> Just a
             NotSet -> Nothing
 
+    commentedField
+        :: PersistSettings
+        -> ([Token], Maybe Text)
+        -> Maybe UnboundFieldDef
+    commentedField ps (tokens, mCommentText) = do
+        unb <- takeColsEx ps (tokenContent <$> tokens)
+        pure $ unb{unboundFieldComments = mCommentText}
+
     autoIdField :: FieldDef
     autoIdField =
         mkAutoIdField ps entNameHS idSqlType
@@ -619,12 +619,6 @@ mkUnboundEntityDef ps parsedEntDef =
     idSqlType :: SqlType
     idSqlType =
         maybe SqlInt64 (const $ SqlOther "Primary Key") primaryComposite
-
-commentedField
-    :: PersistSettings -> ([Token], Maybe Text) -> Maybe UnboundFieldDef
-commentedField ps (tokens, mCommentText) = do
-    unb <- takeColsEx ps (tokenContent <$> tokens)
-    pure $ unb{unboundFieldComments = mCommentText}
 
 defaultIdName :: PersistSettings -> FieldNameDB
 defaultIdName = FieldNameDB . psIdName
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -22,6 +22,8 @@ module Database.Persist.Quasi.Internal
     , sourceLocFromTHLoc
     , parseFieldType
     , takeColsEx
+    , CumulativeParseResult (..)
+    , renderErrors
 
       -- * UnboundEntityDef
     , UnboundEntityDef (..)
@@ -220,11 +222,14 @@ sourceLocFromTHLoc Loc{loc_filename = filename, loc_start = start} =
         }
 
 -- | Parses a quasi-quoted syntax into a list of entity definitions.
-parse :: PersistSettings -> [(Maybe SourceLoc, Text)] -> [UnboundEntityDef]
-parse ps = foldMap $ uncurry parseChunk
+parse
+  :: PersistSettings
+  -> [(Maybe SourceLoc, Text)]
+  -> CumulativeParseResult [UnboundEntityDef]
+parse ps inputs = foldMap toCumulativeParseResult $ map (uncurry parseChunk) inputs
   where
-    parseChunk :: Maybe SourceLoc -> Text -> [UnboundEntityDef]
-    parseChunk mSourceLoc source = fmap (mkUnboundEntityDef ps) (parseSource mSourceLoc source)
+    parseChunk :: Maybe SourceLoc -> Text -> ParseResult [UnboundEntityDef]
+    parseChunk mSourceLoc source = (fmap . fmap) (mkUnboundEntityDef ps) (parseSource mSourceLoc source)
 
 entityNamesFromParsedDef
     :: PersistSettings -> ParsedEntityDef -> (EntityNameHS, EntityNameDB)
@@ -608,8 +613,8 @@ mkUnboundEntityDef ps parsedEntDef =
         :: PersistSettings
         -> ([Token], Maybe Text)
         -> Maybe UnboundFieldDef
-    commentedField ps (tokens, mCommentText) = do
-        unb <- takeColsEx ps (tokenContent <$> tokens)
+    commentedField s (tokens, mCommentText) = do
+        unb <- takeColsEx s (tokenContent <$> tokens)
         pure $ unb{unboundFieldComments = mCommentText}
 
     autoIdField :: FieldDef
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -226,7 +226,7 @@ parse
   :: PersistSettings
   -> [(Maybe SourceLoc, Text)]
   -> CumulativeParseResult [UnboundEntityDef]
-parse ps inputs = foldMap toCumulativeParseResult $ map (uncurry parseChunk) inputs
+parse ps = foldMap $ toCumulativeParseResult . uncurry parseChunk
   where
     parseChunk :: Maybe SourceLoc -> Text -> ParseResult [UnboundEntityDef]
     parseChunk mSourceLoc source = (fmap . fmap) (mkUnboundEntityDef ps) (parseSource mSourceLoc source)
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -309,10 +309,10 @@ data UnboundEntityDef
     -- the field?" yet, so we defer those to the Template Haskell execution.
     --
     -- @since 2.13.0.0
-    , unboundEntityDefSourceSpan :: Maybe SourceSpan
+    , unboundEntityDefSpan :: Maybe SourceSpan
     -- ^ The source code span of this entity in the models file.
     --
-    -- @since 2.16.0.0
+    -- @since 2.15.0.0
     }
     deriving (Eq, Ord, Show, Lift)
 
@@ -336,7 +336,7 @@ unbindEntityDef ed =
             ed
         , unboundEntityFields =
             map unbindFieldDef (entityFields ed)
-        , unboundEntityDefSourceSpan = entitySourceSpan ed
+        , unboundEntityDefSpan = entitySpan ed
         }
 
 -- | Returns the @['UnboundFieldDef']@ for an 'UnboundEntityDef'. This returns
@@ -547,7 +547,7 @@ mkUnboundEntityDef ps parsedEntDef =
                     DefaultKey (FieldNameDB $ psIdName ps)
         , unboundEntityFields =
             cols
-        , unboundEntityDefSourceSpan = parsedEntityDefSourceSpan parsedEntDef
+        , unboundEntityDefSpan = parsedEntityDefSpan parsedEntDef
         , unboundEntityDef =
             EntityDef
                 { entityHaskell = entNameHS
@@ -571,7 +571,7 @@ mkUnboundEntityDef ps parsedEntDef =
                     case parsedEntityDefComments parsedEntDef of
                         [] -> Nothing
                         comments -> Just (T.unlines comments)
-                , entitySourceSpan = parsedEntityDefSourceSpan parsedEntDef
+                , entitySpan = parsedEntityDefSpan parsedEntDef
                 }
         }
   where
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -24,7 +24,6 @@ module Database.Persist.Quasi.Internal
     , takeColsEx
     , CumulativeParseResult (..)
     , renderErrors
-    , cumulativeData
 
       -- * UnboundEntityDef
     , UnboundEntityDef (..)
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -24,6 +24,7 @@ module Database.Persist.Quasi.Internal
     , takeColsEx
     , CumulativeParseResult (..)
     , renderErrors
+    , cumulativeData
 
       -- * UnboundEntityDef
     , UnboundEntityDef (..)
@@ -226,10 +227,11 @@ parse
   :: PersistSettings
   -> [(Maybe SourceLoc, Text)]
   -> CumulativeParseResult [UnboundEntityDef]
-parse ps = foldMap $ toCumulativeParseResult . uncurry parseChunk
+parse ps chunks = toCumulativeParseResult $ map parseChunk chunks
   where
-    parseChunk :: Maybe SourceLoc -> Text -> ParseResult [UnboundEntityDef]
-    parseChunk mSourceLoc source = (fmap . fmap) (mkUnboundEntityDef ps) (parseSource mSourceLoc source)
+    parseChunk :: (Maybe SourceLoc, Text) -> ParseResult [UnboundEntityDef]
+    parseChunk (mSourceLoc, source) =
+      (fmap . fmap) (mkUnboundEntityDef ps) (parseSource mSourceLoc source)
 
 entityNamesFromParsedDef
     :: PersistSettings -> ParsedEntityDef -> (EntityNameHS, EntityNameDB)
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -1,11 +1,9 @@
-{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DeriveLift #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE StrictData #-}
-{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE ViewPatterns #-}
 
@@ -21,58 +19,53 @@ module Database.Persist.Quasi.Internal
     , lowerCaseSettings
     , toFKNameInfixed
     , Token (..)
-    , Line (..)
-    , SourceLoc(..)
+    , SourceLoc (..)
     , sourceLocFromTHLoc
-    , preparse
-    , parseLine
     , parseFieldType
-    , associateLines
-    , LinesWithComments(..)
-    , parseEntityFields
     , takeColsEx
-    -- * UnboundEntityDef
-    , UnboundEntityDef(..)
+
+      -- * UnboundEntityDef
+    , UnboundEntityDef (..)
     , getUnboundEntityNameHS
     , unbindEntityDef
     , getUnboundFieldDefs
-    , UnboundForeignDef(..)
+    , UnboundForeignDef (..)
     , getSqlNameOr
-    , UnboundFieldDef(..)
-    , UnboundCompositeDef(..)
-    , UnboundIdDef(..)
+    , UnboundFieldDef (..)
+    , UnboundCompositeDef (..)
+    , UnboundIdDef (..)
     , unbindFieldDef
     , isUnboundFieldNullable
     , unboundIdDefToFieldDef
-    , PrimarySpec(..)
+    , PrimarySpec (..)
     , mkAutoIdField'
-    , UnboundForeignFieldList(..)
-    , ForeignFieldReference(..)
+    , UnboundForeignFieldList (..)
+    , ForeignFieldReference (..)
     , mkKeyConType
     , isHaskellUnboundField
-    , FieldTypeLit(..)
+    , FieldTypeLit (..)
     ) where
 
 import Prelude hiding (lines)
 
-import Control.Applicative (Alternative((<|>)))
+import Control.Applicative (Alternative ((<|>)))
 import Control.Monad
 import Data.Char (isDigit, isLower, isSpace, isUpper, toLower)
 import Data.List (find, foldl')
-import Data.List.NonEmpty (NonEmpty(..))
+import Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.List.NonEmpty as NEL
-import qualified Data.Map as M
 import Data.Maybe (fromMaybe, listToMaybe, mapMaybe)
 import Data.Monoid (mappend)
 import Data.Text (Text)
 import qualified Data.Text as T
 import Database.Persist.EntityDef.Internal
+import Database.Persist.Quasi.Internal.ModelParser
 import Database.Persist.Types
 import Database.Persist.Types.Base
-import Language.Haskell.TH.Syntax (Lift, Loc(..))
+import Language.Haskell.TH.Syntax (Lift, Loc (..))
 import qualified Text.Read as R
 
-data ParseState a = PSDone | PSFail String | PSSuccess a Text deriving Show
+data ParseState a = PSDone | PSFail String | PSSuccess a Text deriving (Show)
 
 parseFieldType :: Text -> Either String FieldType
 parseFieldType t0 =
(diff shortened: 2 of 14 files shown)


dtpowl:
This PR is bigger than I thought it would be. I think I rearranged some things in `Quasi/Internal.hs` that is making the diff bigger than it needs to be. I'll see if I can change those back and make this easier to review.

dtpowl:
> This PR is bigger than I thought it would be. I think I rearranged some things in `Quasi/Internal.hs` that is making the diff bigger than it needs to be. I'll see if I can change those back and make this easier to review.

Ok, I've done this. Still a big diff — fourmolu made a lot of adjustments to `Quasi/Internal.hs`.

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -767,12 +581,24 @@ mkUnboundEntityDef ps parsedEntDef =
...
 
+    fieldComments =
+        parsedEntityDefFieldComments parsedEntDef
+
+    cols :: [UnboundFieldDef]
+    cols = foldMap (f . commentedField ps) (zip attribs fieldComments)
+      where
+        f = \case
+            Just unb -> [unb]
+            _ -> []
```
parsonsmatt, line 592: This is `Data.Foldable.toList` if you want to make it a bit more concise

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -767,12 +581,24 @@ mkUnboundEntityDef ps parsedEntDef =
     attribs =
         parsedEntityDefFieldAttributes parsedEntDef
 
+    fieldComments =
+        parsedEntityDefFieldComments parsedEntDef
+
+    cols :: [UnboundFieldDef]
+    cols = foldMap (f . commentedField ps) (zip attribs fieldComments)
```
parsonsmatt, line 588: `zip` will truncate the output if one list is shorter than the other - `zip [1,2,3] ['a', 'b'] == [(1, 'a'), (2, 'b')]`. Do we have a guarantee here that `parsedEntityDefFieldComments` is a list of equal length in `parsedEntityDef`? If we do, then we may want to push that guarantee into the datatype itself.

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -797,6 +620,12 @@ mkUnboundEntityDef ps parsedEntDef =
     idSqlType =
         maybe SqlInt64 (const $ SqlOther "Primary Key") primaryComposite
 
+commentedField
+    :: PersistSettings -> ([Token], Maybe Text) -> Maybe UnboundFieldDef
+commentedField ps (tokens, mCommentText) = do
+    unb <- takeColsEx ps (tokenContent <$> tokens)
+    pure $ unb{unboundFieldComments = mCommentText}
```
parsonsmatt, line 627: The interface of this function is tightly coupled to its one call-site, which suggests it may not be a great top-level definition. Locating it a bit more closely in the `where` or a `let` as a helper function may be more appropriate.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+    , parseSource
+    , memberBlockAttrs
+    ) where
+
+import Prelude hiding (lines, unlines)
+import Data.List.NonEmpty (NonEmpty (..))
+import qualified Data.List.NonEmpty as NEL
+import qualified Data.Map as M
+import Data.Maybe
+import Data.Text (Text, pack, unlines, unpack)
```
parsonsmatt, line 29: IIt's more idiomatic to import text qualified, like:

```suggestion
import Data.Text (Text)
import qualified Data.Text as Text
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+    , parsedEntityDefEntityAttributes
+    , parsedEntityDefFieldAttributes
+    , parsedEntityDefFieldComments
+    , parsedEntityDefExtras
+    , parsedEntityDefSpan
+    , parseSource
+    , memberBlockAttrs
+    ) where
+
+import Prelude hiding (lines, unlines)
```
parsonsmatt, line 24: With the qualified import of `Text`, we don't have to have this line
```suggestion
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+import Language.Haskell.TH.Syntax (Lift)
+import Replace.Megaparsec (sepCap)
+import Text.Megaparsec hiding (Token)
+import Text.Megaparsec.Char
+import qualified Text.Megaparsec.Char.Lexer as L
+
+type Parser = Parsec Void String
+
+-- | Source location: file and line/col information. This is half of a 'Span'.
+data SourceLoc = SourceLoc
```
parsonsmatt, line 42: Can you add @since comments?

```suggestion
-- | Source location: file and line/col information. This is half of a 'Span'.
--
-- @since 2.16.0.0
data SourceLoc = SourceLoc
```

Especially to anything that's getting exported or part of a public API

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+        , char ']'
+        , char '_'
+        , char '\''
+        , char '"'
+        , char '!'
+        , char '~'
+        , char '-'
+        , char ':'
+        , try escapedParen
+        , char '\\'
```
parsonsmatt, line 106: 🤔 This is an interesting quirk of the grammar - do we have tests for it, or any documentation on what it's doing?

Just from a code perspsective it may be nicer to factor this out to avoid backtracking - something like

```suggestion
        , do
          backslash <- char '\\'
          nextChar <- peek
          if nextChar == '(' || nextChar == ')' 
            then -- eat the next char and return it
            else pure backslash
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+charLiteral = label "literal character" $ do
+    r <- lookAhead (count' 1 10 anySingle)
+    case lexChar r of
+        Just (c, r') -> c <$ skipCount (length r - length r') anySingle
+        Nothing -> unexpected (errorTokens r)
+  where
+    errorTokens ts = Tokens $
+        case NEL.nonEmpty ts of
+            Just nel -> NEL.head nel :| []
+            Nothing -> error "unreachable" -- ts is known to be non-empty
```
parsonsmatt, line 124: I think we can do better here -

`NEL.nonEmpty :: [a] -> Maybe (NonEmpty a)`, right? So in `Just nel`, we already have a `nel :: NonEmpty a`. Doing `NEL.head nel :| []` is going to strip the first element off and create just a singleton `NonEmpty`, discarding every other element in the input list. That *feels* like maybe not what we want? At the very least, it suggests that our input may not be well formed here.

Let's imagine that we've made a mistake in our programming, and the `Nothing` case fires. We get `unreachable` as the only thing in our output to diagnose. Instead, let's encode the actual expectation in the message.

```suggestion
        case NEL.nonEmpty ts of
            Just nel -> NEL.head nel :| []
            Nothing -> error "errorTokens was somehow empty"
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+nonLineSpaceChar = choice [char ' ', char '\t']
+
+-- This is a replacement for `Text.Megaparsec.Char.Lexer.charLiteral`;
+-- it does nearly the same thing but additionally supports escaped parentheses.
+charLiteral :: Parser Char
+charLiteral = label "literal character" $ do
+    r <- lookAhead (count' 1 10 anySingle)
+    case lexChar r of
+        Just (c, r') -> c <$ skipCount (length r - length r') anySingle
+        Nothing -> unexpected (errorTokens r)
```
parsonsmatt, line 119: hm. so if we get `Nothing`, then `lexChar r` must have been called on `r = ""` as `lexChar "" = Nothing` is the only equation which returns `Nothing`. Which means that we are giving `errorTokens ""`, which *also* means that `errorTokens` will call `error "unreachable"` - because `NEL.nonEmpty "" = Nothing`.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+        ]
+
+nonLineSpaceChar :: Parser Char
+nonLineSpaceChar = choice [char ' ', char '\t']
+
+-- This is a replacement for `Text.Megaparsec.Char.Lexer.charLiteral`;
+-- it does nearly the same thing but additionally supports escaped parentheses.
+charLiteral :: Parser Char
+charLiteral = label "literal character" $ do
+    r <- lookAhead (count' 1 10 anySingle)
```
parsonsmatt, line 116: this function has me so confused lol

so we `lookAhead` somewhere between 1 and 10 characters. literally any character, so this parser will probably always succeed and return the next 10 characters. then we `lexChar` that, which does some backslash logic for escape characters. But we only ever actually look at 2 of the characters - if there's a `\`, then we look for another character that may be escaped, otherwise, we return the whole 10 character string.

And then we return the result of lexing the character, and *skip* the other characters that we lookAhead parsed.

So, naively - I think what I'd do here is something rather different.

```haskell
charLiteral = do
    char1 <- anySingle
    case char1 of
        '\' -> do
            char2 <- anySingle
            case char2 of
                  '(' -> pure '('
                  ')' -> pure ')'
                  -- etc
        _ -> pure char1
```

... but I'll be damned if your impl here isn't a modification of the [upstream `charLiteral`](https://www.stackage.org/haddock/lts-23.18/megaparsec-9.6.1/src/Text.Megaparsec.Char.Lexer.html#charLiteral)! So color me confused about why upstream implemented this so strangely.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+equality :: Parser Token
+equality = label "equality expression" $ do
+    L.lexeme spaceConsumer $ do
+        lhs <- some contentChar
+        _ <- char '='
+        rhs <-
+            choice
+                [ try quotation'
+                , try sqlLiteral
+                , try parentheticalInner
```
parsonsmatt, line 149: Is `try` necessary inside of `choice`? that feels vaguely like an antipattern - there's a blog post (and [reddit discussion](https://www.reddit.com/r/haskell/comments/25ulsv/ezyang_try_a_b_considered_harmful/)) about this pattern which suggests ways to potentially improve this

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+        , ptext
+        ]
+
+data ParsedEntityDef = ParsedEntityDef
+    { parsedEntityDefComments :: [Text]
+    , parsedEntityDefEntityName :: EntityNameHS
+    , parsedEntityDefIsSum :: Bool
+    , parsedEntityDefEntityAttributes :: [Attr]
+    , parsedEntityDefFieldAttributes :: [[Token]]
+    , parsedEntityDefFieldComments :: [Maybe Text]
```
parsonsmatt, line 237: Ah okay - so here I think we have a condition that the length of these two lists must be the same, where the index of the lists is how we associate the comments to the field. 

I'd suggest combining that here - the form `[([Token], Maybe Text)]` is a way of expressing that constraint in the types, which makes this "correct by construction", and all downstream consumers don't have to worry about any bugs.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+    , parsedEntityDefExtras :: M.Map Text [ExtraLine]
+    , parsedEntityDefSpan :: Maybe Span
```
parsonsmatt, line 239: y'know while we're at it, we may want to, uh, change the name of the `Span` type - wish I'd caught this earlier. maybe `SourceSpan` or something?

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+    , extraBlockMembers :: NonEmpty Member
+    }
+    deriving (Show)
+
+data BlockAttr = BlockAttr
+    { blockAttrDocCommentBlock :: Maybe DocCommentBlock
+    , blockAttrTokens :: [Token]
+    , blockAttrPos :: SourcePos
+    }
+    deriving (Show)
```
parsonsmatt, line 283: These datatypes are great

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,536 @@
...
+                , spanEndCol = unPos . sourceColumn $ entityBlockEndPos eb
+                }
+
+parseSource :: Maybe SourceLoc -> Text -> [ParsedEntityDef]
+parseSource mSourceLoc source =
+    case parseEntities filepath (unpack source) of
+        Right blocks -> toParsedEntityDef mSourceLoc <$> blocks
+        Left peb -> error $ errorBundlePretty peb
+  where
+    filepath = maybe "" locFile mSourceLoc
```
parsonsmatt, line 536: Instead of `error`, here, we should use `Either`. This will let us consume the `parseSource` and report an error in a better way.

Specifically, at compile-time, an `error` call has a really bad representation. Instead, we'd want to use `reportError`, which will render the error much nicer in GHC.

persistent/persistent.cabal
```diff
@@ -96,6 +99,7 @@ library
     Database.Persist.TH
     Database.Persist.TH.Internal
     Database.Persist.Types
+    Database.Persist.Types.Span
```
parsonsmatt, line 102: Can we put doc comments in this module? I think mostly just having `-- | @since 2.16.0.0` in everything exported here would be great

(and also maybe yeah uh renaming to `SourceSpan` or something that has fewer clashes)

parsonsmatt (commented):
very nice!!

persistent/Database/Persist/Quasi/Internal.hs
dtpowl: Yep, that's guaranteed — or at least is intended to be; I suppose I could always have a bug. Will look into representing that guarantee somehow

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: Yep, I'm on the same page with you. I figured there was a good reason for why they did it that way and I followed along, but I genuinely don't get it.

If you're _also_ confused, then that gives me more confidence that the upstream implementation is not good.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: Yeah, that would be fantastic

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: Yep, the grammar allows attributes like `(foo ba\\"r baz)`, where a quotation mark is escaped despite not being within something that looks like a quotation-delimited string. There are three tests that exercise this in different ways.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: I think these `try`s may simply be unnecessary. If that turns out to be true, I'll just delete them.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: (The strange NEL stuff above is also basically from upstream — although they used `fromMaybe` instead)

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: > Doing NEL.head nel :| [] is going to strip the first element off and create just a singleton NonEmpty, discarding every other element in the input list. That feels like maybe not what we want? 

So this function is a modification of `charLiteral` in Megaparsec, and it's doing this too. This _is_ what we want, I guess, because `unexpected` expects a NEL and only the first element of that list is meaningful in this context. The underlying question here is why it's looking ahead ten characters.

persistent/Database/Persist/Quasi/Internal.hs
dtpowl: Ah right, I remember now — the `ParsedEntityDef` type is already present, and I was trying to avoid making any changes to it. But it's not exposed, so it _is_ safe to change.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: (My guess is that it's because whoever wrote that code also copy/pasted it from someplace else that was handling longer escape sequences)

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: This is good but I'm going to do it in #1589, since I'm already doing a bunch of error handling stuff there.

persistent/Database/Persist/EntityDef.hs
```diff
@@ -209,12 +209,12 @@ overEntityFields
...
--- | Gets the 'Span' of the definition of the entity.
+-- | Gets the 'SourceSpan' of the definition of the entity.
 --
 -- Note that as of this writing the span covers the entire file or quasiquote
 -- where the item is defined due to parsing limitations. This may be changed in
 -- a future release to be more accurate.
 --
--- @since 2.15.0.0
-getEntitySpan :: EntityDef -> Maybe Span
-getEntitySpan = entitySpan
```
parsonsmatt, line 220: These are compatibility functions - let's leave them in place with the same names so migration is easier. We can deprecate in another version if we want to.

parsonsmatt: or, I guess, the reason we provide these top-level function aliases is so we can make changes that aren't quite as painfully breaking

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
parsonsmatt: TBC, it may be reasonable, but it is very strange. Keeping behavior mostly the same as upstream is reasonable unless we know why we may want to change that. To pursue further I'd write an issue upstream

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -497,10 +309,10 @@ data UnboundEntityDef
     -- the field?" yet, so we defer those to the Template Haskell execution.
     --
     -- @since 2.13.0.0
-    , unboundEntityDefSpan :: Maybe Span
+    , unboundEntityDefSourceSpan :: Maybe SourceSpan
```
parsonsmatt, line 312: kind of same, let's keep function name the same so it's not a breaking change - we can break more gently with warnings/deprecations

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -1329,38 +1087,38 @@ data UnboundForeignDef
...
-    | FieldListHasReferences (NonEmpty ForeignFieldReference)
-    -- ^ You can specify the exact columns you're referring to here, if they
-    -- aren't part of a primary key. Most databases expect a unique index on the
-    -- columns you refer to, but Persistent doesnt' check that.
-    --
-    -- @
-    -- User
-    --     Id           UUID default="uuid_generate_v1mc()"
-    --     name         Text
-    --
```
parsonsmatt, line 1346: tihs file is too fucking big goddamn

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,519 @@
...
+        , char '~'
+        , char '-'
+        , char ':'
+        , do
+            backslash <- char '\\'
+            nextChar <- lookAhead anySingle
+            if nextChar == '(' || nextChar == ')'
+                then single nextChar
+                else pure backslash
+        ]
```
parsonsmatt, line 109: beautiful

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,519 @@
...
+    pure . Comment . Text.pack $ str
+
+indentLevelBacktrack :: Parser () -> Parser Pos
+indentLevelBacktrack sc = lookAhead $ do
+    _ <- sc
+    L.indentLevel
+
+indented :: Parser () -> Pos -> (Parser a -> Parser b) -> Parser a -> Parser b
+indented sc ref f p =
+    f . try $ do
```
parsonsmatt, line 224: encourage folks to use `fmap` or apply functions themselves if they need rather than trivially applying it for them

```suggestion
indented :: Parser () -> Pos -> Parser a -> Parser a
indented sc ref p =
    try $ do
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,519 @@
...
+    _ <- (string "--" <|> string "#") <* hspace
+    str <- many (contentChar <|> nonLineSpaceChar)
+    pure . Comment . Text.pack $ str
+
+indentLevelBacktrack :: Parser () -> Parser Pos
+indentLevelBacktrack sc = lookAhead $ do
+    _ <- sc
+    L.indentLevel
+
+indented :: Parser () -> Pos -> (Parser a -> Parser b) -> Parser a -> Parser b
```
parsonsmatt, line 222: What does the `sc` parser do?

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,519 @@
...
+
+flatBlock :: Parser () -> (Parser a -> Parser b) -> Parser a -> Parser b
+flatBlock sc f p = do
+    ref <- indentLevelBacktrack sc
+    indented sc ref f p
+
+docCommentBlock :: Parser (Maybe DocCommentBlock)
+docCommentBlock = do
+    lines <-
+        flatBlock indentationConsumer many $ choice [docCommentLine, commentLine]
```
parsonsmatt, line 241: ok, hm

prior advice was to avoid trivial recomposition but you have `f` in `flatBlock` also, and then also in `docCommentBlock` - so to rewrite without `f` in `indented`, you'd need, hm,

```suggestion
flatBlock :: Parser () -> Parser a -> Parser a
flatBlock sc p = do
    ref <- indentLevelBacktrack sc
    indented sc ref p

docCommentBlock :: Parser (Maybe DocCommentBlock)
docCommentBlock = do
    lines <-
        flatBlock indentationConsumer $ many $ choice [docCommentLine, commentLine]
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,519 @@
...
+        , parenthetical
+        , ptext
+        ]
+
+data ParsedEntityDef = ParsedEntityDef
+    { parsedEntityDefComments :: [Text]
+    , parsedEntityDefEntityName :: EntityNameHS
+    , parsedEntityDefIsSum :: Bool
+    , parsedEntityDefEntityAttributes :: [Attr]
+    , parsedEntityDefFieldAttributes :: [([Token], Maybe Text)]
```
parsonsmatt, line 283: nice!!

persistent/Database/Persist/Types.hs
```diff
@@ -69,7 +69,6 @@ import Database.Persist.Types.Base
        , PersistValue(..)
        , ReferenceDef(..)
        , SqlType(..)
-       , Span(..)
```
parsonsmatt, line 72: bless't

persistent/Database/Persist/Types/Base.hs
```diff
@@ -157,14 +157,14 @@ data EntityDef = EntityDef
     -- ^ Optional comments on the entity.
     --
     -- @since 2.10.0
-    , entitySpan :: !(Maybe Span)
+    , entitySourceSpan :: !(Maybe SourceSpan)
```
parsonsmatt, line 160: this is internal and should not be used by folks, but keeping the old name would be *less* breaking. up to you on these record field names

persistent/test/Database/Persist/QuasiSpec.hs
```diff
@@ -289,7 +322,7 @@ Car
             entityDB (unboundEntityDef vehicle) `shouldBe` EntityNameDB "vehicle"
 
         it "should parse the `entityAttrs` field" $ do
-            entityAttrs (unboundEntityDef bicycle) `shouldBe` ["-- | this is a bike"]
+            entityAttrs (unboundEntityDef bicycle) `shouldBe` []
```
parsonsmatt, line 325: bugfix hell ya

persistent/test/Database/Persist/QuasiSpec.hs
```diff
@@ -323,6 +356,8 @@ User
     name            Text
     emailFirst      Text
     emailSecond     Text
+    !yes            Int
+    ~no             Int
```
parsonsmatt, line 360: oh hell yeah great catch

parsonsmatt (commented):
OK, code looks good to me! A few remarks on making it a slightly less breaking change here

persistent/test/Database/Persist/QuasiSpec.hs
dtpowl: 🙏 from testing in mwb

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: :+1: and I can get rid of `indented'` entirely by doing this

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: Never mind — I did it in this branch.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,680 @@
...
+            , stateOffset = 0
+            , statePosState = initialPosState
+            , stateParseErrors = []
+            }
+
+-- @since 2.16.0.0
+data CumulativeParseResult a = CumulativeParseResult
+    { cumulativeErrors :: [ParseErrorBundle String Void]
+    , cumulativeData :: a
+    }
```
parsonsmatt, line 101: This datatype is a bit smelly - the expectation here is that we want to represent a parse that *definitely* succeeded (and thus has a `cumulativeData`), but that also may have some list of errors (which may be empty).

But looking down at `toCumulativeParseResult`, we see `(Monoid a) => ` in the constraints - and `cumulativeData` is set to `mempty`. This means that we have a conflation of states: a parser that succeeded and returns `mempty`, and a parse that fails and we provide `mempty`.

persistent/Database/Persist/TH/Internal.hs
```diff
@@ -282,7 +282,11 @@ embedEntityDefsMap existingEnts rawEnts =
 --
 -- @since 2.5.3
 parseReferences :: PersistSettings -> [(Maybe SourceLoc, Text)] -> Q Exp
-parseReferences ps s = lift $ parse ps s
+parseReferences ps s = do
+  let cpr = parse ps s
+  case renderErrors cpr of
+    Nothing -> lift $ cumulativeData cpr
+    Just errorMessage -> fail errorMessage
```
parsonsmatt, line 289: So here's where `CumulativeParseResult` is finally consumed. We use `renderErrors` to determine if there are any errors, and if not, we `lift` the cumulative data. If there are errors, we fail with error message.

I think we probably want to represent `CumulativeParseResult` instead as a `Either (NonEmpty Error) a`. Then the resulting logic is all much simpler.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
parsonsmatt: OK, as stated elsewhere, this should be more like:

```haskell
type CumulativeParseResult = Either (NonEmpty (ParseErrorBundle String Void))
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,680 @@
...
+type ExtraState = [(SourcePos, CommentToken)]
+
+type Parser a =
+    StateT
+        ExtraState
+        (ParsecT Void String (Either (ParseErrorBundle String Void)))
+        a
+type InternalParseResult a =
+    Either (ParseErrorBundle String Void) (a, ExtraState)
+type ParseResult a = Either (ParseErrorBundle String Void) a
```
parsonsmatt, line 59: I recommend reusing synonyms so you get some benefit of centralized naming:

```suggestion
type InternalParseResult a =
    ParseResult (a, ExtraState)
    
type ParseResult a = Either (ParseErrorBundle String Void) a
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
parsonsmatt: ... Which is `type ParseResult a` already!

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,680 @@
...
+toCumulativeParseResult (Left peb) =
+    CumulativeParseResult
+        { cumulativeErrors = [peb]
+        , cumulativeData = mempty
+        }
+toCumulativeParseResult (Right res) =
+    CumulativeParseResult
+        { cumulativeErrors = []
+        , cumulativeData = res
+        }
```
parsonsmatt, line 118: I don't actually see where this is being called. Is GitHub hiding it from me? 🤔

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -187,279 +190,93 @@ data PersistSettings = PersistSettings
...
-        Token (T.concat [beforeEquals, "=", next]) : rest
-
-    | otherwise =
-        let (token, rest) = T.break isSpace t
-         in Token token : tokenize rest
+parse
+  :: PersistSettings
+  -> [(Maybe SourceLoc, Text)]
+  -> CumulativeParseResult [UnboundEntityDef]
+parse ps = foldMap $ toCumulativeParseResult . uncurry parseChunk
```
parsonsmatt, line 229: oh dang it i had marked the file as "viewed" -_-

parsonsmatt: OK, so what we were really wanting here was using `foldMap` to pick the `(Monoid a) => Monoid (CumulativeParseResult a)` instance, which `mappend`ed the relevant errors and results. This would run every `parseChunk` and get all the errors.

This is a pretty common pattern and the type [`Validation`](https://www.stackage.org/haddock/lts-23.21/either-5.0.2/Data-Either-Validation.html) is what we want for it. That type has an `Applicative` instance that does this:

```haskell
instance (Semigroup e) => Applicative (Validation e) where
  pure = Success

  Success f <*> Success a = Success (f a)
  Failure e <*> Failure e' = Failure (e <> e')
  Failure e <*> Success _ = Failure e
  Success _ <*> Failure e = Failure e
```

Then, instead of `foldMap :: (a -> m) -> [a] -> m`, we use `traverse :: (a -> Validation (NonEmpty ParseErrorEtc) b) -> [a] -> Validation (NonEmpty ParseErrorEtc) [b]`.

Finally, we can case on that, and get:

```haskell
case result of
  Failure errors -> renderAndReportErrors errors
  Success entities -> pure entities
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,680 @@
...
+        (string "--" <|> string "#")
+            *> hspace
+            *> takeWhileP (Just "character") (/= '\n')
+    pure (pos, Comment (Text.pack content))
+
+skipComment :: Parser ()
+skipComment = do
+    content <- docComment <|> comment
+    comments <- get
+    put $ comments ++ [content]
```
parsonsmatt, line 202: Appending to the end of a list is bad for performance, recommend using a `DList` instead for this case

(another option is to use it as a cons list and then reverse when you extract it)

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,680 @@
...
+        , parenthetical
+        , ptext
+        ]
+
+class Block a where
+    blockFirstPos :: a -> SourcePos
+    blockMembers :: a -> [Member]
+    blockSetMembers :: [Member] -> a -> a
+    blockSetNELMembers :: NonEmpty Member -> a -> a
+    blockSetDocCommentBlock :: Maybe DocCommentBlock -> a -> a
```
parsonsmatt, line 350: This isn't an idiomatic use of type classes. Where you reach for a class in OOP, in Haskell, you want to reach for a datatype.

```haskell
data Block = Block
    { blockFirstPos :: SourcePos
    , blockMembers :: [Member]
    , blockDocCommentBlock :: Maybe DocCommentBlock
    }
```

And then `EntityBlock` and `ExtraBlock` both contain a `Block` (along with whatever else they end up having)

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,680 @@
...
+    , extraBlockHeaderPos :: SourcePos
+    }
+    deriving (Show)
+
+data ExtraBlock = ExtraBlock
+    { extraBlockDocCommentBlock :: Maybe DocCommentBlock
+    , extraBlockExtraBlockHeader :: ExtraBlockHeader
+    , extraBlockMembers :: NonEmpty Member
+    }
+    deriving (Show)
```
parsonsmatt, line 433: So if we have `data Block` as mentioned above, then we can factor out the structure here:

```suggestion
data EntityBlock = EntityBlock
    { entityBlockBlock :: Block
    , entityBlockEntityHeader :: EntityHeader
    }
    deriving (Show)

data ExtraBlockHeader = ExtraBlockHeader
    { extraBlockHeaderKey :: Text
    , extraBlockHeaderRemainingTokens :: [Token]
    , extraBlockHeaderPos :: SourcePos
    }
    deriving (Show)

data ExtraBlock = ExtraBlock
    { extraBlockExtraBlockHeader :: ExtraBlockHeader
    , extraBlockBlock :: Block
    }
    deriving (Show)
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,680 @@
...
+    blockSetMembers ms b = case NEL.nonEmpty ms of
+        Nothing -> b
+        Just nel -> blockSetNELMembers nel b
+    blockSetNELMembers nel b = b{extraBlockMembers = nel}
+    blockSetDocCommentBlock dcb b = b{extraBlockDocCommentBlock = dcb}
+
+blockLastPos :: (Block a) => a -> SourcePos
+blockLastPos b = case blockMembers b of
+    [] -> blockFirstPos b
+    members -> maximum $ fmap memberEndPos members
```
parsonsmatt, line 371: And now this becomes

```suggestion
blockLastPos :: Block -> SourcePos
blockLastPos b = case blockMembers b of
    [] -> blockFirstPos b
    members -> maximum $ fmap memberEndPos members
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,680 @@
...
+    headerLines = takeWhile (\(spos, _t) -> sourceLine spos < startLine) cls
+    memberLines = dropWhile (\(spos, _t) -> sourceLine spos <= startLine) cls
+    dcb = docCommentBlockFromPositionedTokens headerLines
+    commentedMembers = associateCommentLinesWithMembers memberLines (blockMembers a)
+
+associateCommentLinesWithMembers
+    :: [(SourcePos, CommentToken)] -> [Member] -> [Member]
+associateCommentLinesWithMembers [] ms = ms
+associateCommentLinesWithMembers _ [] = []
+associateCommentLinesWithMembers cls ms@(mh : mt) = do
```
parsonsmatt, line 596: I think we can do this association much easier by using `StateT` while parsing. Briefly,

```haskell
parseThing = do
  comments <- get -- gets the current comments
  put [] -- clears the accumulated comments
  thing <- parseThingUncommented
  pure thing { thingComments = comments }
```

parsonsmatt (commented):
Some more changes :) thanks for pushing so hard on this!

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,617 @@
...
+type EntityParseError = ParseErrorBundle String Void
+type InternalParseResult a = Either EntityParseError (a, ExtraState)
+
+type ParseResult a = Either EntityParseError a
+type CumulativeParseResult a = Either [EntityParseError] a
+
+cumulativeData :: (Monoid a) => CumulativeParseResult a -> a
+cumulativeData cpr = case cpr of
+    Left _ -> mempty
+    Right r -> r
```
parsonsmatt, line 80: You almost never want a function signature like `MaybeOrEither a -> a`. Instead, users should expect to `case` on the `Either` and handle that `Left` themselves (or use a function like `either`, or `for_`, etc to handle it)

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,617 @@
...
+            , statePosState = initialPosState
+            , stateParseErrors = []
+            }
+
+-- | Converts the errors in a CumulativeParseResult to a String
+-- @since 2.16.0.0
+renderErrors :: CumulativeParseResult a -> Maybe String
+renderErrors cpr = case cpr of
+    Right _ -> Nothing
+    Left errs -> Just $ intercalate "\n" $ fmap errorBundlePretty errs
```
parsonsmatt, line 121: This function destroys information - you can see it syntactically in `Right _ -> Nothing` ! 

Instead of operating on a `Either Err a -> Maybe String`, we should do `Err -> String`. That way we are doing as little as possible with each function.

End users can recover the same behavior by using the `either` function to handle the `Either` - `either (Just . renderErrors) (const Nothing)` will work on `Either Err a -> Maybe String` type. But we only want users to be able to render errors if they actually *have* errors, which they'd acquire by way of:

```haskell
case parseResult of
  Left errs -> fail $ renderErrors errs
  Right a -> lift a
```

persistent/Database/Persist/TH/Internal.hs
```diff
@@ -280,9 +280,13 @@ embedEntityDefsMap existingEnts rawEnts =
...
 --
--- @since 2.5.3
+-- @since 2.16.0.0
 parseReferences :: PersistSettings -> [(Maybe SourceLoc, Text)] -> Q Exp
-parseReferences ps s = lift $ parse ps s
+parseReferences ps s = do
+  let cpr = parse ps s
+  case renderErrors cpr of
+    Nothing -> lift $ cumulativeData cpr
+    Just errorMessage -> fail errorMessage
```
parsonsmatt, line 289: If you make `renderErrors` operate on `[EntityError]`, then this line should be:

```suggestion
  case cpr of
    Right res -> lift res
    Left errors -> fail $ renderErrors errors
```

One article I recommend is [Parse, Don't Validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) - which goes into why it's good to use datatypes like `Either` etc to refine information we have about types. You can think of `A -> Maybe B` as separating `A` into `B` and `Nothing`, which removes all hope of understanding what made the `A` a not-`B`. Instead, doing `A -> Either C B` lets you retain that information, and then use case-splitting to recover that information

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -0,0 +1,617 @@
...
+            , entityHeaderRemainingTokens = rest
+            , entityHeaderPos = pos
+            }
+
+appendCommentToState :: (SourcePos, CommentToken) -> Parser ()
+appendCommentToState ptok = do
+    es <- get
+    let
+        comments = esPositionedCommentTokens es
+    void $ put es{esPositionedCommentTokens = ptok : comments}
```
parsonsmatt, line 449:
```haskell
modify f = do
  s <- get
  put (f s)
```

```suggestion
    modify $ \es -> 
        let
            comments = esPositionedCommentTokens es
        in es{esPositionedCommentTokens = ptok : comments}
```

parsonsmatt (approved):
Woohoo! Let's get this in!

parsonsmatt (approved):

sternenseemann:
Can `persistent < 2.16` on Hackage be revised to disallow `persistent-test >= 2.13.1.4`? This causes rather confusing test failures right now since the solver is free to pick a combination of the two packages where the changed parsing doesn't work yet.

PR #1588, opened 2025-04-18
don't export Span from Types

dtpowl:
Per a conversation with @parsonsmatt, exposing `Span` externally was an unintentional change. It is used in spec, however, so this PR moves it into its own module.
 
---

After submitting your PR:


dtpowl:
Closing this since it sounds like we'd rather just get this change and the parser rework out in a single release

PR #1589, opened 2025-04-22
optionally warn when encountering tabs in whitespace

dtpowl:
This PR's branch is based on #1584. Most of the changes in this branch are also present in that PR's branch. I strongly recommend looking at #1584  first, and I've marked this PR as a draft until that one has been reviewed.

Per a conversation with @parsonsmatt, we would like the ability to optionally emit an error or warning when the entity DSL parser encounters tab characters in whitespace. This PR adds that, with the error level (error, warning, or none) configured in `PersistSettings`. 

To maximize compatibility with the old parser, each tab is treated as a single character of whitespace.

---

After submitting your PR:


persistent/test/Database/Persist/QuasiSpec.hs
```diff
@@ -91,169 +98,223 @@ spec = describe "Quasi" $ do
...
         let
+            tokenize :: String -> ParseResult [Token]
             tokenize s = do
-                (d, c) <- runConfiguredParser initialExtraState (some anyToken) "" s
-                pure d
+              let (warnings, res) = runConfiguredParser defaultParserSettings initialExtraState (some anyToken) "" s
+              case res of
+                Left peb ->
+                  (warnings, Left peb)
+                Right (tokens, _acc) -> (warnings, Right tokens)
```
parsonsmatt, line 107: tbh I think I would have this discard warnings for this function and another function `tokenizeWithWarnings` that returns warnings

This allows you to keep most of the tests the same, but then have new tests with new behavior that should have warnings

persistent/Database/Persist/TH/Internal.hs
```diff
@@ -283,8 +283,9 @@ embedEntityDefsMap existingEnts rawEnts =
 -- @since 2.16.0.0
 parseReferences :: PersistSettings -> [(Maybe SourceLoc, Text)] -> Q Exp
 parseReferences ps s = do
-  let cpr = parse ps s
-  case cpr of
+  let (warnings, res) = parse ps s
+  _ <- reportWarning $ renderWarnings warnings
```
parsonsmatt, line 287: would suggest reporting as multiple warnings

```suggestion
  _ <- traverse (reportWarning . renderWarning) warnings
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -1,11 +1,15 @@
...
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TupleSections #-}
 
 module Database.Persist.Quasi.Internal.ModelParser
-    ( SourceLoc (..)
+    ( ParserSettings (..)
+    , defaultParserSettings
```
parsonsmatt, line 11: We'll need to expose a safe interface for `ParserSettings` from `Database.Persist.Quasi`

Suggest creating new module `Database.Persist.Quasi.ParserSettings` which defines the interface + `Database.Persist.Quasi.ParserSettings.Internal` which exports everything

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -113,11 +211,80 @@ runConfiguredParser acc parser fp s = parseResult
...
-toCumulativeParseResult prs = case partitionEithers prs of
-    ([], results) -> Right $ fold results
-    (errs, _) -> Left errs
+-- | Renders a list of ParserWarnings as a String,
+-- separated by line breaks.
+-- @since 2.16.0.0
+renderWarnings :: [ParserWarning] -> String
+renderWarnings warnings = intercalate "\n" $ fmap parserWarningMessage warnings
+
+-- Attempts to parse with a provided parser. If it fails with an error matching
```
parsonsmatt, line 220:
```suggestion
-- | Attempts to parse with a provided parser. If it fails with an error matching
```

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -113,11 +211,80 @@ runConfiguredParser acc parser fp s = parseResult
...
+-- This is useful when registering errors in space consumers and other parsers that are called
+-- with `try`, since a non-delayed error in this context will cause backtracking and not
+-- get reported to the user.
+tryOrRegisterError
+    :: String
+    -> (ParseError String Void -> Bool)
+    -> Parser a
+    -> Parser a
+    -> Parser a
+tryOrRegisterError msg f l r = do
```
parsonsmatt, line 264: If using single variable names, it is idiomatic to use `p` with a predicate (ie `a -> Bool`)

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -53,37 +101,87 @@ data ExtraState = ExtraState
...
+    sm = runReaderT (unParser parser) ps
+    pm = runStateT sm acc
+    wm = runParserT' pm initialInternalState
+    ((_is, either), warnings) = runWriter wm
+
+    -- Due to backtracking, it is possible that we have accumulated duplicate warnings.
+    -- For example, if two parsers which depend on the same subparser both attempt to
+    -- parse the same span of input, any warnings generated by that subparser will be
+    -- registered twice.
+    filteredWarnings = Set.toList $ Set.fromList warnings
```
parsonsmatt, line 183: Perhaps instead of using `[Warning]` in the `WRiterT` you use `Set Warning`? Then you get uniqueness by construction and more efficient additions anyway

persistent/test/Database/Persist/QuasiSpec.hs
```diff
@@ -25,19 +25,26 @@ import Test.QuickCheck
 import Text.Shakespeare.Text (st)
 import Text.Megaparsec (errorBundlePretty, runParser, some)
 
+import Database.Persist.Quasi.Internal
+import Database.Persist.Quasi.Internal.ModelParser
```
parsonsmatt, line 29: so these `Internal` imports are a bit smelly - can you make them explicit + ensure that anything we really need is available from `Database.Persist.Quasi` ?

persistent/test/Database/Persist/QuasiSpec.hs
dtpowl: Hmm — yeah, we could do that, but I'd advocate for keeping it the way it is now, because IMO _not_ having warnings is part of the "successful parse" behavior that the tests are intended to exercise. 

i.e., if we were to in the future make a change that inadvertently caused the parse task in one of the existing tests to generate a warning, we would want that test to start failing.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: Oh yeah, of course — that's way better

dtpowl: I'll still convert it back to a list at this point I think? since we'll want it to be a functor later

persistent/test/Database/Persist/QuasiSpec.hs
dtpowl: So at minimum we need to import `ModelParser` here in order to test it. We _could_ expose `ModelParser` from `Database.Persist.Quasi` simply in order to avoid importing `Internal`, but I'm a bit reluctant to do that because I can't think of any legitimate reason that a consumer of this library would need to interact with `ModelParser`. 

It must be exported from somewhere or it can't be used in these tests, but leaving it in an `Internal` module seems like a way to signal that it shouldn't be used. The library already seems to have the pattern of exposing some things needed in tests from `Internal` modules; see `import Database.Persist.EntityDef.Internal` on line 17.

I think at minimum explicitly importing the stuff we need from `Internal` is a good idea, but I have a slight preference for _not_ exposing `ModelParser` from `Database.Persist.Quasi`.

parsonsmatt: Yes - as discussed elsewhere, I'm fine importing internal modules, but I think we should try to use explicit imports and ensure that we're able to get what we need for a public API from the `Database.Persist.Quasi`

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
parsonsmatt: Do we need `Functor` specificlaly? We can do `Set.map`

persistent/test/Database/Persist/QuasiSpec.hs
dtpowl: 👍  💯

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
dtpowl: well, I guess it doesn't matter — we've thrown away the order by using Set in the first place

dtpowl: So I guess the right thing to do is to sort them by source position after retrieving them from the set and before presenting them

parsonsmatt (commented):
Code and imlpementation look great! A few comments on module organization

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -601,10 +769,14 @@ toParsedEntityDef mSourceLoc eb =
...
-        Left peb -> Left peb
+parseSource
+    :: PersistSettings
+    -> Maybe SourceLoc
+    -> Text
+    -> ParseResult [ParsedEntityDef]
+parseSource ps mSourceLoc source =
+    case parseEntities ps filepath (Text.unpack source) of
+        (warnings, Right blocks) -> (warnings, Right (toParsedEntityDef mSourceLoc <$> blocks))
+        (warnings, Left peb) -> (warnings, Left peb)
```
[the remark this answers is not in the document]
dtpowl: Neat!

persistent/Database/Persist/Quasi.hs
```diff
@@ -919,71 +919,19 @@ module Database.Persist.Quasi
     , PersistSettings
     , upperCaseSettings
     , lowerCaseSettings
-    -- ** Getters and Setters
-    , module Database.Persist.Quasi
```
parsonsmatt, line 923: hmm does this reduce the interface of the module? may be good to look at a diff of the generated haddocks. `stack haddock --open` for here + the haddocks on persistent hackage

persistent/Database/Persist/Quasi/PersistSettings/Internal.hs
```diff
@@ -0,0 +1,177 @@
...
+warningPos :: ParserWarning -> SourcePos
+warningPos = pstateSourcePos . parserWarningPosState
+
+instance Ord ParserWarning where
+    l <= r =
+        if warningPos l == warningPos r
+            then parserWarningMessage l <= parserWarningMessage r
+            else warningPos l <= warningPos r
+
+-- @since 2.16.0.0
```
parsonsmatt, line 84:
```suggestion
-- |
--
-- @since 2.16.0.0
```

these need to be in haddock comment blocks i think to render out

parsonsmatt (approved):
hell yeah let's roll

persistent/Database/Persist/Quasi.hs
dtpowl: Looks like the new interface is a superset of the old one, as it should be.

parsonsmatt (approved):

parsonsmatt (approved):
nice 😎 let's get this in

PR #1600, opened 2025-07-01
Expose Structured Mock Migrations

mercury-kuba:

- [X] Ran `fourmolu` on any changed files (`restyled` will do this for you, so
- [X] Adhered to the code style (see the `.editorconfig` and `fourmolu.yaml` files for details)

After submitting your PR:


persistent-postgresql/Database/Persist/Postgresql.hs
```diff
@@ -28,13 +28,10 @@ module Database.Persist.Postgresql
     ( withPostgresqlPool
     , withPostgresqlPoolWithVersion
     , withPostgresqlPoolWithConf
-
```
mercury-kuba, line 31: A lot of these are due to Fourmolu auto-formatting.

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -240,45 +281,624 @@ instance PGFF.FromField PgInterval where
...
+data AlterTable
+    = AddUniqueConstraint ConstraintNameDB [FieldNameDB]
+    | DropConstraint ConstraintNameDB
+    deriving (Show)
 
+-- | Represents a change to a Postgres DB in a statement.
+--
+-- @since 2.17.0.0
+data AlterDB
+    = AddTable EntityNameDB EntityIdDef [Column]
```
mercury-kuba, line 365: This constructor has changed to store table information instead of the raw text.

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -240,45 +281,624 @@ instance PGFF.FromField PgInterval where
...
+        T.concat
+            -- Lower case e: see Database.Persist.Sql.Migration
+            [ "CREATe TABLE " -- DO NOT FIX THE CAPITALIZATION!
+            , escapeE name
+            , "("
+            , idtxt
+            , if null nonIdCols then "" else ","
+            , T.intercalate "," $ map showColumn nonIdCols
+            , ")"
+            ]
```
mercury-kuba, line 671: We move the code that generates the `AddTable` migration text down here, but it's the same logic.

parsonsmatt:
oof i should have just mass formatted everything in one commit instead of doing it incrementally. apologies

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -240,45 +281,624 @@ instance PGFF.FromField PgInterval where
...
+type SafeToRemove = Bool
+
+-- | Represents a change to a Postgres column in a DB statement.
+--
+-- @since 2.17.0.0
+data AlterColumn
+    = ChangeType Column SqlType Text
+    | IsNull Column
+    | NotNull Column
+    | Add' Column
```
parsonsmatt, line 344: Is this `AddColumn`?

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -240,45 +281,624 @@ instance PGFF.FromField PgInterval where
...
+    | Drop Column SafeToRemove
+    | Default Column Text
+    | NoDefault Column
+    | Update' Column Text
```
parsonsmatt, line 348: Would also like to see a more descriptive name on here

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -240,45 +281,624 @@ instance PGFF.FromField PgInterval where
...
+    | AddReference EntityNameDB ConstraintNameDB [FieldNameDB] [Text] FieldCascade
+    | DropReference ConstraintNameDB
+    deriving (Show)
+
+-- | Represents a change to a Postgres table in a DB statement.
+--
+-- @since 2.17.0.0
+data AlterTable
+    = AddUniqueConstraint ConstraintNameDB [FieldNameDB]
```
parsonsmatt, line 357: Is it valid to have `[]` here? Or can we do `NonEmpty`?

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -240,45 +281,624 @@ instance PGFF.FromField PgInterval where
...
+-- current state in the database to the state described in
+-- Haskell.
+--
+-- @since 2.17.0.0
+mockMigrateStructured
+    :: [EntityDef]
+    -> EntityDef
+    -> IO (Either [Text] [AlterDB])
+mockMigrateStructured allDefs entity = do
+    case partitionEithers [] of
```
parsonsmatt, line 381: I think you need to modify this :)

parsonsmatt (commented):
Versions should be 2.17.1.0.

I have a few questions/refactors- if we're exposing things (even internally) it may be nice to make these refactors ahead of time. Nothing blocking though.

The more structured representations are very nice!

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
mercury-kuba: Updated to `AddColumn`

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
mercury-kuba: Fixed for `AddReference` - it's a little more annoying for `AddUniqueConstraint`. But I don't think it's valid to have it empty here.

mercury-kuba:
Ran the tests, here's the end of the output:

```
ERATED ALWAYS AS (5) STORED)
  should read a generated column [✔]
  should support adding or removing generation expressions from columns [ ]Migrating: CREATe TABLE "gen_migrate_test"("id" SERIAL8  PRIMARY KEY UNIQUE,"sickness" INT8 NOT NULL GENERATED ALWAYS AS (3) STORED,"cromulence" INT8 NOT NULL)
  should support adding or removing generation expressions from columns [✔]

Finished in 2.4905 seconds
283 examples, 0 failures
```

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
parsonsmatt: ok, we can just do followup issue then

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -372,29 +390,71 @@ data AlterDB
...
+-- DB changes required to migrate the Entity to the state
+-- described in Haskell, assuming it currently does not
+-- exist in the database.
+--
+-- @since 2.17.1.0
+mockMigrateStructured
+    :: [EntityDef]
+    -> EntityDef
+    -> IO (Either [Text] [AlterDB])
+mockMigrateStructured allDefs entity = return $ Right migrationText
```
parsonsmatt, line 453: This feels odd to me - if we're always succeeding with `Right`, why not just do that? Same re `IO`. Is this suppose dto be doing more?

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -895,9 +958,324 @@ findAlters defs edef col@(Column name isNull sqltype def _gen _defConstraintName
...
+                "CASCADE" ->
+                    Just Cascade
+                "SET NULL" ->
+                    Just SetNull
+                "SET DEFAULT" ->
+                    Just SetDefault
+                "RESTRICT" ->
+                    Just Restrict
+                _ ->
+                    error $ "Unexpected value in parseCascade: " <> show txt
```
parsonsmatt, line 1147: feels like this should `throwError` instead of `error`

persistent/ChangeLog.md
```diff
@@ -1,5 +1,12 @@
 # Changelog for persistent
 
+# 2.17.1.0
+
+* [#1600](https://github.com/yesodweb/persistent/pull/1600)
+    * Add `migrateStructured` to `Database.Persist.Postgresql.Internal`.
+      This allows you to access a structured representation of the proposed migrations
+      for use in your application.
```
parsonsmatt, line 8: Oh shoot. All the changes in this PR are actually in `persistent-postgresql` package. So this `CHANGELOG` entry needs to be relocated into `persistent-postgresql/CHANGELOG.md` and the relevant versioning followed.

Issue #1599, opened 2025-06-25
New parser has differing behavior on quoted attributes


parsonsmatt:
Surfaced in https://github.com/commercialhaskell/curator/issues/52#issuecomment-3002214345

This is a weird one - the [`stack`](https://github.com/commercialhaskell/stack/blob/9375334d448da6bac248e01f537ae46cb659835c/src/Stack/Storage/User.hs?plain=1#L100) repository has a definition that looks like this:

```haskell
share [ mkPersist sqlSettings
      , mkMigrate "migrateAll"
      ]
      [persistLowerCase|
PrecompiledCacheParent sql="precompiled_cache"
  platformGhcDir FilePath "default=(hex(randomblob(16)))"
  compiler Text
  cabalVersion Text
  packageKey Text
  optionsHash ByteString
  haddock Bool default=0
  library FilePath Maybe
  UniquePrecompiledCacheParent platformGhcDir compiler cabalVersion packageKey optionsHash haddock sql="unique_precompiled_cache"
  deriving Show
```

The `  platformGhcDir FilePath "default=(hex(randomblob(16)))"` line is strange. I'm not sure how this would have parsed in the old parser or been used, but the new one properly parses but fails to generate the right code.

parsonsmatt:
@dtpowl

dtpowl:
Working on this.

PR #1601, opened 2025-07-02
Warn on quoted attributes

dtpowl:
**This is a draft because I'm still testing it with downstream packages.**

For [#1599].

The old version of the parser allowed entity field attributes to be wrapped in quotation marks; the enclosing `"`s would get parsed away and the enclosed text passed through in the attribute text on the `ParsedEntityDefinition`.

I discussed this with @parsonsmatt , and we'd like to deprecate this behavior. In [#1599] it seems to have been in use as a workaround for a parser bug that no longer exists. (If it's necessary to escape whitespace in a field attribute, this can still be done by wrapping the attribute in `(` `)` instead of quotes.)

This PR restores the old behavior and adds a configurable deprecation message as a warning. It's not easy to do this without some further refactoring of the parser. Banning quotes in entity field arguments will also ban them in entity field _types_, which would be incorrect — field types can include typelevel string literals.

This is a good opportunity to make the parser smarter. It's currently pretty naive — essentially, it breaks each line of an entity definition block into tokens and then stops. In doing this, it ignores a lot of genuine syntactic data. It doesn't know that the line starts with the field name, is followed by a type, and is then followed by a series of attributes. It's very easy to write an field definition that parses successfully but fails semantically.

Improving this situation is an incremental directional step towards a formal specification for the language.

This PR:
- Implements a more exact parser for entity field definitions, including structural representation of the field name, type, strictness, and attributes.
- Separates parsing of entity fields from non-field things like `deriving` statements. I'm calling these "directives" here. In the interest of keeping the PR from growing even more huge, directives are parsed naively for now.
  - I'd like to add support for Haskell-style `deriving` syntax in the future; this will be easier when it's parsed separately.
- Temporarily restores the old parser's behavior for entity field definitions and directive arguments that are wrapped in quotes.
- Adds a configurable warning message when entity field definitions and directive arguments are wrapped in quotes.

Currently, all of this new structure is thrown away in `mkUnboundEntityDef`. There's something of an impedance mismatch:
- `mkUnboundEntityDef` doesn't care about the difference between field definitions and directives; it wants the parsed structure to be reduced back to a string of tokens.
- `mkUnboundEntityDef` wants to re-parse types itself.
- `mkUnboundEntityDef` wants to re-parse key/value attributes itself.
- etc, etc, etc

I think this situation could be improved with some effort, but that's outside the scope of this PR.

### DIFF
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -931,8 +931,8 @@ module Database.Persist.Quasi
     , setPsIdName
     , getPsTabErrorLevel
     , setPsTabErrorLevel
-    , getPsQuotedFieldAttributeErrorLevel
-    , setPsQuotedFieldAttributeErrorLevel
+    , getPsQuotedArgumentErrorLevel
+    , setPsQuotedArgumentErrorLevel
     ) where
 
 import Database.Persist.Quasi.PersistSettings
--- a/persistent/Database/Persist/Quasi.hs
+++ b/persistent/Database/Persist/Quasi.hs
@@ -186,7 +186,7 @@ User sql=big_user_table
 This will alter the generated SQL to be:
 
 @
-CREATE TABEL big_user_table (
+CREATE TABLE big_user_table (
     id      SERIAL PRIMARY KEY,
     name    VARCHAR,
     age     INT
@@ -784,9 +784,9 @@ check = do
 convert
     :: (Entity Vehicle, Maybe (Entity Bicycle), Maybe (Entity Car))
     -> Vehicle'
-convert (Entity _ (VehicycleBicycleSum _), Just (Entity _ (Bicycle brand)), _) =
+convert (Entity _ (VehicleBicycleSum _), Just (Entity _ (Bicycle brand)), _) =
     Bike brand
-convert (Entity _ (VehicycleCarSum _), _, Just (Entity _ (Car make model))) =
+convert (Entity _ (VehicleCarSum _), _, Just (Entity _ (Car make model))) =
     Car make model
 convert _ =
     error "The database preconditions have been violated!"
@@ -931,6 +931,8 @@ module Database.Persist.Quasi
     , setPsIdName
     , getPsTabErrorLevel
     , setPsTabErrorLevel
+    , getPsQuotedFieldAttributeErrorLevel
+    , setPsQuotedFieldAttributeErrorLevel
     ) where
 
 import Database.Persist.Quasi.PersistSettings
--- a/persistent/Database/Persist/Quasi/Internal.hs
+++ b/persistent/Database/Persist/Quasi/Internal.hs
@@ -16,7 +16,7 @@ module Database.Persist.Quasi.Internal
     , PersistSettings (..)
     , upperCaseSettings
     , lowerCaseSettings
-    , Token (..)
+    , Attribute (..)
     , SourceLoc (..)
     , sourceLocFromTHLoc
     , parseFieldType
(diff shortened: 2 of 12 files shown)


persistent/Database/Persist/Quasi.hs
```diff
@@ -186,7 +186,7 @@ User sql=big_user_table
 This will alter the generated SQL to be:
 
 @
-CREATE TABEL big_user_table (
+CREATE TABLE big_user_table (
```
dtpowl, line 189: Unrelated typo fix

persistent/Database/Persist/Quasi/Internal.hs
```diff
@@ -546,19 +545,22 @@ mkUnboundEntityDef ps parsedEntDef =
...
+    textFields = entityFieldContent . fst <$> fields
+
+    textDirectives :: [[Text]]
+    textDirectives = directiveContent . fst <$> directives
 
     entityConstraintDefs =
         foldMap
             (maybe mempty (takeConstraint ps entNameHS cols) . NEL.nonEmpty)
-            textAttribs
+            (textFields ++ textDirectives)
```
dtpowl, line 563: To avoid changing the external interface, we'll just smush the field definitions and directives together for now. We should improve this in the future.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -211,19 +211,18 @@ tryOrWarn msg p l r = do
             else parseError err
 
 -- | Attempts to parse with a provided parser. If it fails with an error matching
--- the provided predicate, it registers a delayed error with the provided message and falls
+-- the provided predicate, it registers a delayed error and falls
```
dtpowl, line 214: This comment was inaccurate; the `msg` argument was actually unused. I'm not sure that's what we actually want, but fixing it seemed outside the scope of this PR.

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -258,37 +257,118 @@ data SourceLoc = SourceLoc
...
+        (FieldName fn) = entityFieldName f
+     in
+        s <> fn
 
 commentContent :: CommentToken -> Text
 commentContent = \case
     Comment s -> s
     DocComment s -> s
 
+quotedAttributeErrorMessage :: String
```
dtpowl, line 344: We define a custom error message mostly so that we can fish it out again in `tryOrReport`, because it's otherwise not as easy as you'd think to determine when the quoted-attribute-forbidding parser has failed because of a quoted attribute as opposed to some other reason.

I don't love this but it works fine for now; a better solution would probably add 100+ LOC to this already-big PR.

persistent/Database/Persist/Quasi/Internal/TypeParser.hs
```diff
@@ -0,0 +1,179 @@
...
+
+-- | Parses a type expression in non-top-level contexts, where an unparenthesized type constructor
+-- application is acceptable.
+--
+-- @since 2.17.1.0
+innerTypeExpr :: ((MonadParsec e String) m) => m TypeExpr
+innerTypeExpr = typeExpr' True
+
+typeExpr' :: ((MonadParsec e String) m) => Bool -> m TypeExpr
+typeExpr' isInner = label "type expression" $ do
```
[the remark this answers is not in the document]
dtpowl: you're right; that's nicer

persistent/Database/Persist/Quasi/Internal/ModelParser.hs
```diff
@@ -470,33 +550,44 @@ parenthetical' = do
...
 
-ptext :: Parser Token
-ptext = label "plain token" $ do
+fieldStrictness :: Parser FieldStrictness
+fieldStrictness = label "strictness sigil" $ do
+    c <- char '!' <|> char '~'
+    case c of
+        '!' -> pure Strict
+        '~' -> pure Lazy
+        _ -> error "unreachable"
```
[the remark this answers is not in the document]
dtpowl: Ah, lovely

PR #1604, opened 2025-07-10
Use Interval type from postgresql-simple-interval

tfausak:
This more or less replaces `PgInterval` with `Interval` from [postgresql-simple-interval](https://hackage.haskell.org/package/postgresql-simple-interval-0.2025.7.10).

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql/Internal.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql/Internal.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE ViewPatterns #-}
 
 module Database.Persist.Postgresql.Internal
@@ -13,26 +14,21 @@ import qualified Database.PostgreSQL.Simple.Internal as PG
 import qualified Database.PostgreSQL.Simple.ToField as PGTF
 import qualified Database.PostgreSQL.Simple.TypeInfo.Static as PS
 import qualified Database.PostgreSQL.Simple.Types as PG
+import qualified Database.PostgreSQL.Simple.Interval as Interval
 
 import qualified Blaze.ByteString.Builder.Char8 as BBB
-import qualified Data.Attoparsec.ByteString.Char8 as P
-import Data.Bits ((.&.))
+import Control.Monad ((<=<))
+import Data.Bits (toIntegralSized)
 import Data.ByteString (ByteString)
 import qualified Data.ByteString.Builder as BB
-import qualified Data.ByteString.Char8 as B8
-import Data.Char (ord)
 import Data.Data (Typeable)
-import Data.Fixed (Fixed(..), Pico)
-import Data.Int (Int64)
+import Data.Fixed (Fixed(..), Micro, Pico)
 import qualified Data.IntMap as I
 import Data.Maybe (fromMaybe)
-import Data.String.Conversions.Monomorphic (toStrictByteString)
-import Data.Text (Text)
-import qualified Data.Text as T
-import qualified Data.Text.Encoding as T
-import Data.Time (NominalDiffTime, localTimeToUTC, utc)
+import Data.Time (NominalDiffTime, localTimeToUTC, nominalDiffTimeToSeconds, secondsToNominalDiffTime, utc)
 
 import Database.Persist.Sql
+import Database.Persist.Postgresql.Interval ()
 
 -- | Newtype used to avoid orphan instances for @postgresql-simple@ classes.
 --
@@ -117,7 +113,7 @@ builtinGetters = I.fromList
     , (k PS.time,        convertPV PersistTimeOfDay)
     , (k PS.timestamp,   convertPV (PersistUTCTime. localTimeToUTC utc))
     , (k PS.timestamptz, convertPV PersistUTCTime)
-    , (k PS.interval,    convertPV (PersistLiteralEscaped . pgIntervalToBs))
+    , (k PS.interval,    convertPV $ toPersistValue @Interval.Interval)
     , (k PS.bit,         convertPV PersistInt64)
     , (k PS.varbit,      convertPV PersistInt64)
     , (k PS.numeric,     convertPV PersistRational)
@@ -148,7 +144,7 @@ builtinGetters = I.fromList
     , (1183,             listOf PersistTimeOfDay)
     , (1115,             listOf PersistUTCTime)
     , (1185,             listOf PersistUTCTime)
-    , (1187,             listOf (PersistLiteralEscaped . pgIntervalToBs))
+    , (1187,             listOf $ toPersistValue @Interval.Interval)
     , (1561,             listOf PersistInt64)
     , (1563,             listOf PersistInt64)
     , (1231,             listOf PersistRational)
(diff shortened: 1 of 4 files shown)


persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -237,110 +235,45 @@ unBinary (PG.Binary x) = x
 newtype PgInterval = PgInterval {getPgInterval :: NominalDiffTime}
     deriving (Eq, Show)
 
-pgIntervalToBs :: PgInterval -> ByteString
-pgIntervalToBs = toStrictByteString . show . getPgInterval
-
 instance PGTF.ToField PgInterval where
-    toField (PgInterval t) = PGTF.toField t
+    toField = PGTF.toField . fromMaybe (error "PgInterval.toField") . pgIntervalToInterval
```
tfausak, line 239: This (and `toPersistValue`) should perhaps saturate/clamp the interval rather than throwing an impure exception. For example if you try to insert `PgInterval 9223372036855`, the existing `toPersistValue` method will generate `"9223372036855s"`, which Postgres will complain about: 

``` hs
SqlError
  { sqlState = "22015"
  , sqlExecStatus = FatalError
  , sqlErrorMsg = "interval field value out of range: \"9223372036855s\""
  , sqlErrorDetail = ""
  , sqlErrorHint = ""
  }
```

cabal.project
```diff
@@ -31,3 +31,8 @@ source-repository-package
 --     type: git
 --     location: https://github.com/ysangkok/hedis
 --     tag: 6f36989836b49974f51a6ee8edaf156490590980
+
+source-repository-package
+  type: git
+  location: https://github.com/MercuryTechnologies/postgresql-simple-interval
+  tag: c2e9e886162a37cc177986d223600567c2a156e8
```
tfausak, line 38: https://github.com/MercuryTechnologies/postgresql-simple-interval/pull/8

cabal.project
```diff
@@ -31,3 +31,8 @@ source-repository-package
 --     type: git
 --     location: https://github.com/ysangkok/hedis
 --     tag: 6f36989836b49974f51a6ee8edaf156490590980
+
+source-repository-package
+  type: git
+  location: https://github.com/MercuryTechnologies/postgresql-simple-interval
+  tag: aa85e163194fb2be4d884f3cdebf4d699b96601a
```
tfausak, line 38: https://github.com/MercuryTechnologies/postgresql-simple-interval/pull/9

tfausak:
Hmm, this error is unexpected:

``` hs
SqlError
  { sqlState = "22015"
  , sqlExecStatus = FatalError
  , sqlErrorMsg = "interval field value out of range: \"@ 0 mon 0 day +2147483648 us\""
  , sqlErrorDetail = ""
  , sqlErrorHint = ""
  }
```

That's just above `maxBound :: Int32`, which should very much be in range. The upper bound should be `maxBound :: Int64`, which is `9223372036854775807`. Perhaps this depends on the version of PostgreSQL?

tfausak:
Indeed, before version 15 Postgres was not able to parse all valid interval values. https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=e39f99046

I don't think this can really be handled at the library level. There's no way to encode all the possible intervals in a style that is accepted by all versions of Postgres.

tfausak:
Here are some examples of behavior before and after these changes:

<details><summary>Click to reveal extensions and imports.</summary>

``` hs
:set -XOverloadedStrings
:set -XTypeApplications

import Control.Monad.Logger as L
import Database.Persist.Sql as S
import Database.Persist.Postgresql as P
import Database.PostgreSQL.Simple.Interval as I

L.runStderrLoggingT (P.withPostgresqlConn "user=postgres" (\ c -> S.runSqlConn (S.rawSql @(S.Single P.PgInterval) "..." [...]) c))
```

</details>

``` hs
-- PgInterval can't represent days.
rawSql "select '1 day'::interval" []

-- before:
-- *** Exception: ConversionFailed {errSQLType = "interval", errSQLTableOid = column number 0 is out of range 0..-1
-- Nothing, errSQLField = "column number 0 is out of range 0..-1
-- ", errHaskellType = "PgInterval", errMessage = ":: Failed reading: satisfyWith"}

-- after:
-- *** Exception: user error (invalid interval)
```

``` hs
-- PgInterval can't represent months.
rawSql "select '1 month'::interval'" []

-- before:
-- *** Exception: ConversionFailed {errSQLType = "interval", errSQLTableOid = column number 0 is out of range 0..-1
-- Nothing, errSQLField = "column number 0 is out of range 0..-1
-- ", errHaskellType = "PgInterval", errMessage = ":: Failed reading: satisfyWith"}

-- after:
-- *** Exception: user error (invalid interval)
```

``` hs
-- PgInterval loses precision.
rawSql "select ?::interval" [toPersistValue $ PgInterval 0.0000009]

-- before:
-- [Single {unSingle = PgInterval {getPgInterval = 0.000001s}}]

-- after:
-- [Single {unSingle = PgInterval {getPgInterval = 0s}}]
```

``` hs
-- PgInterval overflows.
rawSql "select ?::interval" [toPersistValue $ PgInterval 9223372036854.775808]

-- before:
-- *** Exception: SqlError {sqlState = "22015", sqlExecStatus = FatalError, sqlErrorMsg = "interval field value out of range: \"9223372036854.775808s\"", sqlErrorDetail = "", sqlErrorHint = ""}

-- after:
-- *** Exception: PgInterval.toPersistValue
```

And for completeness, here's how the new `Interval` type handles those scenarios:

``` hs
-- Interval can represent days.
rawSql "select '1 day'::interval" []
-- [Single {unSingle = MkInterval {months = 0, days = 1, microseconds = 0}}]

-- Interval can represent months.
rawSql "select '1 month'::interval" []
-- [Single {unSingle = MkInterval {months = 1, days = 0, microseconds = 0}}]

-- Interval cannot lose precision (correct by construction).
rawSql "select ?::interval" [toPersistValue $ MkInterval 0 0 1]
-- [Single {unSingle = MkInterval {months = 0, days = 0, microseconds = 1}}]

-- Interval cannot overflow (correct by construction).
rawSql "select ?::interval" [toPersistValue $ MkInterval 0 0 9223372036854775807]
-- [Single {unSingle = MkInterval {months = 0, days = 0, microseconds = 9223372036854775807}}]
```

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -237,110 +240,50 @@ unBinary (PG.Binary x) = x
...
+        . toIntegralSized
+        . (\(MkFixed x) -> x)
+        . (realToFrac :: Pico -> Micro)
+        . nominalDiffTimeToSeconds
+        . getPgInterval
+
+intervalToPgInterval :: Interval.Interval -> Maybe PgInterval
+intervalToPgInterval interval
+    | Interval.months interval /= 0 = Nothing
+    | Interval.days interval /= 0 = Nothing
```
hdgarrood, line 277: It feels to me like it should be possible to handle nonzero `days` values here - can we essentially do `days * 86400 + seconds`?

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -237,110 +240,50 @@ unBinary (PG.Binary x) = x
...
+            . pgIntervalToInterval
+    fromPersistValue =
+        maybe (Left "invalid interval") pure
+            . intervalToPgInterval
+            <=< fromPersistValue
 
 instance PersistFieldSql PgInterval where
     sqlType _ = SqlOther "interval"
 
+pgIntervalToInterval :: PgInterval -> Maybe Interval.Interval
```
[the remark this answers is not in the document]
hdgarrood: Also, rather than having these here, I think I would prefer that `postgres-simple-interval` provide conversions to and from `NominalDiffTime` - in the case where you need to deal with an interval in both the database and in Haskell code, it's going to be a lot more comfortable to work with `NominalDiffTime` on the Haskell side.

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
hdgarrood: I understand that the current behaviour is not ideal, but is it necessary to change the `To/FromField` instances for `PgInterval` here? I'd prefer that we leave it as-is if we can (especially if the aim is for downstream users to stop using `PgInterval`), since it's possible that there are downstream users who are relying on the error being thrown at the database level in this case.

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
```diff
@@ -237,110 +241,50 @@ unBinary (PG.Binary x) = x
 newtype PgInterval = PgInterval {getPgInterval :: NominalDiffTime}
```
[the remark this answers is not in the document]
hdgarrood: I agree - can we add a comment here explaining why it's best avoided and that it might be deprecated in the future?

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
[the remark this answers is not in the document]
parsonsmatt: I think it may not be great for an interval to be backed by a `NominalDiffTime` - the syntax allows, for example, `INTERVAL '1 month'` where the exact `NominalDiffTime` this corresponds to would differ based on both operation and starting month. Same for `INTERVAL '1 year'` though that is down to leap seconds and leap days.

persistent-postgresql/Database/Persist/Postgresql/Internal.hs
tfausak: Unfortunately I think it is necessary to change the instance. The `builtinGetters` are type directed, so I'm forced to decode a PostgreSQL `interval` into a Haskell `Interval`. I could introduce a type that's essentially `Either PgInterval Interval`, but that feels like it's not worth the effort. I can't imagine that someone is relying on the behavior of `PgInterval` throwing a `SqlError` when it's too big.

parsonsmatt (approved):

parsonsmatt:
tagged and released, thank you!

PR #1612, opened 2025-12-03
Avoid N+1 in postgresql migrations

### LINKED ISSUE #1611: Batch migration performance
# Migration Performance

`persistent` migrations issue a *ton* of queries - it is not meant to be fast. However, it sure would be nice if it could be faster for tests!

The main problem is that we run migrations on a single entity def at a time. If we could do a batch of entity defs, we could save a lot of work.

This refers only to the postgres implementation.

# The Code Path

We call on the `connMigrateSql` field in `SqlBackend`.
Postgres populates that [here](https://github.com/yesodweb/persistent/blob/541358872357464809adc7f05ca91bfc586a4421/persistent-postgresql/Database/Persist/Postgresql.hs?plain=1#L509) with [`migrate'`](https://github.com/yesodweb/persistent/blob/541358872357464809adc7f05ca91bfc586a4421/persistent-postgresql/Database/Persist/Postgresql.hs?plain=1#L685-L690), which itself calls [`migrateStructured`](https://github.com/yesodweb/persistent/blob/541358872357464809adc7f05ca91bfc586a4421/persistent-postgresql/Database/Persist/Postgresql/Internal.hs?plain=1#L333-L381).

All of these functions have a signature like:

```haskell
x 
    :: [EntityDef]
    -- ^ The list of other entities, used for foreign key mapping
    -> (Text -> IO Statement) 
    -- ^ prepare a SQL statement from Text
    -> EntityDef
    -- ^ the entity we're migrating
    -> IO _
```

So this means that we run everything single-threaded, one per database table.

# The Per Table Work

We call [`getColumns`](https://github.com/yesodweb/persistent/blob/541358872357464809adc7f05ca91bfc586a4421/persistent-postgresql/Database/Persist/Postgresql/Internal.hs?plain=1#L339), defined [here](https://github.com/yesodweb/persistent/blob/541358872357464809adc7f05ca91bfc586a4421/persistent-postgresql/Database/Persist/Postgresql/Internal.hs?plain=1#L908-L971).

This does [one query to select the columns for the table](https://github.com/yesodweb/persistent/blob/541358872357464809adc7f05ca91bfc586a4421/persistent-postgresql/Database/Persist/Postgresql/Internal.hs?plain=1#L915-L930).
Extending this to support multiple tables would be relatively straightforward.

This also does [one query to select the constraints](https://github.com/yesodweb/persistent/blob/541358872357464809adc7f05ca91bfc586a4421/persistent-postgresql/Database/Persist/Postgresql/Internal.hs?plain=1#L950-L966).
This one has joins, so collating would be a little more tricky.

Unfortunately, we do *another* N+1 - we call `getColumn` for each column.

# The Per Column Work

In [`getColumn`](https://github.com/yesodweb/persistent/blob/541358872357464809adc7f05ca91bfc586a4421/persistent-postgresql/Database/Persist/Postgresql/Internal.hs?plain=1#L1003-L1009), we do one query, conditional on the column being a foreign key.

1. [`getRef`](https://github.com/yesodweb/persistent/blob/541358872357464809adc7f05ca91bfc586a4421/persistent-postgresql/Database/Persist/Postgresql/Internal.hs?plain=1#L1053)

We can feasibly reduce this to a single query, conditional on *any* column hvaing a foreign key.

# How many queries?

We currently do `count of tables + count of foreign keys` queries.
Aggregating us to a batch structure should, relatively cheaply, get us `one query for all tables + one query for each table containing a foreign key`.
That's a big improvement.
We *could* potentially try and get a single query, matching every table for their foreign keys, but that may be more difficult.

hdgarrood:
Fixes #1611. With these changes, we can migrate an arbitrary number of entities with just 4 queries.

- [x] Ran `fourmolu` on any changed files (`restyled` will do this for you, so
- [x] Adhered to the code style (see the `.editorconfig` and `fourmolu.yaml` files for details)

After submitting your PR:

### DIFF
--- a/persistent-postgresql/persistent-postgresql.cabal
+++ b/persistent-postgresql/persistent-postgresql.cabal
@@ -12,30 +12,32 @@ cabal-version:      >=1.10
 build-type:         Simple
 homepage:           http://www.yesodweb.com/book/persistent
 bug-reports:        https://github.com/yesodweb/persistent/issues
-extra-source-files: ChangeLog.md sql/*.sql
+extra-source-files:
+  ChangeLog.md
+  sql/*.sql
 
 library
   build-depends:
-      aeson               >=1.0
+      aeson                       >=1.0
     , attoparsec
-    , base                >=4.9     && <5
+    , base                        >=4.9     && <5
     , blaze-builder
-    , bytestring          >=0.10
-    , conduit             >=1.2.12
-    , containers          >=0.5
-    , file-embed          >=0.0.16
-    , monad-logger        >=0.3.25
+    , bytestring                  >=0.10
+    , conduit                     >=1.2.12
+    , containers                  >=0.5
+    , file-embed                  >=0.0.16
+    , monad-logger                >=0.3.25
     , mtl
-    , persistent          >=2.18   && <3
-    , postgresql-libpq    >=0.9.4.2 && <0.12
-    , postgresql-simple   >=0.6.1   && <0.8
-    , postgresql-simple-interval >=1 && < 1.1
+    , persistent                  >=2.18    && <3
+    , postgresql-libpq            >=0.9.4.2 && <0.12
+    , postgresql-simple           >=0.6.1   && <0.8
+    , postgresql-simple-interval  >=1       && <1.1
     , resource-pool
-    , resourcet           >=1.1.9
+    , resourcet                   >=1.1.9
     , string-conversions
-    , text                >=1.2
-    , time                >=1.6
-    , transformers        >=0.5
+    , text                        >=1.2
+    , time                        >=1.6
+    , transformers                >=0.5
     , unliftio-core
     , vault
 
@@ -71,11 +73,11 @@ test-suite test
   ghc-options:      -Wall
   build-depends:
       aeson
-    , base                       >=4.9 && <5
+    , base                        >=4.9 && <5
     , bytestring
     , containers
     , fast-logger
-    , hspec                      >=2.4
+    , hspec                       >=2.4
     , hspec-expectations
     , hspec-expectations-lifted
     , http-api-data


persistent-postgresql/test/MigrationSpec.hs
```diff
@@ -0,0 +1,404 @@
...
+import Database.Persist.Postgresql.Internal.Migration
+import qualified Database.Persist.SqlBackend.Internal as SqlBackend
+
+getConnPrepare
+    :: (Monad m) => SqlPersistT m (Text -> IO Statement)
+getConnPrepare = do
+    backend <- ask
+    pure (SqlBackend.connPrepare backend)
+
+-- NB: we do not perform these migrations in main.hs
```
hdgarrood: the test suite is kind of cursed in that it leaves the schema in the test DB, because this means that if there's a bug in the code that determines what migrations you need to apply, you only see test failures if you run the tests twice in a row (if you're starting from a fresh DB).

I think I've made things slightly worse here by doing this... but I also think it's important to directly exercise the migrator like this.

persistent-postgresql/Database/Persist/Postgresql/Internal/Migration.hs
```diff
@@ -0,0 +1,1107 @@
...
+        EntityDoesNotExist ->
+            (addTable newcols entity) : uniques ++ references ++ foreignsAlt
+        EntityExists ExistingEntitySchemaState{essColumns, essConstraints} ->
+            let
+                (acs, ats) =
+                    getAlters
+                        allDefs
+                        entity
+                        (newcols, udspair)
+                        (map dubiouslyRemoveReferences essColumns, Map.toList essConstraints)
```
hdgarrood: This function is basically the same as before except for this one line - in the previous version, we were relying on `getColumn` to implicitly remove these references (or rather not fetch them in the first place), now we're doing it explicitly here

persistent-postgresql/Database/Persist/Postgresql/Internal/Migration.hs
```diff
@@ -0,0 +1,1107 @@
...
+    dubiouslyRemoveReferences oldCol =
+        case List.find (\c -> cName c == cName oldCol) newcols of
+            Just new
+                | isNothing (cReference new) ->
+                    oldCol{cReference = Nothing}
+            _ ->
+                -- otherwise no-op, `getAlters` will handle dropping this for us.
+                oldCol
+
+-- | Indicates whether a Postgres Column is safe to drop.
```
hdgarrood: Everything from here downwards is exactly the same as it was before

Issue #1169, opened 2020-12-01
Self Reference foreign keys are not generated


kderme:
A migration like this:
```
Chain
    previousChain ChainId Maybe
    deriving Show Eq
```
will fail to generate any foreign keys at postgreSQL. I have narrowed down the bug to the `getAddReference` function, at the guard `guard $ table /= s && cname /= fieldDB (entityId entity)`, especially the first check. Removing that fixes the problem. I think the check was added for the sql=id option https://github.com/yesodweb/persistent/pull/1066, so simply removing it may break other stuff?

parsonsmatt:
Good catch thanks!

The specific line of code you're referencing is [this](https://github.com/yesodweb/persistent/pull/1066/files#diff-3a6af517565f4d723a74260939388b56ca107830b6a9ae68db5a877ca084f520R1122), I think. The reason I did that is because I wanted to forbid creating a foreign key from a table's `id` column to itself - something like

```
ALTER TABLE foo 
  ADD CONSTRAINT id REFERENCES foo(id)
```

I assumed that `cname :: DBName` was the name of the column that we were possibly creating a reference *from*. This bug report indicates that it's actually the column name of the *target* column. Which is a bit surprising to me!

kderme:
Yes this is the line I meant. There is another line https://github.com/yesodweb/persistent/pull/1066/files#diff-3a6af517565f4d723a74260939388b56ca107830b6a9ae68db5a877ca084f520R1064 which also causes the problem. It's a similar case, but this triggers when there is a migrations on an existing table, while the other when the table is new.

hdgarrood:
> The specific line of code you're referencing is [this](https://github.com/yesodweb/persistent/pull/1066/files#diff-3a6af517565f4d723a74260939388b56ca107830b6a9ae68db5a877ca084f520R1122), I think. The reason I did that is because I wanted to forbid creating a foreign key from a table's id column to itself - something like

> ```
> ALTER TABLE foo 
>  ADD CONSTRAINT id REFERENCES foo(id)
> ```

Do you remember what kind of Persistent entity definition might cause that? I guess something like this?

```
SelfFKPrimary
    Id SelfFKPrimaryId
```

In my local tests, this results in:

```haskell
    EntityDef
        { entityHaskell = EntityNameHS{unEntityNameHS = "SelfFKPrimary"}
        , entityDB = EntityNameDB{unEntityNameDB = "self_f_k_primary"}
        , entityId =
            EntityIdField
[...]

PR #1614, opened 2025-12-11
postgresql: Add foreign key constraints the first time when migrating

hdgarrood:
I'm pretty sure there isn't an issue for this already, which is a bit surprising! But in postgresql, when you add a foreign key column to an existing table, you need to generate migrations twice to get it fully set up: the first run only creates the column, and the second creates the FK constraint.

This PR adds a test demonstrating the issue as well as removing a guard which I'm reasonably confident isn't necessary (see https://github.com/yesodweb/persistent/issues/1169#issuecomment-3643993103). The commit titled "[refactor: don't match on all of the Column fields](https://github.com/yesodweb/persistent/pull/1614/changes/f99c56e362ad9f72e6fdb63ffc3d6b4390e5e991)" also does a refactor to avoid matching on each of the fields of the `Column` values, so that it's obvious which values come from the old vs new column, since I found it quite hard to get my head around what this function was doing to start with, but I'm happy to revert that one if you don't think it's an improvement. I recommend reviewing that commit separately from the others.

After submitting your PR:

### DIFF
--- a/persistent-postgresql/Database/Persist/Postgresql/Internal/Migration.hs
+++ b/persistent-postgresql/Database/Persist/Postgresql/Internal/Migration.hs
@@ -1162,6 +1162,10 @@ findAlters defs edef newCol oldCols =
                     )
   where
     refAdd Nothing = []
+    -- This check works around a bug where persistent will sometimes
+    -- generate an erroneous ForeignRef for ID fields.
+    -- See: https://github.com/yesodweb/persistent/issues/1615
+    refAdd _ | fmap fieldDB (getEntityIdField edef) == Just (cName newCol) = []
     refAdd (Just colRef) =
         case find ((== crTableName colRef) . getEntityDBName) defs of
             Just refdef ->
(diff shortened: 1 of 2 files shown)


persistent-postgresql/Database/Persist/Postgresql/Internal/Migration.hs
```diff
@@ -1097,80 +1097,87 @@ findAlters
...
                     ( modRef ++ modDef ++ modNull ++ modType ++ dropSafe
-                    , filter (\c -> cName c /= name) cols
+                    , filter (\c -> cName c /= cName newCol) oldCols
                     )
+  where
+    refAdd Nothing = []
+    -- This check works around a bug where persistent will sometimes
+    -- generate an erroneous ForeignRef for ID fields.
+    -- See: https://github.com/yesodweb/persistent/issues/1615
+    refAdd _ | fmap fieldDB (getEntityIdField edef) == Just (cName newCol) = []
```
hdgarrood: This guard is basically the same as it was before - before, we were checking against the old column's name, and now we're checking against the new column's name, because there won't always be an old column name to check against (eg in the event that the column doesn't yet exist). There's no meaningful difference - previously, we would only ever run this check after verifying that the old column name matched the new column name.

PR #1623, opened 2026-03-02
Add assignAll and copyAll

cdfa:
This PR adds 2 functions `assignAll` and `copyAll` intended to be used when you need to insert a record into the database, and if it already exists, completely overwrite it. These can be implemented generically if we add a `PersistField a` constraint in `tabulateEntityA`. I think it's unlikely that this breaks existing code, so I propose a minor version increment.

The implementation is slightly hacky because the `tabulateEntityA` wants to build an `Entity`, but neither `assignAll` nor `copyAll` require full entity data. I think adding a method to `PersistEntity` similar to `fieldLens` which works over `record` instead of `Entity record` is also worth considering, but for now I kept with the minimal change.

I also fixed the nix flake to account for the "new" `persistent-postgresql => postgresql-simple-interval>=1` dependency.

----

- [X] Ran `fourmolu` on any changed files (`restyled` will do this for you, so
- [X] Adhered to the code style (see the `.editorconfig` and `fourmolu.yaml` files for details)

After submitting your PR:

### DIFF
--- a/persistent-mysql/Database/Persist/MySQL.hs
+++ b/persistent-mysql/Database/Persist/MySQL.hs
@@ -5,11 +5,11 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE ViewPatterns #-}
-{-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE ScopedTypeVariables #-}
 -- Pattern match 'PersistDbSpecific'
 {-# OPTIONS_GHC -fno-warn-deprecations #-}
 
@@ -29,7 +29,6 @@ module Database.Persist.MySQL
       -- * @ON DUPLICATE KEY UPDATE@ Functionality
     , insertOnDuplicateKeyUpdate
     , insertManyOnDuplicateKeyUpdate
-    , assignAll
     , HandleUpdateCollision
     , copyField
     , copyUnlessNull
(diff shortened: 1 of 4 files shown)


persistent/Database/Persist/Class/PersistEntity.hs
```diff
@@ -176,7 +175,7 @@ class
     -- @since 2.14.0.0
     tabulateEntityA
         :: (Applicative f)
-        => (forall a. EntityField record a -> f a)
+        => (forall a. PersistField a => EntityField record a -> f a)
```
parsonsmatt: This isn't an acceptable change imo

persistent-mysql/Database/Persist/MySQL.hs
```diff
@@ -1556,6 +1582,24 @@ copyField
...
+    => [HandleUpdateCollision record]
+copyAll = snd $ runWriter $ tabulateEntityA $ \field ->
+    -- This assumes the tabulateEntityA implementation is not strict in the
+    -- returned field value (and the default implementation indeed isn't).
+    error "copyAll: field value was used"
+        <$ when
+            ( fieldHaskell (persistFieldDef field)
+                /= fieldHaskell (persistFieldDef @record persistIdField)
+            )
+            (tell [copyField field])
```
parsonsmatt, line 1573: this form feels very strange- i believe there should be a better way to write that 🤔 I know prairie has the right utilities but I'm not sure if persistent has them

persistent/Database/Persist/Class/PersistEntity.hs
parsonsmatt: we should have something like `FieldDict` in prairie - then use-sites would have `FieldDict PersistField rec`

cdfa: > This isn't an acceptable change imo

Can you explain why? Even if there is a better way to support functions like `assignAll` and `copyAll`, does it hurt to add this constraint?

I'm not familiar with `prairie`, but it seems doing it properly would involve replacing `PersistField` with `prairie`'s `Field`. That's a bigger change then what I'm comfortable taking on at the moment.

persistent-mysql/Database/Persist/MySQL.hs
cdfa: I don't know of any `persistent` utilities that could help here, but I'm open to more specific suggestions.

cdfa:
For anyone running into this in the future: I found that [putMany](https://hackage-content.haskell.org/package/persistent-2.18.1.0/docs/Database-Persist-Class.html#v:putMany) actually satisfies my needs.

